PR #3, opened 2017-08-04
Added endpoint to render html content of an NDC file

agnessa:
Endpoint at: `/ndcs/{country code}`

uses redcarpet to transform markdown to html, with a customised renderer to ensure links to media files point to S3. 

What is needed:
- S3 configuration in `.env`
- some country NDC to test with, which needs to be imported from S3 as described here #1

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,6 @@ gem 'turbolinks', '~> 5'
 # Use ActiveModel has_secure_password
 # gem 'bcrypt', '~> 3.1.7'
 gem 'aws-sdk', '~> 2'
-gem 'redcarpet'
 
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ gem 'turbolinks', '~> 5'
 # Use ActiveModel has_secure_password
 # gem 'bcrypt', '~> 3.1.7'
 gem 'aws-sdk', '~> 2'
+gem 'redcarpet'
 
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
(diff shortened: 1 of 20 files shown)


ticklemynausea:
Looks good to me!

We can render this easily in the react app using `dangerouslySetInnerHTML`; I don't really know what steps we can take to make sure the markup is sane and avoid nasty injections. This might not be a concern at all if this data comes from a trusted source.

agnessa:
Data comes from scraped pdf files, so I guess it's safe from injections. Because the html is produced by a markdown parser, hopefully no strange html either.

vectorsize:
Hi guys, whats the original format we get before turning it into `html`? `html` not really the best format to work with, as @ticklemynausea said we can use `dangerouslySetInnerHTML` which is discouraged, hence the name ^^

vectorsize:
If you have a markdown already that would be a better format for the frontend.

j8seangel:
I totally agree with @vectorsize, I prefer to get the .md and parse in the frontend using for example https://github.com/rexxars/react-markdown#notes (what doesn't use `dangerouslySetInnerHTML`) this makes the thing a lot more flexible.

simaob:
We have the NDC content in both Markdown and HTML. They are static and won't be changing, and are from a trusted source. So the risk is virtually zero.

But we can discuss next week, it wouldn't be a big change here.

agnessa:
So apparently some of the content won't be available in .md, so it's going to have to be html sadly. I will simplify this endpoint so that it goes straight for the .html version from S3.

agnessa:
Multiple changes resulting from recent discussions & developments:
- branch rebased with develop
- path is now `ndcs/:code/full`
- serves html straight from files provided from the client, since we won't have `.md` versions for all of the countries
- uses Nokogiri to amend image paths (images in S3, previously that was done by the markdown renderer)
- introduces more consistent naming for the respective controller, importer and rake task, to avoid confusion with NDC content a.k.a. NCD overview
- also renamed `content` -> `full_text` in database table
- importing rake task is now: `rake ndcs:full:import`

agnessa:
agreed, will fix - thanks!

PR #4, opened 2017-08-04
Import locations from csv to db

agnessa:
`rake locations:import`

File format consistent with proposed template (https://github.com/Vizzuality/climate-watch/blob/feature/upload_template_specs/doc/upload_template_specs.md#countries)


ticklemynausea:
it looks good to me. my only question is if the `locations.csv` file should be versioned or something that is read from an S3 bucket like we discussed for other data imports.

agnessa:
Yes, I'll make it read from S3, and add the iso2 property.

PR #8, opened 2017-08-10
Feature: header wrapper component

edbrett:
Very simple component for the header of each page that accepts props to style and children:
- Image for background
- className for specific layouts


app/javascript/app/components/header/header-styles.scss
```diff
@@ -1,29 +1,8 @@
...
-  padding: 0;
-  list-style: none;
-  display: flex;
-}
-
-.navItem {
-  margin-right: 10px;
+.header {
+  background-color: $theme-color;
+  min-height: 300px;
```
j8seangel: I think I'd move to a prop in the component to define small or large because we have differents heights:
https://app.zeplin.io/project/598b1dc89689e51e1cd1b5ce/screen/598b1e52169b9d278d8f8091 (small)
https://app.zeplin.io/project/598b1dc89689e51e1cd1b5ce/screen/598b1e4885316be317c8f2ab (large)

app/javascript/app/page/page.jsx
```diff
@@ -3,12 +3,15 @@ import Proptypes from 'prop-types';
...
-import styles from './page-styles.scss';
+import Header from 'components/header';
+
+import styles from "./page-styles.scss"; // eslint-disable-line
 
 const Layout = ({ children }) =>
   (<div>
     <TopBar />
     <Nav />
+    <Header />
```
j8seangel: I don't think that this component should be listed in the layout level because each page has its own Header... so I'd move it to the page.

app/javascript/app/components/header/header-styles.scss
edbrett: I thought the designs might change so it would be easier to change the css than the component.

app/javascript/app/page/page.jsx
edbrett: This was because you were working with routes and I didn't want to create a page with a route just to display it when it might change. So I put here so you could see it.

PR #10, opened 2017-08-10
Feature/navbar component

j8seangel:
Create the navbar component
- Basic styles as them will change soon.
- Mark as active in the current path.
- Redirect on route not found.


app/javascript/app/components/error-page/error-page-component.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+import layout from 'styles/layout.scss';
+
+const ErrorPage = () =>
+  (<div className={layout.content}>
+    <h1>Ups that route does not exist</h1>
```
edbrett: Ups should be 'Oops'

PR #11, opened 2017-08-10
Feature/buttons

edbrett:
Simple PR to add button and button group components:
- Button component that wraps the children with props for styles
- ButtonGroup component that shows the same 4 buttons each time


app/javascript/app/page/page.jsx
```diff
@@ -3,12 +3,31 @@ import Proptypes from 'prop-types';
...
 
 const Layout = ({ children }) =>
   (<div>
     <TopBar />
     <Nav />
+    <Header>
+      <Intro title="Climate Watch" description="A very useful site" />
+      <div className={styles.buttonGroup}>
+        <ButtonGroup />
+        <Button className="download-button">
```
j8seangel: Does it make something?

app/javascript/app/components/button/button-component.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+import styles from './button-styles.scss';
+
+const Button = (props) => {
```
j8seangel: I like the idea of having a button and pass as children the icon or the text!
👏

PR #13, opened 2017-08-11
Feature: anchor nav

edbrett:
Maps an array of 'links' into a nav bar that pushes the path to the url:
- Similar styles to main nav but adjusted
- Placement in other component of header
- Const ANCHOR_LINKS defined in other page comp for clarity

### DIFF
--- a/app/javascript/app/components/anchor-nav/anchor-nav-component.jsx
+++ b/app/javascript/app/components/anchor-nav/anchor-nav-component.jsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { NavLink } from 'react-router-dom';
+
+import styles from './anchor-nav-styles.scss';
+
+const AnchorNav = (props) => {
+  const { links } = props;
+  return (
+    <div className={styles.anchorNav}>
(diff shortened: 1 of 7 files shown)


app/javascript/app/components/anchor-nav/anchor-nav-component.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+import styles from './anchor-nav-styles.scss';
+
+const AnchorNav = (props) => {
+  const { links } = props;
+  return (
+    <div className={styles.anchorNav}>
+      {links.map((link, index) =>
+        (<NavLink
+          exact={index === 0}
```
j8seangel: Why do you only do exact with the first element? I think all of them should be exact https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Route.md#exact-bool

edbrett: Good point! I cannot remember why but I will change.

edbrett: Updated 🌻

PR #14, opened 2017-08-11
Feature/dropdown

edbrett:
Basic dropdown component extending React Select:
- Dropdown component that extends react select to add defaults
- Default icon for dropdown
- Module styles to overide library

OTHER: webpack node module importing now possible through components

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "history": "^4.6.3",
-    "raw-loader": "^0.5.1",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -16,13 +16,15 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "history": "^4.6.3",
+    "raw-loader": "^0.5.1",
```
j8seangel: Could we move it to devDependencies?

edbrett:
Updated.

PR #19, opened 2017-08-15
Feature/map component

j8seangel:
New map component using https://www.react-simple-maps.io/ with a open source contribution 
[included](https://github.com/zcreativelabs/react-simple-maps/issues/12)!

kapture 2017-08-15 at 17 05 17

It accepts the following props:
- zoomEnable: default false,
- dragEnable: default true,
- initial: object with the zoom and center initial position, default: { center: [0, 20], zoom: 1 }
- paths: array of paths to render, how to get it [here](https://github.com/Vizzuality/climate-watch/compare/feature/map-component?expand=1#diff-f087db8f67c3732fee143628f0500c25R8)
- onCountryClick: function to handle the country click
- computedStyles: function to get the styles of each country, by default uses [this](https://github.com/Vizzuality/climate-watch/compare/feature/map-component?expand=1#diff-d0166110212e6572db98e0ce55b7bf50R124).
  That function will receive the geography in each iteration so in that way we are able to style each one individually.

Remember to run `npm install` as we have new dependencies


app/javascript/app/components/map/map-component.jsx
```diff
@@ -0,0 +1,147 @@
...
+      zoomEnable,
+      dragEnable,
+      onCountryClick,
+      computedStyles
+    } = this.props;
+    return (
+      <div className={styles.wrapper}>
+        {zoomEnable &&
+          <div className={styles.actions}>
+            <button onClick={this.handleZoomIn}>+</button>
```
edbrett: could you not use the Button component here to include the SVGs from the assets folder for the buttons?

PR #21, opened 2017-08-16
Feature: compare accordion

edbrett:
comparecountries

An additional PR to Accordion feature which extends it to handle changing from normal accordion to the compare accordion. Visit http://localhost/ndcs/bra to view all features:
- Button in header to switch between compare and regular
- sync all compare and active country comparisons in the url
- only render additional definition list data inside accordion if in compare mode

Note: this is to demo the compare function inside a single accordion component. The accordion could be adapted to accept children but the store model for this view is quite specific. Please let me know your thoughts. The reason I have made this way is to create the simplest set of components possible.

Comments: I think there must be a nicer way to manage these url changes outside of the component, maybe creating a helper to do so.

### DIFF
--- a/app/javascript/app/components/accordion/accordion-component.jsx
+++ b/app/javascript/app/components/accordion/accordion-component.jsx
@@ -82,20 +82,21 @@ class Accordion extends Component {
                           {def.description}
                         </dd>
                         {compare &&
-                          <div className={styles.compare}>
-                            <dd className={styles.definitionCompare}>
-                              {data[countriesToCompare[0]]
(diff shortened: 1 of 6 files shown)


j8seangel:
I'd prefer to have a new page to the compare feature because I think it makes the things easier to handle, because:
- The header is different and we need to include the dropdown.
- We can parametrize the Accordion and merge the requested data in this new component.

So for example to compare Brazil and Spain I'll go for this url `/compare/bra/esp`

j8seangel:
We need to think about in a very different approach, so let's close this and use the other branch.

PR #22, opened 2017-08-17
Feature/countries component

j8seangel:
This PR includes the new countries select component what includes:
- Search filter
- Filtered list
- Highlighted map with matches
  - New feature: disable cache prop (to rerender the interactions)

This new component reuses all of the components above and customizes the actions with custom actions and reducer.

kapture 2017-08-17 at 10 15 38

TODO:  💅


j8seangel:
### BONUS
 .zip file to generate the component boilerplate, best hack ever by @vectorsize 🥇

app/javascript/app/components/map/map-component.jsx
```diff
@@ -11,47 +12,26 @@ import Button from 'components/button';
...
-  handleZoomOut() {
-    const { zoom } = this.state;
-    if (zoom > 1) {
-      this.setState({
-        zoom: zoom / 2
-      });
-    }
-  }
-
+class Map extends PureComponent {
```
vectorsize: Can we agree on having components be a functional component instead of a class?

app/javascript/app/components/ndc/ndc-component.jsx
```diff
@@ -1,16 +1,25 @@
...
 
-import { feature } from 'topojson-client';
-import geoData from 'app/data/world-50m.json';
+import paths from 'app/data/world-50m-paths';
 
-const paths = feature(geoData, geoData.objects[Object.keys(geoData.objects)[0]])
-  .features;
+const NDC = (props) => {
+  const handleCountryClick = (geography) => {
+    props.history.push(`ndcs/${geography.id}`);
```
vectorsize: Id like to think of this url "side effects" and perhaps encapsulating them in actions, it doesnt feel right doing that inside the component.

app/javascript/app/data/initial-state.js
```diff
@@ -1,10 +1,238 @@
...
   },
   countryNDC: {
     loading: false,
     loaded: false,
     data: {}
+  },
+  countrySelect: {
+    query: '',
+    countries: [
+      // TODO get countries list from carto
```
vectorsize: Could we turn this todo comments in issues instead? these comments often end up as garbage in the code... you can keep them along with gh-issues and then remove the comment when you're done if you prefer.

app/javascript/app/data/initial-state.js
```diff
@@ -1,10 +1,238 @@
...
     open: false
   },
   countryNDC: {
     loading: false,
     loaded: false,
     data: {}
+  },
+  countrySelect: {
+    query: '',
+    countries: [
```
vectorsize: One comment regarding big `initialState` files, we can chop the parts into individual files and export all of them merged as one data structure if it starts growing a lot.

app/javascript/app/utils/map.js
```diff
@@ -0,0 +1,24 @@
+export const getDefaultStyles = () => ({
```
vectorsize: My problem with this file inside utils is that when piece of functionality is not used in more than one component, by moving it to the top level you are breaking the modularisation of that component without any benefit. We have a couple of options here, either you have a utils file inside the module's folder, or you can also declare that function in the file where it's used.

app/javascript/app/components/map/map-component.jsx
j8seangel: isn't it the same?

app/javascript/app/components/ndc/ndc-component.jsx
j8seangel: But I need the history prop that the `withRouter` HOC provide me in the component, so could we do it in an action?

app/javascript/app/data/initial-state.js
j8seangel: Agree and I like it more, so I'll create and export the chopped part inside the component-reducer.

app/javascript/app/data/initial-state.js
j8seangel: Good point, I'll review the code and add them in the PT

app/javascript/app/utils/map.js
j8seangel: The modularization is done inside the own component with [defaultProp](https://github.com/Vizzuality/climate-watch/blob/develop/app/javascript/app/components/map/map-component.jsx#L128) styles, but I thought that the default styles for the app (and not for the component) will be necessary for other map containers and that is why I've moved there.

app/javascript/app/components/map/map.js
```diff
@@ -1,3 +1,34 @@
...
+import { connect } from 'react-redux';
 
-export default Component;
+import MapComponent from './map-component';
+import actions from './map-actions';
+
+export { default as reducers } from './map-reducers';
+export { default as actions } from './map-actions';
+
+const ZOOM_STEP = 2;
```
edbrett: Will we want this a a config prop?

PR #24, opened 2017-08-22
Feature/ css grid example

edbrett:
Simple css grid example on homepage for proof of concept.


j8seangel:
I like this way, @vectorsize what do you think about it?

j8seangel:
And thanks to @hectoruch to recommend us this chrome extension to see the grid https://github.com/ademilter/chrome-css-grid-highlighter

j8seangel:
@edbrett please review the changes that we did! I think this is a better approach.

PR #25, opened 2017-08-22
Feature/upload template specs

agnessa:
I propose to merge this branch, even if it's still in progress, so that I can reference the sections that are ready using links that are not going to disappear.

In this branch me and @benlaken worked on specifications of the upload templates, and this will be useful to reference to the client.


agnessa:
I'll merge this @benlaken since nobody opposed ;) I suggest to open branches to add / refine particular sections of this file as we go.

PR #29, opened 2017-08-23
Feature/historical emissions relational model and data import

ticklemynausea:


ticklemynausea:
Hey!

I did what you asked with regards to the year/value pairs; it does make much more sense this way. 

I added a spec to test the service, but I had to change a few things around with regards to the s3 stubbing; I couldn't find any recommended way to handle multiple stubs for multiple requests but what I came up with seems sensible enough.

I also added a spec to the models I created.

agnessa:
I like what you did with the spec stubbing. I'll resolve the conflicts and merge.

PR #34, opened 2017-08-23
Feature: auto complete search

edbrett:
Extending the search and results list components this creates a reusable autocomplete search:
- Set data from filters and country data
- Show suggestions on search for ndc endpoints
- Show 'go to search' option also
- additional css and theming


app/javascript/app/components/autocomplete-search/autocomplete-search.js
```diff
@@ -0,0 +1,46 @@
...
+export { default as component } from './autocomplete-search-component';
+export { initialState } from './autocomplete-search-reducers';
+export { default as reducers } from './autocomplete-search-reducers';
+export { default as styles } from './autocomplete-search-styles';
+export { default as actions } from './autocomplete-search-actions';
+
+const mapStateToProps = (state) => {
+  const { query } = state.autocompleteSearch;
+  return {
+    query: deburrUpper(query),
```
j8seangel: As we have a selector for that, I'll use it here

app/javascript/app/components/autocomplete-search/autocomplete-search.js
```diff
@@ -0,0 +1,46 @@
...
+    searchList: getFilteredCountriesWithPath(state.autocompleteSearch)
+  };
+};
+
+const AutocompleteSearchContainer = (props) => {
+  const onOptionClick = (selected) => {
```
j8seangel: To avoid bugs I'm going to move this to a component so we can use the react life cycle and in the component will unmount the query will be removed.

PR #36, opened 2017-08-24
Feature: NDCs country index page

edbrett:
Restructure the country page to use container correctly and rebuild the header with styles and actions:
- New header
- Container
- Styles

OTHER: small fixes to other components to improve reusability


app/javascript/app/pages/ndc-country/ndc-country.js
```diff
@@ -1,19 +1,34 @@
...
-import Component from './ndc-country-component';
+import NDCCountryComponent from './ndc-country-component';
 import actions from './ndc-country-actions';
 
 export { default as component } from './ndc-country-component';
 export { initialState } from './ndc-country-reducers';
 export { default as reducers } from './ndc-country-reducers';
 export { default as actions } from './ndc-country-actions';
 
+const getCountry = (countries, iso) =>
```
j8seangel: As we have selectors, could you move that logic to one of them?
Thanks!

app/javascript/app/pages/ndc-country/ndc-country-component.jsx
```diff
@@ -3,22 +3,47 @@ import Proptypes from 'prop-types';
...
-        <Header>
-          <Intro title={countryName} />
+        <Header image={background}>
+          <div className={cx(styles.doubleFold, styles.header)}>
+            <div className={styles.title}>
+              <Button
+                className={styles.backIcon}
+                color="transparent"
+                link="/ndcs"
+                type="icon"
```
j8seangel: I don't like it too much if we need a square, why don't pass a prop called square as a bolean? I think is much easy to understand. What do you think?

j8seangel:
Thank you for the changes 👏

PR #38, opened 2017-08-24
Feature/country list endpoint and geometry

ticklemynausea:


ticklemynausea:
The frontend team wanted to have this endpoint return geometry (topojson) info for each country; I found it adequate to add this to the `Location` model but I could've asked for opinions before implementing it like this.

What are your thoughts?

PR #40, opened 2017-08-25
Feature: countries data to components

edbrett:
Update the countries data provider to fetch from the API, and then remove any dummy initial state in other components to use the new live country data:
- Only load container page component if country data has been fetched
- autocomplete component updated to use new data
- countries select component updated to use new data
- sort nav countries list by name


app/javascript/app/pages/container/container.js
```diff
@@ -4,19 +4,26 @@ import { connect } from 'react-redux';
...
   constructor(props) {
     super(props);
     props.getCountries();
   }
 
   render() {
-    return createElement(Component, this.props);
+    return this.props.countriesLoaded
+      ? createElement(Component, this.props)
+      : null;
```
j8seangel: we'll need a loading state or something like that here, but is a good starting!

j8seangel:
And please remember to run `bundle exec rake locations:import` to get the countries data.

PR #42, opened 2017-08-25
Feature/ndcs page

j8seangel:
This PR includes the ndcs page with the following features:
- Fake ndcs data fetch and parser
- Filters population and router sync
- Choropleth map 🎉 

kapture 2017-08-25 at 23 24 23

**Known issues**: 
- first selection render doesn't fill the map
- open selectors up

**Todo**: 
- styles
- map legend
- table component (router ready)

**Other changes**:
- Dropdown now accepts label prop
- Header size prop


app/javascript/app/components/ndcs-map/ndcs-map.js
```diff
@@ -0,0 +1,143 @@
...
+    case 4:
+      return '#7aabd3';
+    default:
+      return '#25597c';
+  }
+};
+
+class NDCMapContainer extends PureComponent {
+  handleCountryClick = (geography) => {
+    this.props.history.push(`ndcs/country/${geography.id}`);
```
edbrett: I think this needs to be /ndcs.... otherwise if you click twice it will become /ndcs.../ndcs...

j8seangel: Good catch! changed and merged

PR #43, opened 2017-08-29
Feature/import ndc sdg linkages

agnessa:
`bundle exec rake sdgs:import` - will import SDGs (goals and targets)
`bundle exec rake ndc_sdg_targets:import` - will import thee NDC - SDG linkages

NOTE: for the second task to link correctly locations & ndcs need to be present in the database. Because it gets difficult to remember which tasks to run and in which order I added 2 general tasks `db:import` and `db:reimport` that does stuff in the right order; the difference between them is that the second one will drop all "dictionary" data like locations and SDGs, which are not normally removed by importers (only appended / updated) as they are shared resources, so use it for a full refresh and normally it should be fine to stick with the first one.

I prefixed all the new table names and put the models under a namespace.

### DIFF
--- a/app/models/ndc_sdg.rb
+++ b/app/models/ndc_sdg.rb
@@ -1,0 +1,5 @@
+module NdcSdg
+  def self.table_name_prefix
+    'ndc_sdg_'
+  end
+end
--- a/app/services/import_ndc_sdg_targets.rb
+++ b/app/services/import_ndc_sdg_targets.rb
@@ -1,0 +1,70 @@
+require 'csv'
+
+class ImportNdcSdgTargets
+  def call
+    NdcSdgTargetSector.delete_all
+    SdgSector.delete_all
+    NdcSdgTarget.delete_all
+    bucket_name = Rails.application.secrets.s3_bucket_name
+    file_name = 'data/ndc_sdg_targets.csv'
+    s3 = Aws::S3::Client.new
+    begin
+      file = s3.get_object(bucket: bucket_name, key: file_name)
+    rescue Aws::S3::Errors::NoSuchKey
+      Rails.logger.error "File #{file_name} not found in #{bucket_name}"
+      return
+    end
+    content = file.body.read
+    import_ndc_sdg_targets(content)
+  end
+
+  private
+
+  def import_ndc_sdg_targets(content)
+    CSV.parse(content, headers: true).each.with_index(2) do |row|
+      ndc = ndc(row)
+      sdg_target = sdg_target(row)
+      next unless ndc && sdg_target
+      ndc_sdg_target = NdcSdgTarget.find_or_create_by(
+        ndc: ndc,
+        sdg_target: sdg_target,
+        indc_text: row['INDC_text'],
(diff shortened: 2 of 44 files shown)


agnessa:
I'll work on the respective API endpoint in this branch, so please don't merge yet.

app/models/ndc_sdg.rb
```diff
@@ -0,0 +1,5 @@
+module NdcSdg
```
ticklemynausea: yay, models in modules!

app/services/import_ndc_sdg_targets.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  private
+
+  def cleanup
+    NdcSdg::NdcTargetSector.delete_all
+    NdcSdg::NdcTarget.delete_all
+    NdcSdg::Sector.delete_all
+  end
+
+  def read_from_s3(file_name)
```
ticklemynausea: After merging the changes in my current branch we can reuse the S3CSVReader module I moved this function into

agnessa:
```
{
   "iso_code3":"AFG",
   "links":{
      "self":"/api/v1/ndcs/AFG/sdgs"
   },
   "sectors":{
      "76":{
         "name":"Agriculture"
      },
      "77":{
         "name":"Water"
      },
      "75":{
         "name":"cross-cutting"
      },
...
   },
   "sdgs":{
      "1":{
         "title":"No poverty",
         "1.1":{
            "title":"By 2030, eradicate extreme poverty for all people everywhere, currently measured as people living on less than $1.25 a day",
            "sectors":[
               77,
               76,
               75
            ]
         },
         "1.2":{
            "title":"By 2030, reduce at least by half the proportion of men, women and children of all ages living in poverty in all its dimensions according to national definitions"
         }
      },
...
}
```

@j8seangel @edbrett here's what I propose in terms of format of the response to render the SDG section, please let me know how that goes with what you had in mind.

The important thing to note is the `sectors` property, which is a dictionary of sectors for the dropdown, indexed by sector identifiers. The bulk of the data goes into the `sgds` property, which is a nested structure of goals (indexed by goal number) and targets (also indexed by target number, which is a string!), which gives you names of goals, names of targets and the information about which sectors have been matched for the given target in the current country. The array of `sectors` contains identifiers of sectors matched to a target. This property is missing altogether if no sectors have been matched - let me know if perhaps it's better for you to return an empty array in that case.

edbrett:
Hey @agnessa and @j8seangel. Something that might be worth including here as this is very specific to the SDG goals and their branding, is the colour of each SDG?

app/services/import_ndc_sdg_targets.rb
ticklemynausea: We can do this now!

ticklemynausea: hm it might need a slightly bigger refactor since the module already returns a `CSV` object. we can always do it later on.

ticklemynausea:
lgtm! can merge

PR #47, opened 2017-09-01
Correctly handle the `topojson` field

ticklemynausea:
The `topojson` field in the cartodb payloads used for the topojson import is, sadly, a string instead of an object.

This commit fixes the import process so that this string is correctly handled. As such, some errors may occur with invalid payloads.

If an error occurs, the import process will jump over it and print an error message.


j8seangel:
👏

PR #49, opened 2017-09-04
Feature/cait dataset data model and import process

ticklemynausea:
I'm making the PR so that I can get some feedback on the work. Let's not merge this for now, I still might have changes to make depending on the feedback from the frontend team.


app/services/import_cait_indc.rb
```diff
@@ -0,0 +1,170 @@
...
+      indicator: CaitIndc::Indicator.
+        find_by(name: label[:indicator_name]),
+      name: label[:legend_item_name],
+      color: label[:color]
+    }
+  end
+
+  def location_datum_attributes(datum)
+    {
+      location: Location.find_by(wri_standard_name: datum[:country]),
```
agnessa: this can hopefully become an iso code lookup once data is amended

app/services/import_cait_indc.rb
```diff
@@ -0,0 +1,170 @@
...
+      each do |ind_v|
+        CaitIndc::Label.create!(label_attributes(ind_v))
+      end
+  end
+
+  def import_location_data
+    @data.
+      each do |d|
+        a = location_datum_attributes(d)
+        CaitIndc::LocationDatum.create!(a) if a[:location]
```
agnessa: maybe add some error logging when location not matched?

app/services/import_cait_indc.rb
```diff
@@ -0,0 +1,170 @@
...
+      ),
+      value: datum[indicator_key]
+    }
+  end
+
+  def import_categories
+    @indicators.
+      map { |r| r[:category] }.
+      sort_by.
+      uniq.
```
agnessa: I'd add stripping here as well

lib/tasks/cait_indc.rake
```diff
@@ -0,0 +1,6 @@
+namespace :cait_indc do
+  desc 'Imports the CAIT INDC dataset from the csv sources'
+  task import: :environment do
```
agnessa: in this PR https://github.com/Vizzuality/climate-watch/pull/43 there's a "import them all" rake task which I think is going to be handy as order of import starts to matter, just letting you know in case you were contemplating adding sth like that :)

app/services/import_cait_indc.rb
ticklemynausea: yep, I'm waiting on that

ticklemynausea:
I talked with the frontend guys and I'm going to do some deep changes to the serialization. After that we should be good to do the full review and merge.

lib/tasks/cait_indc.rake
ticklemynausea: yes, thanks for the heads up! 👍

agnessa:
👍

PR #72, opened 2017-09-12
Feature: polish search list results

edbrett:
A series of fixes and improvements to the NDC search:
- Set the search bar from the query param and create and action to update the results when the search is updated
- Add an on hover link button to each results taking you to the ndc/:iso/full?query=:search page
- Filter the results by the number of results and set the order to descending
- Updated styles
- Fix child key rendering


app/javascript/app/components/result-card/result-card-component.jsx
```diff
@@ -13,19 +18,28 @@ const ResultCard = props => {
...
-            id={match.idx}
+            to={`/ndcs/country/${result.iso_code3}/full?search=${query}&idx=0`}
             className={styles.match}
-            dangerouslySetInnerHTML={{ __html: match.fragment }}
-          />
+          >
+            <div
+              className={styles.text}
+              id={match.idx}
+              dangerouslySetInnerHTML={{ __html: match.fragment }}
```
j8seangel: Could you highlight the fragment with the highlighted part?
image

app/javascript/app/components/result-card/result-card-styles.scss
```diff
@@ -11,6 +11,9 @@
   border-right: solid 1px #e5e5eb;
   padding-top: 30px;
   padding-right: 25px;
+  position: relative;
+  margin-left: -2000px;
```
j8seangel: why do we need this?

app/javascript/app/components/result-card/result-card-component.jsx
```diff
@@ -13,19 +18,28 @@ const ResultCard = props => {
       </div>
       {result.matches &&
         result.matches.map(match => (
-          <div
+          <NavLink
             key={match.fragment}
-            id={match.idx}
+            to={`/ndcs/country/${result.iso_code3}/full?search=${query}&idx=0`}
```
j8seangel: Good work ! We need to update that `idx` once the API is fixed.

app/javascript/app/components/result-card/result-card-component.jsx
edbrett: Well spotted! I forgot about this thanks 👀

PR #78, opened 2017-09-18
Feature/about page

j8seangel:
Includes the about page with partner images and links:
image


app/javascript/app/pages/about/about.js
```diff
@@ -0,0 +1,160 @@
...
+import gizImage from 'assets/partners/giz.png';
+import googleImage from 'assets/partners/google.png';
+import jgcriImage from 'assets/partners/jgcri.png';
+import fmencbnsImage from 'assets/partners/fmencbns.png';
+import dbeisImage from 'assets/partners/dbeis.png';
+import swissImage from 'assets/partners/swiss.png';
+
+import Component from './about-component';
+
+const jointInitiative = {
```
edbrett: I can't remember what we decided about constants like this. Whether to keep them in initial state or in the container file?

PR #79, opened 2017-09-18
Feature/import and model adaptation dataset

ticklemynausea:
Endpoint is `/adaptations`

Filters available: locations
`/adaptations?location=PRT,ESP,USA,RUS`

### DIFF
--- a/app/controllers/api/v1/adaptations_controller.rb
+++ b/app/controllers/api/v1/adaptations_controller.rb
@@ -5,8 +5,24 @@ module Api
         variables = ::Adaptation::Variable.
           includes(values: [:location])
 
+        if location_list
+          variables = variables.where(
+            values: {locations: {iso_code3: location_list}}
+          )
+        end
+
         render json: variables,
                each_serializer: Api::V1::Adaptation::VariableSerializer
+      end
+
+      private
+
+      def location_list
+        if params[:location].blank?
+          nil
+        else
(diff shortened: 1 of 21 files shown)


app/controllers/api/v1/adaptations_controller.rb
```diff
@@ -0,0 +1,29 @@
+module Api
+  module V1
+    class AdaptationsController < ApiController
+      def index
+        variables = ::Adaptation::Variable.
+          includes(values: [:location])
+
+        if location_list
+          variables = variables.where(
```
simaob: Filter out variables for which there are no values matching the request.

app/services/import_adaptation.rb
```diff
@@ -0,0 +1,112 @@
+class String
```
simaob: Document and move method to a more clear place to avoid surprises. 🐒

spec/factories/adaptation_variable.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :adaptation_variable, class: 'Adaptation::Variable' do
+    name 'MyText'
+    slug { (0...5).map { (65 + rand(26)).chr }.join }
```
simaob: Use sequence instead of random, to avoid issues with uniqueness.

PR #81, opened 2017-09-18
Feature/historical emissions endpoints

ticklemynausea:
Add historical emissions endpoint: `/emissions`

The data behind this endpoint can be very large, so it is advised to always use filters. I believe I should also force the usage of some filters (or use defaults) because this endpoint could be a potential dos vector for the reasons mentioned above, but I'll need to know more about the kind of usage the FE will need.

This endpoint has the following filters available:
- location takes a comma-separated list of ISO country codes (3 character)
- source takes the id attribute of the `historical_emissions_data_sources` table
- gas takes the id attribute of the `historical_emissions_gases` table
- sector takes the id attribute of the `historical_emissions_sectors` table
- date_before will list only emissions for the year given and before
- date_after will list only emissions for the year given and after

example:

```
/emissions?location=PRT,ESP&source=1&gas=1&sector=1&date_before=2010&date_after=2000
```

### DIFF
--- a/app/models/historical_emissions/record.rb
+++ b/app/models/historical_emissions/record.rb
@@ -17,8 +17,6 @@ module HistoricalEmissions
       filters(records, params)
     end
 
-    private_class_method
-
     def self.filters(records, params)
       unless params[:location].blank?
         records = records.where(
@@ -36,5 +34,7 @@ module HistoricalEmissions
 
       records
     end
+
+    private_class_method :filters
   end
 end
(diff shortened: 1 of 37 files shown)


app/models/historical_emissions/record.rb
```diff
@@ -4,5 +4,37 @@ class Record < ApplicationRecord
...
+          :location,
+          :data_source,
+          :sector,
+          :gas
+        )
+
+      filters(records, params)
+    end
+
+    private_class_method
```
simaob: I don't think that this works like this. It seems that `private_class_method` takes the method name as argument.

PR #87, opened 2017-09-20
Include pages and component generator

j8seangel:
Were you very tired of the components boilerplate? 

I was, so now run `npm run genererate` and it will prompt with a name and a redux connection questions to generate your component. 

same with pages running `npm run generate:page`
kapture 2017-09-20 at 11 45 19


app/javascript/generators/component/index.js
```diff
@@ -0,0 +1,80 @@
...
+            : true;
+        }
+
+        return 'The name is required';
+      }
+    },
+    {
+      type: 'list',
+      name: 'redux',
+      message: 'Am I redux connected?',
```
edbrett: Better english would be "Am I connected to Redux?"

j8seangel:
Fixed the typo, thanks!

PR #89, opened 2017-09-20
Feature/ghg emissions boilerplate

j8seangel:
Page boilerplate with basic component to start with
image

Also includes the recharts default config that we set up yesterday


app/javascript/app/components/chart-stackedarea/chart-stackedarea-styles.scss
```diff
@@ -0,0 +1,2 @@
+@import '~styles/settings.scss';
```
edbrett: If you import layout you dont need settings, as layout depends on settings.

app/javascript/app/components/ghg-emissions/ghg-emissions-styles.scss
```diff
@@ -0,0 +1,2 @@
+@import '~styles/settings.scss';
```
edbrett: same again here.

app/javascript/app/pages/ghg-emissions/ghg-emissions-styles.scss
```diff
@@ -0,0 +1,2 @@
+@import '~styles/settings.scss';
```
edbrett: and here.

PR #92, opened 2017-09-22
Feature/ghg emissions filters

j8seangel:
This PR includes the filters selections for the ghg emissions graph connected with real API data.
kapture 2017-09-22 at 15 07 57

TODO: 
- support multiple selections
- colors

BONUS:
- Nav routes from router config adding `nav: true` and label param
- Style the error page and include our lovely 🙈 
- Fix NDC SDG hacking styles


app/javascript/app/components/nav/nav-component.jsx
```diff
@@ -19,50 +19,16 @@ class NavBar extends PureComponent {
               <Icon className={styles.logo} icon={cwLogo} />
             </NavLink>
           )}
-          <div className={styles.linkWrapper}>
+          {routes.map(route => (
             <NavLink
+              key={route.path}
```
edbrett: Nice.

app/javascript/app/utils/navigation.js
```diff
@@ -0,0 +1,22 @@
+import qs from 'query-string';
+
+export function getLocationParamUpdated(
```
edbrett: Do we want to do this? Or keep the util local for a more modular approach?

PR #94, opened 2017-09-25
Feature/fix ndcs endpoint filter

ticklemynausea:
There was an oversight when merging the last cait indc changes PR. A field that was used by the `filter` param of the `/ndcs` endpoint got removed, and an api call to `/ndcs?location=SOMELOCATION&filter=overview` was broken.

This PR fixes the backend to correctly filter NDCS by category type (category types are categories of indicators: `map` or `overview`, meaning the area of the site where the indicators will be displayed)

### DIFF
--- a/app/models/cait_indc/category.rb
+++ b/app/models/cait_indc/category.rb
@@ -4,5 +4,8 @@ module CaitIndc
                             join_table: :cait_indc_indicators_categories
     validates :name, presence: true
     validates :category_type, inclusion: {in: %w(map overview)}
+
+    scope :type_map, -> { where(category_type: 'map') }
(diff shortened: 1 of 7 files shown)


app/models/cait_indc/category.rb
```diff
@@ -4,5 +4,8 @@ class Category < ApplicationRecord
                             join_table: :cait_indc_indicators_categories
     validates :name, presence: true
     validates :category_type, inclusion: {in: %w(map overview)}
+
+    scope :type_map, -> { where(category_type: 'map') }
```
edbrett: great :). Always love them scopes.

PR #95, opened 2017-09-25
Fix: indicators map filters

edbrett:
Maintenance PR to handle change of value type for 'category_id' to 'category_ids' with an array in indicators. This fixes that parsing of the data in all the NDCs view to handle this correctly.


ticklemynausea:
Looks good! Thanks!

PR #100, opened 2017-09-28
Feature: multiselect for ghgs

edbrett:
This is mixed PR, covering the implementation of rendering the fetched data from the API into the GHG emissions graph. 
- Add a new selector for navigating between the version of the data (IPCC, ```gwp```)
- Fetch and parse correct data based on updated selectors (now all data for each break value is fetched and then filtered in the selector (see filterData selector)
- Function within selectors file are now held in ./...-utils to make life easier for us all
- Add a default "TOP_EMITTERS" constant to the selectors for 'location' to allow for setting limit in initial countries
- Abstract the parsing of the data for the graph to a single filter point => this allows for the graph to render independently of the selectors so that the legend and graph operate as true components.
- Change the graph to a line chart, keeping the StackedArea component for later :).

sep-28-2017 19-02-35

Other:
- Small changes to the config for the graph: domain of axes, formatting of the tooltip value and 

Note: the endpoint for /emissions/meta has changed structure. As a result there is a big difference in the selectors for the data

TODO: 
- Fix bug with duplicate keys in legend/tooltip
- Sort the order of Legend, Tooltip, Selector
- Limit the amount of data that cant be selected into the graph giving the user feedback
- Add a checkbox multi-select to the Filters dropdown
- Styles

### DIFF
--- a/app/javascript/app/components/ghg-emissions/ghg-emissions-actions.js
+++ b/app/javascript/app/components/ghg-emissions/ghg-emissions-actions.js
@@ -26,7 +26,8 @@ const fetchGhgEmissionsMeta = createThunkAction(
               dataParsed[key] = data[key].map(item => {
                 let newItem = {
                   value: item.id,
-                  label: upperFirst(item.name)
+                  label:
+                    key === 'locations' ? item.iso_code3 : upperFirst(item.name)
                 };
                 if (key === 'data_sources') {
                   newItem = {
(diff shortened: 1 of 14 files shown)


app/javascript/app/components/ghg-emissions/ghg-emissions-actions.js
```diff
@@ -23,10 +23,31 @@ const fetchGhgEmissionsMeta = createThunkAction(
           if (data) {
             const dataParsed = {};
             Object.keys(data).forEach(key => {
-              dataParsed[key] = data[key].map(item => ({
-                value: item.id,
-                label: upperFirst(item.name)
-              }));
+              dataParsed[key] = data[key].map(item => {
```
ticklemynausea: So here is all the ugliness that we discussed this morning... Sorry that my changes didn't help that much.

Perhaps it might help the code look a bit better if the transformation is factored out into a separate function.

PR #101, opened 2017-10-02
Fix/ndcs map selects

j8seangel:
Fix `getLocationParamUpdated` helper to accept a single param or an array of params.

Also:
- Added the European countries in the map.
- Fix prettier double quote in lint staged.

### DIFF
--- a/app/javascript/app/data/european-countries.js
+++ b/app/javascript/app/data/european-countries.js
@@ -1,0 +1,37 @@
+export const europeSlug = 'EU28';
+
+export const europeanCountries = [
+  'AUT',
+  'BEL',
+  'BGR',
+  'HRV',
+  'CYP',
+  'CZE',
+  'DNK',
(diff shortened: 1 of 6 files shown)


app/javascript/app/data/european-countries.js
```diff
@@ -0,0 +1,37 @@
+export const europeSlug = 'EU28';
```
edbrett: Excellent!

app/javascript/app/utils/navigation.js
```diff
@@ -1,9 +1,11 @@
 import qs from 'query-string';
+import isArray from 'lodash/isArray';
 
-export function getLocationParamUpdated(location, params, clear) {
+export function getLocationParamUpdated(location, params = [], clear = false) {
```
edbrett: Well done. Sorry for merging such terrible code :|

PR #102, opened 2017-10-03
Feature/multi select component

edbrett:
PR to handle converting the GHG emissions graph tag selection into a true multi select:
- Convert React Select to React Selectize, allowing wider range of functions and groups rendering
- Convert old select components into React Selectize
- Update data parsing and remove general bugs in the selector work flow for the graph
- Styles for lots of places :)

Note: this component extends the react selectize library without making and customisations, just using the props available of the library. As a result it can be reused and themed easily from project to project.

selector


app/javascript/app/components/dropdown/dropdown-styles.scss
```diff
@@ -1,123 +1,13 @@
...
+}
+
 .label {
   display: block;
   font-size: $font-size-s;
   font-weight: $font-weight-bold;
   color: $theme-color;
   margin-bottom: 4px;
 }
-
```
j8seangel: yaaayy 🎉

PR #103, opened 2017-10-04
Feature/styles alignment

j8seangel:
- Include new images.
- Get headers images directly by [route](https://github.com/Vizzuality/climate-watch/compare/feature/styles-alignment?expand=1#diff-8bd4004e99b976de4c53d34ee24999aeR56) adding `headerImage`.
- Import images [dynamically](https://github.com/Vizzuality/climate-watch/commit/cc48005927e83801af56f88c246866b92b8d5ca3) (very nice I think).
- Add that logic using recompose to make it simpler.
- Styles theme folder fixes.
- Clean up old files.

kapture 2017-10-04 at 12 31 27


app/javascript/app/components/header/header.js
```diff
@@ -1,3 +1,26 @@
...
+import { withProps } from 'recompose';
 
-export default Component;
+import HeaderComponent from './header-component';
+
+function importAllImagesFromFolder(r) {
+  const images = {};
+  r.keys().map(item => {
+    images[item.replace('./', '').replace('.jpg', '')] = r(item);
+    return null;
```
edbrett: Do you need this return. Can you not use forEach instead of map to remove need for return null

app/javascript/app/pages/about/about.js
```diff
@@ -155,6 +155,8 @@ const foundingBy = {
   ]
 };
 
-const sections = [jointInitiative, partnershipWith, adicionalData, foundingBy];
+const withSections = withProps(() => ({
+  sections: [jointInitiative, partnershipWith, adicionalData, foundingBy]
```
edbrett: adicional => additional (spelling)

j8seangel:
Resolved conflicts and fixed your comments, thanks for the review!

PR #105, opened 2017-10-04
Feature: sdg linkages update

edbrett:
Update to the SDG Linkages component for the country pages to handle an improved visualisation of the SDG targets in relation to the countries involvement.

Previously we were showing a coloured dot if the country has intersecting sectors for a given target, when a sector was selected, and then hiding the grey dots (lack of involvement). Now we are allowing the targets that the country has no involvement with in grey so the data can better be explored. In addition to this, the dots now change size based on whether the given target contains the sector that has been selected. 

Changes include:
- Active sector is now hold in the url as a param for persistence and sharability
- SDG dots variable in size and colour based on data
- Icons updated to svgs for better rendering
- Icons now in colour
- Selector and title updated to match new designs (dropdown top right)

screen shot 2017-10-04 at 19 21 41


app/javascript/app/components/sdg-card/sdg-card-component.jsx
```diff
@@ -2,23 +2,25 @@ import React, { PureComponent } from 'react';
...
-import sdg9 from 'assets/sdg-icons/icon_9';
-import sdg10 from 'assets/sdg-icons/icon_10';
-import sdg11 from 'assets/sdg-icons/icon_11';
-import sdg12 from 'assets/sdg-icons/icon_12';
-import sdg13 from 'assets/sdg-icons/icon_13';
-import sdg14 from 'assets/sdg-icons/icon_14';
-import sdg15 from 'assets/sdg-icons/icon_15';
-import sdg16 from 'assets/sdg-icons/icon_16';
-import sdg17 from 'assets/sdg-icons/icon_17';
+import Icon from 'components/icon';
```
j8seangel: Why don't use this https://github.com/Vizzuality/climate-watch/commit/cc48005927e83801af56f88c246866b92b8d5ca3#diff-c73b648ba0a802388cc26a5734d76122R14 ?
It imports the images dynamically what makes less annoying, isn't it?

j8seangel:
👏

PR #106, opened 2017-10-04
Enforce usage of filter parameters

ticklemynausea:
The endpoint will return an http 400 if not enough filters are given.

I've made sure that this is compatible with the current usage of the
endpoint in the historical emissions section, but this can change
in the future


simaob:
Can you check the specs, they seem to be failing. Thanks!

PR #111, opened 2017-10-06
Feature/countries ghg map

j8seangel:
Map sdg emissions component:

- Move all the meta needed to provider so we can use it in all components easily.
- Request its own data and filter by year.
- Get new centroid data in the locations request to be able to centre in the selected country.
- Filter data by current year

To improve:
- prepare the default values in the new provider to avoid doing it over and over in each component
- remove the static default values using the above params.
- set year from graph in the url to make it dynamic

Bugs to fix:
- color scale not working properly.
- with the new provider, the ghg-emissions doesn't render going from countries before.
- the map country colours is not getting correctly in first render so to test it go to `/countries` and select a country.

To think about:
- The height of the graph doesn't allow to see the legend at the same time so is hard to understand what colour means.

image
image


app/serializers/api/v1/historical_emissions/record_serializer.rb
```diff
@@ -3,6 +3,7 @@ module V1
     module HistoricalEmissions
       class RecordSerializer < ActiveModel::Serializer
         belongs_to :location
+        belongs_to :iso_code3
```
ticklemynausea: iso_code3 is not a relation but an attribute, so you should replace `belongs_to` with `attribute`.

it would perhaps make sense to show the iso code and location names under a `location` key.

j8seangel: I've changed to attribute, and I agree but let's put both inside the location key later as it means some other changes.

Thanks Mario!

PR #112, opened 2017-10-06
Fix hash links in anchor nav

edbrett:
We had a problem where when clicking the anchor link it was appending duplicates, and ignoring search queries. Now it doesn't.


j8seangel:
Good job !
I've fixed a small bug because we need that query param to save the state in the ndcs page.

PR #114, opened 2017-10-09
Add relation to location in ndc_target entity

ticklemynausea:
This will allow the API to show linkage data for locations
that do not yet have a full ndc imported. This is an ugly temporary hack
and should be reverted as soon as possible!

This change requires an update to the db and a re-import of linkage data. Please
remember to:

```
rails db:migrate
rails ndc_sdg_targets:import
```

### DIFF
--- a/app/controllers/api/v1/ndc_sdgs_controller.rb
+++ b/app/controllers/api/v1/ndc_sdgs_controller.rb
@@ -14,6 +14,23 @@ module Api
         render json: @ndc,
                serializer: Api::V1::NdcSdg::NdcSerializer
       end
+
+      # Temporary hack to render NDC-SDG linkages for countries
+      # that do not have full NDCs imported
+      def show_by_location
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/ndc_sdgs_controller.rb
```diff
@@ -18,11 +18,11 @@ def show
...
-        @location = Location
-          .includes(:ndc_targets)
-          .where(iso_code3: params[:code].upcase)
-          .first
-        unless @location.ndc_targets.length > 0
+        @location = Location.
+          includes(:ndc_targets).
+          where(iso_code3: params[:code].upcase).
+          first
+        if @location.ndc_targets.length.empty?
```
simaob: You made Rubocop happy, but made me sad :P `undefined method empty? for Integer` . You forgot to remove the `length` bit. I'll fix and merge.

PR #117, opened 2017-10-10
Feature/country map improvements

j8seangel:
- Fix map styles to fit in the viewport + make the chart size dynamic.
- Get scaled colours for the legend.
- Sync graph and map year on hover.
- Fix ghg emissions graph when already have metadata ready.

- Please review [this](https://github.com/Vizzuality/climate-watch/compare/feature/country-map-improvements?expand=1#diff-115166989ddbaca172eb7c3dac8cbd70R87) carefully. It improves the react-simple-maps performance a lot but we should be careful.

### DIFF
--- a/app/javascript/app/components/charts/stacked-area/stacked-area-component.jsx
+++ b/app/javascript/app/components/charts/stacked-area/stacked-area-component.jsx
@@ -15,9 +15,9 @@ import { format } from 'd3-format';
 
 class ChartStackedArea extends PureComponent {
   render() {
-    const { config, data } = this.props;
+    const { config, data, height } = this.props;
     return (
-      <ResponsiveContainer height={500}>
+      <ResponsiveContainer height={height}>
         <AreaChart
           data={data}
           margin={{ top: 0, right: 0, left: -20, bottom: 0 }}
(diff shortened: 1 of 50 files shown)


app/javascript/app/components/charts/stacked-area/stacked-area-component.jsx
```diff
@@ -62,7 +63,17 @@ class ChartStackedArea extends PureComponent {
...
-  data: PropTypes.array.isRequired
+  data: PropTypes.array.isRequired,
+  height: PropTypes.oneOfType([
+    PropTypes.number,
+    PropTypes.string // % accepted
+  ]).isRequired,
+  onMouseMove: PropTypes.func.isRequired
+};
+
+ChartStackedArea.defaultProps = {
```
edbrett: Nice! Thanks for updating :)

app/javascript/app/components/country-ghg-emissions/country-ghg-emissions.js
```diff
@@ -63,9 +73,17 @@ class CountryGhgEmissionsContainer extends PureComponent {
...
   handleSourceChange = category => {
-    this.updateUrlParam({ name: 'source', value: category.value }, true);
+    if (category) {
+      this.updateUrlParam({ name: 'source', value: category.value }, true);
+    }
   };
 
+  handleYearHover = throttle(data => {
+    if (data) {
+      this.updateUrlParam({ name: 'year', value: data.activeLabel });
```
edbrett: I know this works nicely, but is it needed in the URL as it is bound to the tooltip which will only show on hover? Would it be better in the state? Not sure this param needs to be shared?

app/javascript/app/components/country-ghg-map/country-ghg-map-selectors.js
```diff
@@ -22,20 +23,20 @@ const buckets = [
   '#1b4a75',
   '#163449'
 ];
+const steps = [0, 0.4, 0.8, 2, 4, 8, 16, 32, 64];
 let colorScale = null;
 function setScale(ranges) {
-  colorScale = scaleLinear()
+  colorScale = scalePow()
```
edbrett: Good find!

PR #119, opened 2017-10-10
Undo all changes done by feature/change-linkage-relation-to-location

ticklemynausea:
This PR undoes the one with the hack that got merged today.

I already added the new NDC documents l (169 documents for 163 countries) to the s3 bucket, so linkage information for more countries will be available under the proper data model.

This change requires an update to the db and a re-import of linkage data. Please
remember to:

```
rails db:migrate
rails ndcs:full:import ndcs:full:index
rails ndc_sdg_targets:import
```


simaob:
It seems that running the commands in that order doesn't work. It fails because when it tries to drop the ndcs:full there are targets and it's not cascading. Maybe worth adding the cascading to the database to ensure we can re-import OK?

simaob:
Are you going to add that in this PR?

ticklemynausea:
Yes. Sorry. I'll commit a fix by the end of the day. We can merge this later.

ticklemynausea:
Done!

PR #126, opened 2017-10-13
Feature/country sdg info

j8seangel:
Show info on countries sdg linkages on info btn click
kapture 2017-10-13 at 12 37 33

BONUS: thanks to @sorodrigo we have better source map in development!


j8seangel:
It is working well for me locally, did you run the import?
kapture 2017-10-16 at 13 03 23

PR #130, opened 2017-10-16
Add google analytics with React GA 

Bluesmile82:
Add google analytics with React GA 
- [x] Add google analytics to the root component
- [x] Add key


j8seangel:
The track event was trigger each time the location changes, for example when a query param changes... but this doesn't mean a page change and we don't want to track as a change so I added a simple check there. Also, we weren't tracking the first page.

I'll add the key to the import later, meanwhile could you explain a bit better what React GA does ? 
I mean, what is that `.set` method doing ?

Bluesmile82:
Very good point! React GA is just a wrapper in top of google analytics.js methods (ga function). Set is setting a value, in this case page, to send it to analytics. It might be duplicated with the pageview. It would be nice to see how it affect the ga console.

PR #135, opened 2017-10-17
Feature/sdg map api

j8seangel:
Connect map with API real data:

- [TODO](https://github.com/Vizzuality/climate-watch/compare/feature/sdg-map-api?expand=1#diff-43ec6a248f422016bae077f81202d5cfR13) with @Bluesmile82 the goal selection
- Check with @ticklemynausea why the API only gets a couple of countries

image

BONUS: intelligent map cache identifying when the path has changed


edbrett:
screen shot 2017-10-17 at 10 02 30

Eastern Australia is having a bad day. ..

PR #136, opened 2017-10-17
NDC SDG targets and goals list

Bluesmile82:
This PR manages the interaction with the NDS SDG menu showing a table of NDGs for each goal

- [x] Select a goal and change the parameter in the url
- [x] Show list of SDGs


j8seangel:
Still pending todo the hover interaction and the targets map data visualization.

PR #137, opened 2017-10-17
Fix bug happening in production

ticklemynausea:
Rails was not using the correct serializers for the
collections presented on the `/ndcs/sdgs` api endpoint.

This PR explicitly tells which serializers to be used in
the `has_many` relations of `NdcsSdgs` `MetaSerializer`.

To reproduce this weird behaviour locally, you can set
`config.eager_load = true` in `config/environments/development.rb`


j8seangel:
Nice catch! thanks Mario

PR #138, opened 2017-10-17
Return first category with indicators by default

Bluesmile82:
Return first category with indicators by default in NDC Content

### DIFF
--- a/app/javascript/app/components/ndcs-map/ndcs-map-selectors.js
+++ b/app/javascript/app/components/ndcs-map/ndcs-map-selectors.js
@@ -31,8 +31,8 @@ export const getIndicatorsParsed = createSelector(
 );
 
 export const getSelectedCategory = createSelector(
-  [state => state.categorySelected, getCategories],
-  (selected, categories = []) => {
+  [state => state.categorySelected, getCategories, getIndicatorsData],
+  (selected, categories = [], indicators) => {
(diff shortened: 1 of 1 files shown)


app/javascript/app/components/ndcs-map/ndcs-map-selectors.js
```diff
@@ -31,16 +31,21 @@ export const getIndicatorsParsed = createSelector(
 );
 
 export const getSelectedCategory = createSelector(
-  [state => state.categorySelected, getCategories],
-  (selected, categories = []) => {
+  [state => state.categorySelected, getCategories, getIndicatorsData],
+  (selected, categories = [], indicators) => {
     if (categories.length > 0) {
```
edbrett: We have been following a format for this. Something like
``` if(!categories || !categories.length || !selected) return null```

File: app/javascript/app/components/ndcs-map/ndcs-map-selectors.js [hunk 2]
```diff
@@ -31,16 +31,21 @@ export const getIndicatorsParsed = createSelector(
...
       if (selected) {
         const filtered = categories.filter(
```
edbrett: can we avoid ambiguous names for variables if we are mixing indicators and categories in one selector. 'categoriesFiltered' or similar.

PR #139, opened 2017-10-17
Fix/suggestive search from params

edbrett:
We had a problem where searching in the ndcs full content removed the params of the document, preventing multi-search and persistence. This was not cool. Super uncool. Now it is cooler than ever 😎 .

To fix this it needed a small refactor of how params were handled in the url and the selectors, but as a result the component is much more reusable.


j8seangel:
Looks good for me, it is much easier now.

However, there are a lot of suggestions useless as this country doesn't match with that, so let's discuss with @ticklemynausea how to filter that data later.

PR #149, opened 2017-10-18
Add attribution metadata slug to emissions metadata

ticklemynausea:


j8seangel:
I've included the changes in the frontend to use the source id instead of the hardcoded one and also added the metadata modal in the ghg-emissions page.

PR #151, opened 2017-10-19
Implement search by sector, goal and target in collection endpoint

ticklemynausea:
This might not be done yet, I need to talk with @edbrett first regarding how highlight indexes behave when multiple matches occur for the same string.


ticklemynausea:
Offline approval by @simaob

PR #155, opened 2017-10-20
Feature/ghg calculations

Bluesmile82:
image

- [x] Create dropdown and add selection to the url
- [x] Create Provider to fetch the wb extra country data
- [x] Create selector to filter the information
- [x] Calculate the values depending on the selected calculation
- [x] Handle countries that dont have calculation values


j8seangel:
Good work 👏

PR #157, opened 2017-10-23
climate vulnerability section

Bluesmile82:
This is a PR for the Climate vulnerability section a.k.a poverty-country section
- [x] Create new section in country page
- [x] Connect component to redux
- [x] Create adaptations provider
- [x] Add sections to the component
- [x] Create graphs for the widgets
- [x] Make all the info buttons work


app/javascript/app/components/country-climate-vulnerability/country-climate-vulnerability-styles.scss
```diff
@@ -0,0 +1,69 @@
...
+  color: $theme-color;
+  display: flex;
+  align-items: center;
+}
+
+.info {
+  flex-basis: 100%;
+}
+
+.infoText {
```
j8seangel: We are not using this selector, aren't we?

app/javascript/app/components/country-climate-vulnerability/country-climate-vulnerability-selectors.js
```diff
@@ -0,0 +1,83 @@
...
+  );
+};
+
+const getListData = (sections, iso, slug) => {
+  const items = sections.filter(section => section.slug.startsWith(slug));
+  const itemList = [];
+  if (
+    !items.some(item => item.values.some(country => country.location === iso))
+  ) {
+    itemList.push('No data');
```
j8seangel: Why don't return the empty array directly?

app/javascript/app/actions.js
```diff
@@ -1,6 +1,7 @@
 import { actions as countriesProvider } from 'providers/countries-provider';
 import { actions as regionsProvider } from 'providers/regions-provider';
 import { actions as ndcsSdgsMetaProvider } from 'providers/ndcs-sdgs-meta-provider';
+import { actions as adaptationsProvider } from 'providers/adaptations-provider';
```
j8seangel: Please update to use the new syntax so we don't need this file anymore.

app/javascript/app/components/info-icon/info-icon-component.jsx
```diff
@@ -0,0 +1,35 @@
...
+        })}
+        onClick={toggleInfo}
+      >
+        <Icon icon={infoIcon} />
+      </Button>
+    );
+  }
+}
+
+InfoIcon.propTypes = {
```
j8seangel: We don't need all of this props as we could pass our custom className to customize the styles.

app/javascript/app/components/country-climate-vulnerability/country-climate-vulnerability-component.jsx
```diff
@@ -0,0 +1,98 @@
+import React, { PureComponent } from 'react';
+import Proptypes from 'prop-types';
+import cx from 'classnames';
+
+import PrepLink from 'components/prep-link';
+import Button from 'components/button';
+import Card from 'components/card';
+import Icon from 'components/icon';
+import VulnerabilityGraph from 'components/vulnerability-graph';
+import infoIcon from 'assets/icons/info.svg';
```
j8seangel: I like the idea of having the info icon only if we use it

app/javascript/app/components/prep-link/prep-link-component.jsx
```diff
@@ -0,0 +1,23 @@
+import React, { PureComponent } from 'react';
+import prepLogo from 'assets/icons/prep-logo.svg';
+import Icon from 'components/icon';
+import styles from './prep-link-styles.scss';
+
+class PrepLink extends PureComponent {
```
j8seangel: I don't like the idea of creating a component just for that... at least we could create a component that accepts the image and content as props or include it in the poverty folder.
But having a general component just for this I don't think is very useful.

app/javascript/app/components/info-icon/info-icon-component.jsx
```diff
@@ -0,0 +1,35 @@
+import React, { PureComponent } from 'react';
```
j8seangel: Could we move this component to the buttons folder? There are so many components in the root that is hard to find what you are looking for.

app/javascript/app/components/vulnerability-graph/vulnerability-graph-component.jsx
```diff
@@ -0,0 +1,109 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import CircularChart from 'components/circular-chart';
+import InfoIcon from 'components/info-icon';
+import { COUNTRIES_NUMBER } from 'data/country-codes';
```
j8seangel: I'd prefer to calculate this number dynamically with the number of countries in the request.
Is that possible? Maybe in the action ?

app/javascript/app/components/vulnerability-graph/vulnerability-graph-component.jsx
```diff
@@ -0,0 +1,109 @@
...
+    ) : (
+      'No data'
+    );
+  }
+
+  renderLine() {
+    const { sectionData } = this.props;
+    const value =
+      sectionData.data && Math.floor(sectionData.data.value * 10) / 10;
+    const riskValue = parseInt(sectionData.data.rank.absolute, 10) * 100;
```
j8seangel: When we have more than 100% the line is out of the graph, could we add a limit a 100% limit?
image

app/javascript/app/components/country-climate-vulnerability/country-climate-vulnerability-selectors.js
Bluesmile82: It is maybe different having no vulnerabilities or having no data about a country. We will have to check the final data.

app/javascript/app/components/vulnerability-graph/vulnerability-graph-component.jsx
Bluesmile82: We shouldn't have more than 100% 😮  That's pretty risky

app/javascript/app/components/vulnerability-graph/vulnerability-graph-component.jsx
```diff
@@ -69,7 +69,7 @@ class VulnerabilityGraph extends PureComponent {
         </div>
       </div>
     ) : (
-      'No data'
+      <div className={styles.noData}>No data</div>
```
j8seangel: We already have a component called `NoContent` for that kind of this, could you please use it?

PR #158, opened 2017-10-23
Feature/import unfccc data

ticklemynausea:
Added the `/timeline` endpoints to power the documents timeline. Two new endpoints exist:

- `/timeline` returns an array of date-ordered document objects from all available locations. In this endpoint each object contains a `location` key. sample: http://termbin.com/396z2

- `/timeline/:iso_code` returns an array of date-ordered document objects from the given location, or a 404 otherwise. In this endpoint the `location` key is hidden. sample: http://termbin.com/wckd


j8seangel:
I don't know too much about the rails code, looks good to me as I've imported the data, tested the endpoints and is working very good.
Thanks!

ticklemynausea:
Thanks! Perhaps @simaob can take a quick look, with regards to the rails code.

PR #163, opened 2017-10-25
Force source key to lowercase and strip spaces

ticklemynausea:
This PR forces the conversion of dataset names to lowercase and strips any leading/trailing spaces. Since this is done on the import process, please run `rails wri_metadata:import` after merging so that changes become visible.


j8seangel:
Thanks !

PR #167, opened 2017-10-26
Removes initiators from footer from all sections.

simaob:
This PR removes the NDC Partnership and WRI references from the footer. It basically makes it work the same across the site, as it was working on the about page.

BC request: https://basecamp.com/1756858/projects/13795275/todos/329135659

Hope I didn't break it! =D


simaob:
Thanks! =)

PR #170, opened 2017-10-26
Add teaser for sectors page

Bluesmile82:
- [x] Sectors teaser
- [x] Compare countries teaser
---
Texts are missing


j8seangel:
We have another teaser page, could you please add it ?
https://zpl.io/bJErzdE

app/javascript/app/components/teaser/teaser-component.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+class Teaser extends PureComponent {
+  // eslint-disable-line react/prefer-stateless-function
+  render() {
+    const { screenshot, title, description } = this.props;
+    return (
+      <div className={styles.wrapper}>
+        <Section className={styles.section}>
+          <div className={styles.column}>
+            <img
```
j8seangel: I see a border button that doesn't exist in the design 
image
It is very easy adding a margin negative in the img and a overflow hidden in the parent component

PR #171, opened 2017-10-26
Small text change and database improvement

simaob:
- Small text change on readiness module of country pages
- Adds cascade to wb_extra_country_data table to avoid derps on database cleaning task.


j8seangel:
Tested and works locally so lgtm

PR #173, opened 2017-10-27
Feature/about page and text changes

Bluesmile82:
- [x] About page tabs
- [x] About page partners update
- [ ] Fix text in countries page

image
image
image


app/javascript/app/components/about-partners/about-partners.js
```diff
@@ -0,0 +1,183 @@
...
+  partners: [
+    {
+      link: 'http://www.bmub.bund.de/en/',
+      img: {
+        alt: 'FMENCBNS',
+        src: fmencbnsImage
+      }
+    },
+    {
+      link: 'https://www.google.es',
```
simaob: Maybe we should use google.com :)

app/javascript/app/routes.js
```diff
@@ -162,10 +164,53 @@ export default [
...
+          },
+          {
+            path: '/about/partners',
+            component: AboutPartners,
+            exact: true,
+            anchor: true,
+            label: 'Climate Watch Partners'
+          },
+          {
+            path: '/about/contact',
```
simaob: I think we might have to hide the tabs that have no content yet. Let's check this on Monday morning.

simaob: @Bluesmile82 let me know when you're online and we can go over the other tabs and see which ones to add.

simaob: Looking at the file let's add the "Permissions & Licensing" Tab, as we seem to have the right content for that. I would also add the Contact Us and Feedback one.

We can use the same form that we are using here: http://vizzuality.github.io/climate-watch-splash , (click the sign up button). So it's an iframe showing up in a modal window.

app/javascript/app/components/about-description/about-description-component.jsx
```diff
@@ -0,0 +1,72 @@
...
+    </div>
+    <div>
+      Climate Watch brings together dozens of datasets for the first time to let
+      users analyze and compare the Nationally Determined Contributions (NDCs)
+      under the Paris Agreement, access historical emissions data, discover how
+      countries can leverage their climate goals to achieve their sustainable
+      development objectives, and use models to map new pathways to a lower
+      carbon, prosperous future.
+    </div>
+    <div>
```
j8seangel: Semantically makes sense to use a <p> element instead the <div> here.

app/javascript/app/components/about-partners/about-partners.js
j8seangel: 👍

app/javascript/app/routes.js
j8seangel: I agree with Simao hiding the empty sections... should be enough setting the anchor to false in the routes config.
Related to the Permissions and licensing we should propose to move this into the footer
image

simaob:
The text for the Licensing is quite big, so we do need a tab to show it, I believe.

I just talked with the guys from WRI the tabs that we should have at the moment are:

- About
- Partners
- Contact us
- Permissions & Licensing

On the contact us they have some small adjustments to the text, but could we get the About, Partners and Permissions & Licensing done first? And then we can update the Contact us section.

PR #175, opened 2017-10-27
Fix/html in ndc content

Bluesmile82:
[Render HTML when rendering table view](https://www.pivotaltracker.com/story/show/151612564)
- Render HTML if the table has the parseHtml prop
- Scroll if the content overflows

image

### DIFF
--- a/app/javascript/app/components/table/table-styles.scss
+++ b/app/javascript/app/components/table/table-styles.scss
@@ -20,6 +20,8 @@
 
   .ReactVirtualized__Table__row {
     border-bottom: solid 1px #e5e5eb;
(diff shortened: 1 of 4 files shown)


app/javascript/app/components/table/table-styles.scss
```diff
@@ -26,6 +26,10 @@
     color: $theme-color;
     font-size: $font-size-sm;
 
+    div {
+      overflow: scroll;
```
edbrett: I cannot see this working. Please review the position of the content for Mitigation > Sectors covered. The flexbox positioning is rendering cenetered vertically without scroll.

PR #177, opened 2017-10-27
Use calculation for map and pass search prop to force rerender

Bluesmile82:
[Country page's map's legend should update with the selection of calculation on the graph.](https://www.pivotaltracker.com/story/show/152344219)

- Use calculation to get correct legend for the map
- Pass search to the GHGCountryMap component to force rerender when the calculation is changed

image

### DIFF
--- a/app/javascript/app/components/country-ghg-map/country-ghg-map-selectors.js
+++ b/app/javascript/app/components/country-ghg-map/country-ghg-map-selectors.js
@@ -1,7 +1,7 @@
 import { createSelector } from 'reselect';
 import { scalePow } from 'd3-scale';
 import isEmpty from 'lodash/isEmpty';
-import CALCULATION_OPTIONS from 'app/data/constants';
+import { CALCULATION_OPTIONS } from 'app/data/constants';
 
 import worldPaths from 'app/data/world-50m-paths';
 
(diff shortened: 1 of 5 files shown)


app/javascript/app/components/country-ghg-map/country-ghg-map-selectors.js
```diff
@@ -146,10 +147,22 @@ export const getPathsWithStyles = createSelector([getDataParsed], data => {
...
-  title: `GHG Emissions per capita in ${year}`,
-  buckets
-}));
+export const getLegendData = createSelector(
+  [getCalculationSelection, getYearSelected],
+  (calculation, year) => {
+    let calculationText = '';
+    if (calculation === 'PER_CAPITA') {
+      calculationText = 'per capita ';
+    } else if (calculation === 'PER_GDP') {
```
edbrett: Do you need this? Why not set it as the default and only do one if statement?

simaob: We have one value that is the empty value. When it's not PER_CAPITA or PER_GDP.

j8seangel: I agree, we could use this: https://github.com/Vizzuality/climate-watch/blob/fix/countries-ghg-map-update-legend/app/javascript/app/components/country-ghg-emissions/country-ghg-emissions-selectors.js#L41 a global variables so we don't have strings comparisons. 
what about export a const with the values ?

Bluesmile82:
Ready!

j8seangel:
Fixed a crash when there isn't a default calculation selected and exported the constants in a namespace

PR #179, opened 2017-10-30
Update import process to latest ver. of dataset

ticklemynausea:
This version of the dataset includes the absolute ranks of values

There are no db schema changes. You must re-import the dataset with `rails adaptation:import`.

### DIFF
--- a/app/javascript/app/components/vulnerability-graph/vulnerability-graph-component.jsx
+++ b/app/javascript/app/components/vulnerability-graph/vulnerability-graph-component.jsx
@@ -44,7 +44,7 @@ class VulnerabilityGraph extends PureComponent {
   renderLine() {
     const { sectionData, countriesCount } = this.props;
     const value =
-      sectionData.data && Math.floor(sectionData.data.value * 10) / 10;
+      sectionData.data && Math.floor(sectionData.data.value * 100) / 100;
     const riskValue =
       sectionData.data && parseInt(sectionData.data.rank.absolute, 10) * 100;
     const riskAbsoluteValue = riskValue / countriesCount;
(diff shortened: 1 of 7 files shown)


app/javascript/app/components/vulnerability-graph/vulnerability-graph-component.jsx
```diff
@@ -65,7 +65,7 @@ class VulnerabilityGraph extends PureComponent {
               {sectionData.data.rank.absolute}
               <span className={styles.countriesNumber}>
                 {' '}
-                of {countriesCount}
```
simaob: As part of this task: https://www.pivotaltracker.com/story/show/152389076 , and the fact that the ranking is not over the total number of countries I'm now obtaining this value from the API. This means that the "countriesCount" is not needed for this component, as far as I can tell. So maybe someone that knows Javascript could come here and remove where needed.

app/serializers/api/v1/adaptation/value_serializer.rb
```diff
@@ -13,7 +13,9 @@ def location
         def rank
           {
             absolute: object.absolute_rank,
-            relative: object.relative_rank
+            relative: object.relative_rank,
+            maximum: ::Adaptation::Value.where(variable_id: object.variable_id).
```
simaob: @ticklemynausea let me know if this looks weird to you =)

app/javascript/app/components/vulnerability-graph/vulnerability-graph.js
```diff
@@ -3,5 +3,4 @@ import { actions } from 'components/modal-metadata';
 
 import Component from './vulnerability-graph-component';
 
-const mapStateToProps = () => {};
-export default connect(mapStateToProps, actions)(Component);
+export default connect(null, actions)(Component);
```
simaob: Thanks, Álvaro! =)

Is it looking good otherwise?

PR #182, opened 2017-10-30
Add contact form iframe and update text

Bluesmile82:
- Add iframe with contact form with fixed height
- Update text of contact tab

image


simaob:
the submit button is actually on WRI's side :( maybe they'll update it... we had sent some colours :s

PR #186, opened 2017-10-30
Uses document_type and language to enable switching from docs of the same type

simaob:
This PR fixes the following bug: https://www.pivotaltracker.com/story/show/152392874

The bug prevented us from switching between ndc documents when a country had multiple.
I've used both the document_type and the language to set the value of the dropdown, and now we can happily switch between multiple NDCs. We just got to wait for them to start having multiple NDCs on the same language.... =D but I think that's not going to happen, because they told us that it wouldn't. =)

## TEST

Go to the China full ndc page and switch between both NDCs. =) http://localhost:3000/ndcs/country/CHN/full


j8seangel:
Good work! Just added a filter to remove duplicates, thanks!

PR #187, opened 2017-10-30
Include country name tooltip by default in map

j8seangel:
Show country name tooltip by default in maps that could be overwritten passing an external tooltipId.

### DIFF
--- a/app/javascript/app/components/map/map-component.jsx
+++ b/app/javascript/app/components/map/map-component.jsx
@@ -7,6 +7,7 @@ import {
   Geographies,
   Geography
 } from 'react-simple-maps';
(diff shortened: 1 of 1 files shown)


simaob:
I've marked this PR as "don't-merge-yet", as the names that we have in the geometries are not the best fit... For instance "Kingdom of Spain" and "Republic of Portugal".

j8seangel:
Yes, I know we are cleaning the topojson file that fix will be included inside this task https://www.pivotaltracker.com/story/show/152374938

j8seangel:
Names updated in the PR https://github.com/Vizzuality/climate-watch/pull/189

simaob:
Thanks! =) good names now!

PR #206, opened 2017-10-31
Feature/ghg emissions review

j8seangel:
https://www.pivotaltracker.com/n/projects/2083759/stories/152432295

- [x] Removed subsectors.
- [x] Include a line for total values.
- [x] Change graph type on calculations change.
- [ ] Update tooltip label on calculations change.


simaob:
I see a small issue, when we have values that are much lower than 0.0t the graph doesn't show the data properly, as the yy axis is a bit funky. See attached:

screen shot 2017-10-31 at 18 49 57

But otherwise looks good with what is built! I'm going to merge this version and get it deployed. And maybe you can check the yy axis with the remaining task of the legend.

PR #210, opened 2017-11-01
Feature/consume rss feed

simaob:
## Overview

This PR adds the basic support to consume stories from WRI's RSS feed and to share these via our own API.

## What it includes 

- [x] create model
- [x] create migration
- [x] create import script
- [x] create rake task to use import script
- [x] create controller to expose the data
- [x] added specs for the controller

## Outstanding for another time

This overall task will still require some work once they update their feed, we will need to:

- [ ] add pagination support (hopefully)
- [ ] parse the data based on the tags of the elements
- [ ] hook up the background image which is currently missing from the feed
- [ ] setup a cron job to consume stories every so often

## To test

Run `rails db:migrate` and `rails stories:import`, and then visit the endpoint `GET /api/v1/stories`, you should see some stories. =)

Feel free to merge if it looks good.

### DIFF
--- a/app/services/import_stories.rb
+++ b/app/services/import_stories.rb
@@ -2,10 +2,8 @@ class ImportStories
   require 'rss'
   require 'open-uri'
 
-  def call reset=false
-    if reset
-      cleanup
-    end
+  def call reset = false
+    cleanup if reset
 
     import_stories
   end
@@ -18,7 +16,7 @@ class ImportStories
 
   def import_stories
     existing = Story.count
-    url = "http://feeds.feedburner.com/WRI_News_and_Views.rss"
+    url = 'http://feeds.feedburner.com/WRI_News_and_Views.rss'
     rss = open(url)
     feed = RSS::Parser.parse(rss)
     puts "Channel Title: #{feed.channel.title}"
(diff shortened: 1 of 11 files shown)


simaob:
@ticklemynausea can you review this when you have the chance, please. =) The outstanding items will be done in a different task.

lib/tasks/stories.rake
```diff
@@ -0,0 +1,19 @@
...
+  task import: :environment do
+    puts '##############################'
+    puts '# Starting to import stories #'
+    puts '##############################'
+    ImportStories.new.call
+    puts '########### ENDED ############'
+  end
+
+  desc 'Delete stories and re-import'
+  task fresh_import: :environment do
```
ticklemynausea: I'd add a ref to this in the main `db:import` task

app/services/import_stories.rb
```diff
@@ -0,0 +1,35 @@
+class ImportStories
+  require 'rss'
+  require 'open-uri'
+
+  def call reset = false
```
ticklemynausea: For consistency with the rest of the app, I'd use parenthesis around the args in the method definition

simaob: Agreed!

lib/tasks/stories.rake
simaob: That makes sense.

PR #212, opened 2017-11-02
Fix/country map interactions

Bluesmile82:
Calculation data was not being used on the country map. The map was not showing the real difference between absolute values and per capita or per GDP.
We need to import all the countries wb_extra data to be able to show them with the calculations  in the map.

- [x] Create a WB data index endpoint
- [x] Update the provider to parse all the countries data
- [x] Update calculation for all the countries in the map selector

image


j8seangel:
⚠️  I've found a bug! 
The map is not being coloured in the first load when the metric is the absolute value, could you please review it?
image

PR #215, opened 2017-11-02
Fix/sort sdg card dots

Bluesmile82:
Correctly sort dots

Created compareIndexByKey method in utils which can be reused for some other sort cases

e.g array.sort(compareIndexByKey('objectKey'))

1.1 < 1.2 < 1.10 < 1.a
image

### DIFF
--- a/app/javascript/app/utils/utils.js
+++ b/app/javascript/app/utils/utils.js
@@ -10,27 +10,30 @@ export function isCountryIncluded(countriesIncluded = [], iso) {
   return countriesIncluded.includes(iso);
 }
 
-export function compareIndex(a, b) {
-  const divideByDot = /(.*)\.(.*)/;
-  const isNotNumber = /[^0-9.]/;
-  const match = x => x.number.match(divideByDot);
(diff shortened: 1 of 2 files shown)


app/javascript/app/utils/utils.js
```diff
@@ -10,6 +10,26 @@ export function isCountryIncluded(countriesIncluded = [], iso) {
   return countriesIncluded.includes(iso);
 }
 
+export function compareIndex(a, b) {
```
edbrett: It would be great if we could make this more reusable for different data structures. Simply be passing in the data key you want to access the sdg number on (i.e. 'value' or 'number') and then comparing and returning the sorted list. This would mean we can reuse in other parts of the site where we aren't necessarily using the targets data structure.

Bluesmile82: Good idea!

edbrett: Something like lodash/sortBy would be great.

simaob:
Ed, this should be ready for you to re-review =)

PR #216, opened 2017-11-03
Display country description only if data available

Bluesmile82:
- Add check for GDP and population data independently

Andorra is alive again ⛷


simaob:
Much better! =)

PR #217, opened 2017-11-03
Fix: UNFCCC gas sectors

edbrett:
Due to UNFCCC not containing a WORLD filter category for regions, we were unable to correctly select its data when breakingBy sectors or gases. Now that this is possible on the backend this PR updates the GHG emissions graph to handle this:
- when breaking by sector or gas for UNFCCC send 'ANNEXI' to the API as location
- Due to variations in data available for GWP versions between data sets, this selectors values is no longer calculated from the meta but the data (to make it view specific on selector change)
- Add a loader and no content feedback views to the component


simaob:
Looks like it works! =) Thanks, Ed.

Álvaro if after your review you're happy with this feel free to merge. Functionality wise seems to do it!

simaob:
Hey guys, is this ready to be merged?

simaob:
Ready to merge, Álvaro? =)

PR #218, opened 2017-11-03
Change gdp scale

Bluesmile82:
https://www.pivotaltracker.com/story/show/152503610

GDP was shown in $ instead of million $ the chart lines weren't even showing because the value was too low.

We could also change this scale in the import of the data in the backend. Only if we are not going to use dollars in some other place. That will need a reimport.

### DIFF
--- a/app/javascript/app/components/country-ghg-emissions/country-ghg-emissions-selectors.js
+++ b/app/javascript/app/components/country-ghg-emissions/country-ghg-emissions-selectors.js
@@ -203,7 +203,8 @@ export const filterData = createSelector(
 
 const calculatedRatio = (selected, calculationData, x) => {
   if (selected === CALCULATION_OPTIONS.PER_GDP.value) {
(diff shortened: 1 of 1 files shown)


app/javascript/app/components/country-ghg-emissions/country-ghg-emissions-selectors.js
```diff
@@ -203,7 +203,8 @@ export const filterData = createSelector(
 
 const calculatedRatio = (selected, calculationData, x) => {
   if (selected === CALCULATION_OPTIONS.PER_GDP.value) {
-    return calculationData[x][0].gdp;
+    // GDP is in dollars and we want to display it in million dollars
+    return calculationData[x][0].gdp / 1000000;
```
edbrett: This already exists as a constant at the top of the file. Please use this.

simaob:
Well this was the one where I meant to say "much better", but the other one was good too. Thanks!

PR #219, opened 2017-11-03
Feature/country cuantifications

j8seangel:
Include quantifications points to the country ghg emissions graphs.
image

TODO: improve the tooltip and the points interaction.


simaob:
Yes! =D

simaob:
There seems to be a conflict here. Can it be addressed and the branch merged? Thanks!

j8seangel:
Fixed about and country map graph and map styles in IE and FF

PR #221, opened 2017-11-03
Fix: sorting of lists + bug fixes

edbrett:
A whirlwind of a PR to wrap up some much needed love and attention to styles and interaction whilst getting all that pesky data in order. Thanks to @juancarlosalonso for going through everything with me.
- Sort ordering: all list of data (accordions, selectors, SDGs goals, etc etc) are now in alphabetical order
- Fix: SDG sorting function wasn't able to handle certain cases. Now it always works.
- Now we are showing the regions in the GHG breakBy Regions selector (note: API update needed to handle correct data rendering)
- Styles fixes for home page video placeholder
- Loader where loaders should be
- Refactor buttons and selectors to correctly and consistently select styles and themes
- Correct no results feedback on NDCs content when searching
- Fix: logo color was wrong
- Search map not populating when no document in search -> now it does
- Fix: disable button logic not working on search

BONUS: some lovely transitions for buttons and selectors to add some quality to the site.


j8seangel:
Fixed conflicts and thanks!
It looks much better 👏

PR #223, opened 2017-11-05
Fix/general fixes and bugs

edbrett:
An brush up of the site addressing any outstanding bugs or style issues.

Bugs:
- Fix: selectors for NDC Compare selectors has unwanted box shadow when hovering, NDCs autocomplete selectors were missing a border
- Fix: country pages map height and padding weren't matching the design
- Fix: buttons hover interactions and disabled buttons (especially for groups) wasn't working correctly
- Fix: NDCs table selectors styles were wrong
- Fix: NDCs content overview height of the cards was broken
- Fix: sticky nav anchor links on country pages were not scrolling to correct place

Updates:
- Loader for NDCs full content view (only on first load)
- Set UNFCCC as default on NDCs map view

### DIFF
--- a/app/javascript/app/components/button-group/button-group-component.jsx
+++ b/app/javascript/app/components/button-group/button-group-component.jsx
@@ -19,7 +19,11 @@ const ButtonGroup = ({ className, onInfoClick, disabled }) => (
       className
     )}
   >
-    <Button className={cx(styles.button, styles.first)} onClick={onInfoClick}>
+    <Button
+      className={cx(styles.button, styles.first)}
+      onClick={onInfoClick}
+      disabled={!onInfoClick}
+    >
       <Icon icon={iconInfo} />
     </Button>
     <Button className={styles.button} disabled>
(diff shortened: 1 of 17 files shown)


app/javascript/app/components/button-group/button-group-component.jsx
```diff
@@ -11,8 +11,14 @@ import PropTypes from 'prop-types';
...
-  <div className={cx(styles.buttonGroup, className)}>
+const ButtonGroup = ({ className, onInfoClick, disabled }) => (
+  <div
+    className={cx(
+      styles.buttonGroup,
+      disabled ? styles.disabled : '',
+      className
+    )}
+  >
     <Button className={cx(styles.button, styles.first)} onClick={onInfoClick}>
```
Bluesmile82: We may use a  disabled={!onInfoClick} here to show the not allowed hover also for if the info button is not enabled

edbrett: I am not seeing these at all on my branch, and I remember fixing them in a previous PR. Could you delete the branch and re pull (using --prune, incase of duplicate name?

edbrett:
Fixed @Bluesmile82

PR #229, opened 2017-11-06
Calculate top position even on not table first level elements

Bluesmile82:
In the country full content page, some queries of the linkages were apparently highlighting the text but not scrolling to it, i. e: Goal 1: No poverty.
Actually what it was happening is that only the offset to the parent div was added to the scroll. So for list elements or so the top position was not the correct.

- Calculate top position even on not table first level elements

---
There are still some goals without linkages (e.g. Goal 2)


Bluesmile82:
Found a solution for the bug. Ready again!

j8seangel:
Good catch!
I'm still having some empty targets/goals for some countries but the behaviour is correct when we have a match so this is something we have to double check the data.

PR #230, opened 2017-11-06
Optimize searching by goal, sector and target

ticklemynausea:
Instead of retrieving all phrases associated with goals, sectors and targets and then searching for all of them using wildcard, search in ndc_targets texts, retrieve ndc_targets directly using the given params and the relations between models.

Instead of using regular expressions to find and replace the relevant parts with highlighted parts, we store the start and end position of each linkage quote upon import, and then use this information to create the highlighted text. 

Also we added an endpoint `/api/v1/ndcs/linkages_dataset` that returns the linkages dataset file `ndc_sdg_targets.csv`, with an extra column `starts_at`, which tells at which character position the linkage quote occurs in the original text. This can be used by the WRI to identify errors in the linkages dataset.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -64,16 +64,17 @@ Metrics/ModuleLength:
   Enabled: false
 
 Metrics/AbcSize:
+  Max: 20
   Exclude:
     - 'lib/modules/*'
 
 Metrics/MethodLength:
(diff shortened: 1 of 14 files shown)


edbrett:
A lot better! Well done!

PR #232, opened 2017-11-07
Feature/copy page link

Bluesmile82:
We had the copy link option in the share menu but it was not implemented

image

- Add copy Url and change icon for a check when the link is copied


edbrett:
You need to make sure the component resets after either a preiod of time or after a route change otherwise it is always ticked after one click.

app/javascript/app/components/simple-menu/simple-menu-component.jsx
```diff
@@ -13,46 +14,91 @@ class SimpleMenu extends PureComponent {
...
+  handleClick(action) {
+    this.setState({ actionSuccesful: action() });
+  }
+  renderInsideLink(option, withAction = false) {
+    const { actionSuccesful } = this.state;
+    return (
+      <div className={styles.documentLink} key={option.label}>
+        {option.icon &&
+          (withAction && actionSuccesful ? (
+            <Icon icon={checkIcon} className={styles.icon} />
```
edbrett: This is unnecessary. You can do the ternary inside the icon prop for more clarity and less code.

j8seangel:
I've fixed those changes, thanks Ed!

PR #233, opened 2017-11-07
Feature/regions for ghg emissions

edbrett:
Likely a confusing PR but very important. It contains two key features separated below.

### Update defaults for selectors 
Currently the GHG emissions graph is lying to you. It's deceit knows no bounds. This injustice must come to an end. I know its hard to believe, but the default selectors for each breakBy value are not correct. Currently we are using Energy for each one when we should be using a curated list of super excellent default values to give the correct overall values. To rectify this the defaults selector to return a specific list.

### Allow the selection of regions
Due to the world not being as simple sack of countries, regions are required to add greater understanding of our emissions data. This new functionality allows, when breaking by region, to select the top (max 10) emitters from each region to be displayed on the chart. It defaults to top world emitters. If a user selects additional countries this region is removed and the countries act again previously. If the user removes a region OR a selection the chart clears. Nice. To allow this quite a few changes were needed to the data parsing and the components to get the correct data and to improve performance sufficiently:
- Update how params are pushed to allow clearing of Filters selector
- If a region is selected get and show the region name in the selector
- Sort the data in two stages, to allow the selection of top emitters from the regions members list
- Add reselect nulls and uppdate components/container to handle these
- Simplify fetch logic
- Styles fixes to handle no data states for the chart

### BONUS ROUND
When exploring these changes I found the following bugs which have been addressed:
- Multi select overflow styles without elypsis
- Dropdown select overflow not showing elypsis
- Multi select jumping to beginning when props changing

### DIFF
--- a/app/javascript/app/components/ghg-emissions/ghg-emissions-selectors.js
+++ b/app/javascript/app/components/ghg-emissions/ghg-emissions-selectors.js
@@ -297,13 +297,11 @@ export const getActiveFilterRegion = createSelector(
 
 // Map the data from the API
 export const filterData = createSelector(
-  [getData, getFiltersSelected, getBreakSelected],
+  [filterAndSortData, getFiltersSelected, getBreakSelected],
   (data, filters, breakBy) => {
     if (!data || !data.length || !filters || !filters.length) return null;
     const filterValues = filters.map(filter => filter.label);
-    return sortEmissionsByValue(
-      data.filter(d => filterValues.indexOf(d[breakBy.value]) > -1)
-    );
+    return data.filter(d => filterValues.indexOf(d[breakBy.value]) > -1);
   }
 );
 
--- a/app/javascript/app/components/ghg-emissions/ghg-emissions-selectors.js
+++ b/app/javascript/app/components/ghg-emissions/ghg-emissions-selectors.js
@@ -23,7 +23,8 @@ const TOP_EMITTERS = [
   'BRA',
   'IDN',
   'CAN',
-  'MEX'
+  'MEX',
+  'TOP'
 ];
 
 const DATA_SCALE = 1000000;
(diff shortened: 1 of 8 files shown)


app/javascript/app/components/ghg-emissions/ghg-emissions.js
```diff
@@ -128,24 +130,37 @@ class GhgEmissionsContainer extends PureComponent {
...
   handleBreakByChange = breakBy => {
     const params = [
-      { name: 'breakBy', value: breakBy.value },
-      { name: 'version', value: '' },
-      { name: 'filter', value: '' }
+      { name: 'source', value: this.props.sourceSelected.value },
+      { name: 'breakBy', value: breakBy.value }
     ];
-    this.updateUrlParam(params);
+    this.updateUrlParam(params, true);
```
Bluesmile82: We lose the IPCC version when we change the break by. Not sure if its important. I think you do it because you want to get rid of all the countries for example

edbrett: Additionally I have just pushed a fix for the filter also.

app/javascript/app/components/ghg-emissions/ghg-emissions-selectors.js
```diff
@@ -212,38 +254,61 @@ export const getFiltersSelected = createSelector(
...
-    const sourceData = sources.find(d => d.value === sourceSelected.value);
-    return {
-      sector: sourceData.sector[0],
-      gas: sourceData.gas[0]
-    };
+  [getSourceSelected, getMeta],
+  (sourceSelected, meta) => {
+    if (!sourceSelected || !meta) return null;
+    const defaults = {};
+    switch (sourceSelected.label) {
```
j8seangel: crazy stuff working perfectly 👏

PR #235, opened 2017-11-07
Include quantifications interactions

j8seangel:
A picture is worth 1000 words:
kapture 2017-11-07 at 15 56 32


j8seangel:
Two pictures are worth 1000 words because I forgot to include the ranges:
kapture 2017-11-07 at 18 25 19

PR #236, opened 2017-11-07
Feature/nested sectors ndc accordions

edbrett:
Big update to the ndcs country and compare accordions to handle nested sectors (as opposed to nested indicators):
- Show the sectoral tabs again
- Fix: interactions on accordion have been simplified so now it always open and closes correctly when clicked
- New selectors to parsed sectors instead of indicators first in hierarchy
- Performance improvements for compare accordion by only rendering when needed

### DIFF
--- a/app/javascript/app/components/accordion/accordion-component.jsx
+++ b/app/javascript/app/components/accordion/accordion-component.jsx
@@ -16,7 +16,8 @@ class Accordion extends PureComponent {
       handleOnClick,
       openSlug,
       children,
-      isChild
+      isChild,
+      hasNestedCollapse
     } = this.props;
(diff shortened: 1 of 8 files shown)


app/javascript/app/components/accordion/accordion-component.jsx
```diff
@@ -82,7 +89,8 @@ Accordion.propTypes = {
     })
   ),
   children: PropTypes.node,
-  isChild: PropTypes.bool
+  isChild: PropTypes.bool,
+  hasNestedCollapse: PropTypes.bool
```
j8seangel: Good doc read!

PR #237, opened 2017-11-08
Includes multiple values for same year

j8seangel:
One more iteration over the quantifications interactions: 
- accepts multiple values for the same year
- group them when having the exact value and year

kapture 2017-11-08 at 11 56 36


edbrett:
screen shot 2017-11-08 at 12 36 43

j8seangel:
Good catch and fixed!

edbrett:
Awesome! Merge away!

PR #239, opened 2017-11-08
Add emission pathways header, intro and section anchor links

Bluesmile82:
Starts the emission pathways page with the header, intro, and links to different sections


Bluesmile82:
Cool! Disabled for now

PR #246, opened 2017-11-09
Feature/display disclaimer for translated ndc texts

ticklemynausea:
Adds support for the 'translated' NDC attribute. This attribute indicates if an NDC has been automatically translated. This attribute is inferred from the NDC filename (if the language field ends with `_TR`).

The front-end development was not based in any design specs: https://cl.ly/1R3z3n142f2V

You'll need to run:

`rails ndcs:full:import` to see it locally. And Andorra (AND) is a good country to test, after the import.

### DIFF
--- a/app/javascript/app/components/ndc-translation-disclaimer/ndc-translation-disclaimer-component.jsx
+++ b/app/javascript/app/components/ndc-translation-disclaimer/ndc-translation-disclaimer-component.jsx
@@ -1,0 +1,26 @@
+import React, { PureComponent } from 'react';
+import cx from 'classnames';
+
+import layout from 'styles/layout';
+import styles from './ndc-translation-disclaimer-styles.scss';
+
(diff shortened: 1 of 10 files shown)


simaob:
Is this already working? Which reimport do I need to run to test it?

simaob:
@juancarlosalonso this is how the disclaimer text is looking, any suggestions to improve the styles?


screen shot 2017-11-10 at 10 57 51

edbrett:
React stuff is spot on, so just need JC's review of styles.

juancarlosalonso:
I would only change the width of the lines. 
Let's apply a `max-width: 770px;` like we have on other texts.

edbrett:
I will sort it, as we are using a specific css grid set up for this.

edbrett:
Updated. Feel free to merge.

PR #247, opened 2017-11-09
Fix/fix rubocop and random refactors

ticklemynausea:
This pull request contains some refactors and a less strict version of rubocop.yml. Some rubocop checks have been disabled inline since there are no plans to refactor those parts yet. After merging, rubocop will not throw any errors or warnings.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -64,17 +64,17 @@ Metrics/ModuleLength:
   Enabled: false
 
 Metrics/AbcSize:
-  Max: 30
+  Max: 25
   Exclude:
     - 'lib/modules/*'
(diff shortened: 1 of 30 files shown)


.rubocop.yml
```diff
@@ -49,7 +49,7 @@ Style/RegexpLiteral:
   AllowInnerSlashes: true
 
 Metrics/LineLength:
-  Max: 80
+  Max: 100
```
simaob: this one does make me sad :( I like 80 columns.

simaob: though I can understand that it's a bit arbitrary these days =D

PR #251, opened 2017-11-10
Add parent sector information to accordion

ticklemynausea:
I decided to modify the accordion component in case they decide that a string format
of "parent | sector" isn't good enough and in that case we'd need to do custom styling.

An alternative could be to have a selector modify the title attribute of each sector,
in that case we wouldn't have to put the concern of checking for 'parent' attributes
inside the accordion which is something that might be a bit weird. Thoughts?

https://cl.ly/3P392K3H2C0v
https://cl.ly/133Y25470Q0g

bc thread: https://basecamp.com/1756858/projects/13795275/todos/328481533#comment_575648775

edit: wip because I broke sorting

edit: sorting fixed

### DIFF
--- a/app/javascript/app/components/accordion/accordion-component.jsx
+++ b/app/javascript/app/components/accordion/accordion-component.jsx
@@ -35,6 +35,9 @@ class Accordion extends PureComponent {
                 isOpen = false;
               }
             }
+            const title = section.parent
+              ? `${section.parent.name} | ${section.title}`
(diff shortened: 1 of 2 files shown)


simaob:
This looks like it works :) Ed if you can review it and check if it's all good, feel free to merge!

edbrett:
Looks good to me! If that is what they want then I'll merge.

PR #260, opened 2017-11-13
Feature/esp graph filters

edbrett:
A starting PR for the ESP filters. This adds the filters that are currently possible from the API, and filters the data a necessary:
- Add a models and indicators dropdown to view
- Filter data correctly based on selector choice
- Set defaults for locations, models, and indicators (indicators i currently disabled so that at least some data is visible).
- Add support for .env urls in the development environment
- Update providers to use .env variable

###ATTENTION: please add the 'ESP_API' key to you .env and give it a url (local or staging or whatever. I don't discriminate when it comes to APIs).

BONUS: Finally fix an annoying bug with 100% width and CSS grid on the dropdown selectors. Add fix for anchor links absolute positioning when scrolling.

NOTE: it seems that the data from the API is not what is expected by design. Year values are only until 2013 (or similar), there is very little data for most cases, and indicators very rarely match a model. It is possible that when applying an association between Models and Locations this will reduce the options more => leading to better data views, but I am skeptical. Please review with caution. We are still missing categories filtering also, which will only refine the empty data more.

### DIFF
--- a/app/javascript/app/providers/esp-indicators-provider/esp-indicators-provider-actions.js
+++ b/app/javascript/app/providers/esp-indicators-provider/esp-indicators-provider-actions.js
@@ -5,6 +5,7 @@ import isEmpty from 'lodash/isEmpty';
 const fetchEspIndicatorsInit = createAction('fetchEspIndicatorsInit');
 const fetchEspIndicatorsReady = createAction('fetchEspIndicatorsReady');
 const fetchEspIndicatorsFail = createAction('fetchEspIndicatorsFail');
+const { ESP_API } = process.env;
 
 const fetchEspIndicators = createThunkAction(
   'fetchEspIndicators',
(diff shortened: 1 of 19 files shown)


app/javascript/app/providers/esp-indicators-provider/esp-indicators-provider-actions.js
```diff
@@ -5,6 +5,7 @@ import isEmpty from 'lodash/isEmpty';
 const fetchEspIndicatorsInit = createAction('fetchEspIndicatorsInit');
 const fetchEspIndicatorsReady = createAction('fetchEspIndicatorsReady');
 const fetchEspIndicatorsFail = createAction('fetchEspIndicatorsFail');
+const { ESP_API } = process.env;
```
j8seangel: 👏

simaob:
If it looks good, could it be merged?

PR #263, opened 2017-11-14
Fix: correct defaults for GHG emissions in countries

edbrett:
Small PR to update the countries ghg emissions graphs and map to use the correct defaults for each source case.


edbrett:
Good point! I have moved to utils and updated the selectors! Thanks!

PR #268, opened 2017-11-16
Feature/show total emissions for country calc

edbrett:
When on country GHG emissions graph, and calculating by GDP or Capita, show only the sum of the sectors as total and not the individuals.

screen shot 2017-11-16 at 14 05 20


simaob:
the map doesn't seem to be updating when you hover these graphs.

edbrett:
It seems it wasn't working before either. I will investigate.

edbrett:
Fixed! It broke when we swapped to the LineChart as it was missing a debounce function

PR #271, opened 2017-11-17
Fix/save nicaragua

Bluesmile82:
Nicaragua was failing to load in countries page: 
- Nicaragua has no 'api/v1/ndcs/NIC/sdgs` data
- Ergo data was undefined but the overview component tried to access some child
- Also loading in ndcs sdg linkages was always on

So:
- [x] Added errors managing to some providers
- [x] Added fail safe conditions to country-ndc-overview-component

--- 

Note: Data must be fixed

### DIFF
--- a/app/javascript/app/components/country-ndc-overview/country-ndc-overview-component.jsx
+++ b/app/javascript/app/components/country-ndc-overview/country-ndc-overview-component.jsx
@@ -25,7 +25,9 @@ class CountryNdcOverview extends PureComponent {
       );
     }
     const hasCoverageSectorsShort =
-      values.coverage_sectors_short && values.coverage_sectors_short.length;
+      values &&
+      values.coverage_sectors_short &&
+      values.coverage_sectors_short.length;
 
     return (
       <div className={styles.wrapper}>
--- a/app/javascript/app/components/country-ndc-overview/country-ndc-overview-component.jsx
+++ b/app/javascript/app/components/country-ndc-overview/country-ndc-overview-component.jsx
@@ -24,6 +24,9 @@ class CountryNdcOverview extends PureComponent {
         />
       );
     }
+    const hasCoverageSectorsShort =
+      values.coverage_sectors_short && values.coverage_sectors_short.length;
+
     return (
       <div className={styles.wrapper}>
         <div className={layout.content}>
@@ -40,7 +43,9 @@ class CountryNdcOverview extends PureComponent {
                       'Overview'
                     )
                   }
-                  description={values.indc_summary[0].value}
+                  description={
+                    values.indc_summary[0] && values.indc_summary[0].value
+                  }
                   textColumns={textColumns}
                 />
                 {actions && (
@@ -67,7 +72,7 @@ class CountryNdcOverview extends PureComponent {
               <div className={styles.cards}>
                 <Card title="GHG Target">
                   <div className={styles.cardContent}>
-                    {values.ghg_target_type.length ? (
+                    {values.ghg_target_type && values.ghg_target_type.length ? (
                       <div>
                         <span className={styles.metaTitle}>Target type</span>
                         <p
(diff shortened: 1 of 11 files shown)


app/javascript/app/components/country-ndc-sdg-linkages/country-ndc-sdg-linkages.js
```diff
@@ -23,16 +23,20 @@ const mapStateToProps = (state, { match, location }) => {
     data: ndcsSdgsMeta.data,
     activeSector: search.sector
   };
+
   return {
-    fetched: ndcsSdgsData.data[iso],
+    fetched: ndcsSdgsData.data && ndcsSdgsData.data[iso],
```
edbrett: It will always have ndcsSdgsData.data as it is an object, so this is unnecessary.

app/javascript/app/providers/ndcs-documents-meta-provider/ndcs-documents-meta-provider-reducers.js
```diff
@@ -3,20 +3,22 @@ import groupBy from 'lodash/groupBy';
 export const initialState = {
   loading: false,
   loaded: false,
-  data: {}
+  data: {},
+  error: false
```
edbrett: As this is a meta provider, we aren't going to fetch per country. So it seems pointless to have an error param on the state. If there is an error set the data to {}. In the front end we are only going to show "no content" at this point. So this serves no purpose.

app/javascript/app/providers/ndcs-sdgs-meta-provider/ndcs-sdgs-meta-provider-actions.js
```diff
@@ -4,6 +4,7 @@ import isEmpty from 'lodash/isEmpty';
 
 const getNdcsSdgsMetaInit = createAction('getNdcsSdgsMetaInit');
 const getNdcsSdgsMetaReady = createAction('getNdcsSdgsMetaReady');
+const getNdcsSdgsMetaFail = createAction('getTimelineFail');
```
edbrett: Fail is the same as Ready, but with {} instead of data as the payload. Can we not reuse Ready in each of these situations?

app/javascript/app/components/country-ndc-overview/country-ndc-overview-component.jsx
```diff
@@ -24,6 +24,11 @@ class CountryNdcOverview extends PureComponent {
         />
       );
     }
+    const hasCoverageSectorsShort =
```
edbrett: It seems strange to me to do this for one value and not the others. Seems hacky to me. Can we not address the selectors or the data to prevent rendering undefined?

app/javascript/app/components/country-ndc-sdg-linkages/country-ndc-sdg-linkages.js
```diff
@@ -23,16 +23,20 @@ const mapStateToProps = (state, { match, location }) => {
...
     activeSector: getSectorSelected(sdgsData),
     tooltipData: countrySDGLinkages.tooltipData,
     sectors: getSectorsMapped(sdgsData),
     sectorOptions: getSectorOptionsSorted(sdgsData),
-    goals: ndcsSdgsMeta.data.goals || [],
+    goals: (!ndcsSdgsData.error && ndcsSdgsMeta.data.goals) || [],
```
edbrett: can't we just check for empty. Data is never reaching the component.

Bluesmile82:
Ok, ready again. We will have to discuss error handling in the providers at some point

app/javascript/app/components/country-ndc-sdg-linkages/country-ndc-sdg-linkages-component.jsx
```diff
@@ -55,20 +55,53 @@ class CountrySDGLinkages extends PureComponent {
     ) : null;
   }
 
-  render() {
+  renderCards() {
```
edbrett: I don't understand the reason for this function? It does that same as below with a different name. I dont mind too much but can we not call it render'something' as it is confusing with react render. We have been using getSomething elsewhere in the app I believe.

PR #275, opened 2017-11-20
Add API docs API.md

ticklemynausea:


API.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Locations
+
+### `GET /locations/countries`
+
+Lists all locations that are countries
+
+#### Query parameters
+
+* `topojson` - if present, the location's topojson structure will be sent with the object.
```
simaob: Should this be treated as a boolean? i.e.: `topojson=true`? Or it doesn't matter?

API.md
```diff
@@ -0,0 +1,204 @@
...
+
+Collection is serialized as an object with the metadata entity as key
+
+### `GET /ndcs/:iso/sdgs`
+
+Shows linkage data for the given country code.
+
+### `GET /ndcs/sdgs_overview`
+
+Shows an overview of linkage data to meet specific front-end requirements:e .
```
simaob: there's a trailing `:e .` I guess it's a vim artefact =)

API.md
```diff
@@ -0,0 +1,204 @@
...
+
+#### Query parameters
+
+* `location` - Filter by location iso code3. Accepts multiple values separated by commas.
+
+## Quantifications
+
+### `GET /quantifications`
+
+Lists quantification data
```
simaob: Suggest making it a bit more clear: "Lists quantification data, representing NDC emissions targets.

API.md
ticklemynausea: It doesn't matter. The only thing required is that the parameter is present.

API.md
ticklemynausea: Derp.

API.md
ticklemynausea: :+1:

PR #276, opened 2017-11-20
Feature/legend manager for charts

edbrett:
Due to the fact that scenarios could not be managed for the ESP graph, this PR creates the Legend Manager component detailed in the designs that we have put off developing:
- Move all tags and selector logic for charts active data to a new component -> LegendCharts
- Update all charts to use legend and allow data to be added and removed
- Selector logic needed to make the component work for Countries pages


app/javascript/app/components/country-ghg-emissions/country-ghg-emissions.js
```diff
@@ -110,6 +123,24 @@ class CountryGhgEmissionsContainer extends PureComponent {
     }
   };
 
+  handleRemoveTag = tagData => {
```
Bluesmile82: I was wondering, do you think this could be done inside the Legend Chart Component so we don't have to repeat the code?

app/javascript/app/components/charts/legend-chart/legend-chart-component.jsx
```diff
@@ -0,0 +1,65 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import MultiSelect from 'components/multiselect';
+import Icon from 'components/icon';
+import Tag from 'components/tag';
+
+import plusIcon from 'assets/icons/plus.svg';
+import styles from './legend-chart-styles.scss';
+
+class LegendChart extends PureComponent { // eslint-disable-line react/prefer-stateless-function
```
Bluesmile82: Nice to share this component thru all the graphs! 😎

app/javascript/app/components/country-ghg-emissions/country-ghg-emissions-component.jsx
```diff
@@ -70,30 +75,32 @@ class CountryGhgEmissions extends PureComponent {
...
-                key={`${column.value}`}
-                data={{
-                  color: config.theme[column.value].stroke,
-                  label: column.label,
-                  id: column.value
-                }}
-              />
-            ))}
+          }
+          {!loading && !data.length > 0 &&
```
Bluesmile82: Could we keep the add Icon when the selected data is empty? Not sure if someone is going to do but the fact that if you delete all the filters you cant go back unless you go back in history or modify the URL it is annoying.

Bluesmile82: Actually is just annoying in ESP as in GHG you can select them from the dropdown

edbrett: That was the desired behaviour but maybe it is not working. I will review.

app/javascript/app/components/country-ghg-emissions/country-ghg-emissions.js
edbrett: Yes it can, I wasn't sure which way I preferred as there were some differences at first. I think you're right, I will move it.

edbrett:
@Bluesmile82 please review again. I have made some larger changes to reduce code. Now there is a new component called Chart which handles all charts, with a conistent loader and no content module to stop us having to set it up everytime.

PR #279, opened 2017-11-21
Add strip whitespace converter to S3CSVReader

ticklemynausea:
- Add strip whitespace converter to S3CSVReader
- Change some importers to use S3CSVReader
- Remove individual usage of string.strip (S3CSVReader already does this)
- Clean dead code

To see changes re-import the data (`rails db:drop db:create db:migrate db:import`)


edbrett:
Works wonderfully for me!

PR #280, opened 2017-11-21
Feature/users

santostiago:
Added the resources for `My Climate Watch` as well as the authentication using the GFW Control Tower.


santostiago:
Two new environment variables were created to run it in development and testing modes.
`DEV_USER_ID` and `DEV_USER_TOKEN`.
There's also a task `db:user_creation` to create a default user in dev

PR #284, opened 2017-11-21
remove sort ordering of ndc content when selecting

edbrett:
Quick PR to remove sort ordering of NDC content for the accordions now that the API returns the correct order.


simaob:
I'm getting a javascript error on the Sectoral tab of the NDC Content. Do you have this too?

Bluesmile82:
Might have to check for section.sectors existence in ndcs-country-accordion-component (line 53)
I would do it myself but hound is getting crazy with the indentation

edbrett:
Correct! I have pushed a fix :)

PR #287, opened 2017-11-24
Feature/indicators trend line

Bluesmile82:
**Trend line for indicators in esp scenario show page**
- [x] Add provider for Indicators trend
- [x] Render all indicator trendlines
- [x] Filter by location

image


simaob:
Hi Álvaro there seem to be some conflicts that will prevent the merge, can you check it, please?

simaob:
Nice, I really like the trendlines. =) Good job. Ed can you check the code bits and merge if you're happy with it? tks!

app/javascript/app/components/emission-pathways-scenario-table/emission-pathways-scenario-table-selectors.js
```diff
@@ -33,8 +49,12 @@ const scenarioIndicatorsData = createSelector(
 
 export const getCategories = createSelector(scenarioIndicatorsData, data => {
   if (!data) return null;
+  const categories = [];
+  data.forEach(c => {
+    if (c.category) categories.push(c.category);
+  });
   return sortLabelByAlpha(
```
edbrett: excellent.

app/javascript/app/components/emission-pathways-scenario-table/emission-pathways-scenario-table-selectors.js
```diff
@@ -49,19 +69,29 @@ export const getSelectedCategoryOption = createSelector(
...
+    value: location.id.toString(),
+    label: location.name
+  }));
+});
+
+export const getSelectedLocationOption = createSelector(
+  [getLocations, getLocationSelected],
+  (locations, selectedLocation) => {
+    if (!locations) return null;
+    const location = locations.find(l => l.id.toString() === selectedLocation);
```
edbrett: Can you not just filter from the getLocationsOptions, so that you dont have a selected value before you have your locations? As you have done with categories before?

app/javascript/app/components/table/table-component.jsx
```diff
@@ -71,6 +79,10 @@ class SimpleTable extends PureComponent {
                     flexGrow={flexGrow}
                     cellRenderer={cell => {
```
edbrett: Looking to the future, this is the sort of thing that should be passed in by the parent. Custom render functions should be exposed by our extended components like Table so that you can pass this from the parent and the table can be reused in other projects without knowing what a TrendLine is. This can be done easily by spreading the parent props right after the customised props. No problem for now, but certainly other tables shouldn't be worrying about trendlines. What do you think?

app/javascript/app/providers/esp-indicators-trend-provider/esp-indicators-trend-provider-reducers.js
```diff
@@ -0,0 +1,35 @@
...
+    const locationId = payload.locationId;
+    const scenarioId = payload.scenarioId;
+    const newState = {
+      ...state,
+      data: {
+        locations: {
+          ...state.data.locations,
+          [locationId]: {
+            scenarios: {
+              ...(state.data.locations && state.data.locations[locationId]
```
edbrett: Is this correct? Spreading a conditional?

PR #290, opened 2017-11-29
Add overview model modal

Bluesmile82:
Modal displaying all the data from a model in ESP model show page:

- Create a modal overview component (a little bit different from the modal-metadata)
- Select and pass the data from the overview component
- Blacklist undesired attributes
- Pass the full name as the title
- Remove the Full name in the attributes as it is included in the title

image


Bluesmile82:
Ready again! Thanks!

PR #291, opened 2017-11-30
Fix/general fixes

Bluesmile82:
- [x] Update all graph axes to match designs:
  - Force 0 in first tick or y-axis
  - Margin-top on x-axis
- [x] Review all filter changes for location history
- [x] Countries menu - Recenter the map to show New Zealand, now hidden
- [x] Country>Climate Vulnerability and Readiness - Center "No data" label
- [x] Homepage - Review spacing between elements, now is too tight. The grid should have `grid-column-gap: 30px`
- [x] Fix react warning in Linkages page


app/javascript/app/components/countries-menu/countries-menu-actions.js
```diff
@@ -0,0 +1,7 @@
+import { createAction } from 'redux-actions';
+
+const openCountriesMenu = createAction('openCountriesMenu');
```
edbrett: I think we will need to use component state for this for performance. With a simple state show:true/false to change a class which affects the opacity. I am finding it very slow.

app/javascript/app/components/countries-menu/countries-menu-styles.scss
```diff
@@ -0,0 +1,43 @@
+@import '~styles/layout.scss';
+
+.fixed {
+  position: fixed;
```
edbrett: You cant used fixed for something like this. Try scrolling down the page to see the effect. It needs to be absolute, or cleverly made relative to the top bar.

app/javascript/app/components/map/map-component.jsx
```diff
@@ -58,14 +57,20 @@ class Map extends PureComponent {
           }}
           style={{
             z: spring(zoom, { stiffness: 240, damping: 30 }),
-            x: spring(center[0], { stiffness: 240, damping: 30 }),
-            y: spring(center[1], { stiffness: 240, damping: 30 })
+            x: spring((customCenter && customCenter[0]) || center[0], {
```
edbrett: handy!

app/javascript/app/pages/country/country-component.jsx
```diff
@@ -18,35 +19,37 @@ class Country extends PureComponent {
...
-            <AnchorNav links={anchorLinks} className={layout.content} />
-          </Sticky>
-        </Header>
-        {route.sections &&
-          route.sections.length > 0 &&
-          route.sections.map(section => (
-            <div key={section.hash} className={styles.section}>
-              <div id={section.hash} className={styles.sectionHash} />
-              <section.component />
+      <ClickOutside onClickOutside={() => this.handleClickOutside(false)}>
```
edbrett: Why is this needed?

app/javascript/app/styles/layout.scss
```diff
@@ -10,7 +10,7 @@
 
 %grid {
   display: grid;
-  grid-column-gap: $gutter-padding;
+  grid-column-gap: $grid-gap;
```
edbrett: Why does this need changing? This is universal to the whole app?

edbrett:
Additionally the styles to the 'countries' item in the menu is now different to all the others

app/javascript/app/styles/layout.scss
Bluesmile82: Isn't this change for the whole app?

edbrett: But i dont understand why you have changed to grid-gap instead of just changing the value of gutter-padding

PR #293, opened 2017-11-30
Style tweaks

juancarlosalonso:
- General grid gutter
- Country page cards margins and overflow behaviour
- Apply grid in sections using percentages
- Recenter map
- Fix link styles
- Tweak timeline margins and colors


app/javascript/app/components/country-climate-vulnerability/country-climate-vulnerability-styles.scss
```diff
@@ -11,14 +11,12 @@
 }
 
 .cards {
-  display: flex;
-  flex-wrap: wrap;
-  justify-content: space-between;
-}
+  @extend %grid;
```
edbrett: 👏

app/javascript/app/components/country-ndc-overview/country-ndc-overview-styles.scss
```diff
@@ -21,13 +21,9 @@
 .actions {
   @extend %grid;
 
-  @include msGridColumns(1fr, 20fr, 32fr);
+  @include msGridColumns(45px, 1fr, 47%);
```
edbrett: Amazingly this seems to work! I think the 47% is going to pollute the grid at lower screen sizes as it calculates it from true 100% width, so the gutter gets ignored sometimes.

PR #298, opened 2017-12-07
Split text into two columns

Bluesmile82:
Little change in the styles to display the countries/NDC overview description in two columns as is quite long.

Using column css here, quite supported (IE 10+):
https://css-tricks.com/almanac/properties/c/columns

image


Bluesmile82:
I cant find the way to use the Intro description because its always in the same grid element than the text and while the buttons and title should take up only half of the width the description should take up the full width

PR #302, opened 2017-12-11
Esp info modal

Bluesmile82:
- Add modal to ESP graph with tabs for each selected models, scenarios and indicators

The selectors are very verbose. Maybe we can find some common functions but then they won't be cached. Also, that file is growing huge.

image


app/javascript/app/components/emission-pathway-graph/emission-pathway-graph-selectors.js
```diff
@@ -258,9 +272,110 @@ export const getChartConfig = createSelector(
...
+export const getScenariosSelectedMetadata = createSelector(
+  [getScenarios, getScenariosSelected],
+  (scenarios, scenariosSelected) => {
+    if (isEmpty(scenarios) || !scenariosSelected) return null;
+    const selectedScenarioIds = scenariosSelected.map(s => s.value);
+    const scenariosMetadata = scenarios.filter(
+      s => selectedScenarioIds.indexOf(s.id.toString()) > -1
+    );
+    return (
+      scenariosMetadata &&
```
edbrett: this will fail is the map above returns an empty array. Use length, as filter always returns an array.

app/javascript/app/components/modal-overview/modal-overview-component.jsx
```diff
@@ -0,0 +1,95 @@
...
+  constructor() {
+    super();
+    this.handleOnRequestClose = this.handleOnRequestClose.bind(this);
+    this.state = {
+      selectedIndex: 0
+    };
+  }
+
+  handleOnRequestClose() {
+    this.props.onRequestClose();
```
edbrett: do you need this if it is coming from props. Just put it in the onClick.

app/javascript/app/components/modal-overview/modal-overview-component.jsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  renderData() {
```
edbrett: This whole function seems rather elaborate. Can you the data not be better mapped to render simply. That is why we have selectors in the first place, to keep the components stupid and reusable. This seem like a complicated specific case.

Bluesmile82:
Ready again! I know that the modal overview can become dumber but I don't have the time to do it now. Definitely a TBD

edbrett:
Good stuff. Going to merge.

PR #304, opened 2017-12-12
Feature/graph only available models

Bluesmile82:
This PR filters the models that are available for each location in the ESP graph 🎸 

- Create an espGraph reducer and actions to fetch only the available models with time series data for each location. The responses are grouped by location to be able to cache them.
- Filter the models in the selector

image

### DIFF
--- a/app/javascript/app/components/emission-pathway-graph/emission-pathway-graph.js
+++ b/app/javascript/app/components/emission-pathway-graph/emission-pathway-graph.js
@@ -14,8 +14,10 @@ import {
   getFiltersOptions,
   getFiltersSelected
 } from './emission-pathway-graph-selectors';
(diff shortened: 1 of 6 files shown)


app/javascript/app/components/emission-pathway-graph/emission-pathway-graph.js
```diff
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import { withRouter } from 'react-router';
 import qs from 'query-string';
 import { getLocationParamUpdated } from 'utils/navigation';
+import { WORLD_LOCATION_ID } from 'data/constants';
```
edbrett: Agree with Simao. Get from selector.

PR #307, opened 2017-12-13
Defaults to filtering by AR4 when there's no gwp defined

simaob:
When requesting historical emissions data we should default to GWP of AR4 when there's no param for GWP. This will fix an issue on the front end of double counting data.

### DIFF
--- a/app/models/historical_emissions/record.rb
+++ b/app/models/historical_emissions/record.rb
@@ -20,8 +20,11 @@ module HistoricalEmissions
     end
 
     def self.filter_gwp(params)
-      selected_gwp = params[:gwp] ||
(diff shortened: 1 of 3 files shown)


app/models/historical_emissions/record.rb
```diff
@@ -19,18 +19,25 @@ def self.find_by_params(params)
       filters(records, params)
     end
 
+    def self.filter_gwp(params)
+      selected_gwp = params[:gwp] ||
+        HistoricalEmissions::Gwp.where(name: 'AR4').first
```
qoobaa: `where(name: 'AR4').first` has a shorter equivalent of `find_by(name: 'AR4')`

PR #314, opened 2017-12-18
Feature/share widgets

Bluesmile82:
- [x] Add share menu and implement it in the Button group
- [x] Add hash to link and remove previous hash
- [x] Styling

image


j8seangel:
Included a new route to embed the shared widget with the `/embed/` prefix, so for example we can do this:
image


To improve:
- Include helmet to share social metadata.
- Share with params to keep the same state.
- Embed sdg linkages map
- Embed ESP graph.

PR #317, opened 2017-12-19
Add category name to filter to avoid conflicts

Bluesmile82:
We had conflicts when filtering the ESP table. e. g: When filtering categories in indicators table scenarios where filtered too by the same categories as they had also the categories field. 

- Add a category- prefix to the URL param: 'scenarios-categories' and filter only by the selected category field


j8seangel:
Good catch !

PR #323, opened 2017-12-22
Fix/share embed

Bluesmile82:
- Extract linkages content component
- Add embed for linkages component
- Add params to shared links

---
Take a look at the table change not working when clicked

Extra: Fix button group no clickable


j8seangel:
We needed to wrap the new component `withRouter` HOC, working perfectly.
Thanks !

PR #335, opened 2018-01-05
Feature/seo

Bluesmile82:
- Added some title and description meta for the main pages with Helmet

---
This will work as the google bots will crawl the js: https://medium.freecodecamp.org/seo-vs-react-is-it-neccessary-to-render-react-pages-in-the-backend-74ce5015c0c9
The way to check if this info is ok is the fetch as Google tool. It seems to return a 307 right now. It will be nice to do some further investigation after the deploy.


j8seangel:
I had to remove the meta added in the rails template part because as those were the first one in the list the changes with react-helmet didn't work.

PR #338, opened 2018-01-11
Fix missing indicators in content overview

ticklemynausea:
This was originally reported in https://basecamp.com/1756858/projects/13795275/todos/334869422#comment_584595970

The problem has to do with the way we are reading CSV files from S3.
We are using Ruby's CSV module to read file's contents into an array of
hashes and then using the `symbol` built-in header converter to downcase
and turn the header strings into symbols. These symbols then become the
hashes' keys. However, we are matching data from the headers in
`indc/NDC_CAIT_data.csv` with data from the columns in
`indc/NDC_CAIT_legend.csv` and `indc/NDC_metadata.csv`. It turns out
that there were only two indicator slugs with mixed case: `a_DRM` and
`a_LULUCF`, so this problem wasn't obvious until now (and even the bug report
only mentions missing `a_DRM` elements)..

This PR submits changes to hardcoded references to these two
problematic indicators. I've also uploaded changed versions of the
problematic CSV files to S3. Merging this PR and reimporting INDC data will fix the issues.


j8seangel:
Thanks @ticklemynausea !!

PR #341, opened 2018-01-15
Include list of right to left countries to ndc full text

j8seangel:
Include whitelist of countries where the text direction is from right to left to fix this https://basecamp.com/1756858/projects/13795275/todos/337534990:

image

### DIFF
--- a/app/javascript/app/utils/utils.js
+++ b/app/javascript/app/utils/utils.js
@@ -35,6 +35,11 @@ export function compareIndexByKey(attribute) {
   };
 }
 
+const r2lWrittedLanguages = ['AR'];
+export function isR2LWrittedLanguage(lang) {
+  return r2lWrittedLanguages.indexOf(lang) > -1;
+}
+
 export default {
   compareIndexByKey,
   deburrUpper,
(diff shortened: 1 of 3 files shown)


app/javascript/app/utils/utils.js
```diff
@@ -35,6 +35,11 @@ export function compareIndexByKey(attribute) {
   };
 }
 
+const r2lWrittedCountries = ['IRQ'];
```
Bluesmile82: Couldn't find the Iraq full NDC text. The code seems alright. Check http://www.i18nguy.com/temp/rtl.html
To find the whole rtl countries list:

Country/Region | Script | Direction1 | Language
-- | -- | -- | --
Afghanistan | Arabic | RTL | Pashto
Israel | Hebrew | RTL | Hebrew
Middle East | Arabic | RTL | Arabic
Pakistan | Urdu | RTL | Urdu

j8seangel:
Good catch!

I've changed the helper to languages because we could have Israel with the text in English or other countries as you mentioned above. Thanks!

PR #342, opened 2018-01-15
Add subcategory dropdown and filter indicators by that subcategory

Bluesmile82:
https://www.pivotaltracker.com/story/show/154321538

- [x] Add subcategory dropdown and filter indicators
- [x] Force 2 decimals in ESP graph tooltip
- [x] Use indicator unit in the ESP graph tooltip (CO2e by default)
- [x] Show only available indicators for model

image


j8seangel:
The two decimals fix seems to work well, good catch!

PR #343, opened 2018-01-15
Include dots in lines for ghg

j8seangel:
This PR includes dots on the ghg emissions graph to avoid this:
image

This is how it looks now:
image

Bonus: new `animation` props is available on the line.

The dots appear once the line animation finish and I disabled it to avoid an intermediate state when nothing appears.


app/javascript/app/components/charts/line/line-component.jsx
```diff
@@ -13,6 +13,7 @@ import {
 import TooltipChart from 'components/charts/tooltip-chart';
 import { format } from 'd3-format';
 import debounce from 'lodash/debounce';
+import isUndefined from 'lodash/isUndefined';
```
Bluesmile82: Nice, I never liked typeof === 'undefined'

PR #352, opened 2018-01-18
feature/SSR-SEO

weberjavi:
This PR serves from Rails site meta tags needed to display on search engines and on facebook and twitter cards.  
The previous approach, letting React to add the meta tags, failed due to the fact that web crawlers do not have access to content not sent from the server.
#### Context and questions  
Assets paths (images sources to display the cards) have been set as URL's (pointing to development server), this could be problematic when the site is in production?  
Does text constants should be created in another file and required?


simaob:
Hardcoding the paths like that won't work well. If we have the assets we should use asset paths helpers that will be generated when the asset pipeline does its thing. If we don't use the screenshots in React we can even move them to the rails assets folders and this will be easier to generate. Maybe we can schedule something to go over your PR together tomorrow morning? =)

j8seangel:
We're using the screenshots on the react part so what I've thought is to duplicate them in the public folder too and access them directly using an absolute path.

Related to the paths, you'll probably have something better to handle them but in the routes.rb config we have this:
`root 'application#index'`
`get '(*frontend)', to: 'application#index'

and I don't know how to handle them.

simaob:
Oh you're right on both things. This will be fun then =D But at the moment I need to put some time in my presentation. So I can check this with Javi tomorrow, if that's OK with you?

weberjavi:
@simaob I'll be in the office since nine o'clock in the morning, if you're available we can check it then.

j8seangel:
Good work! I discussed with JC the fallback image and I changed it to use the home one.

Thanks!

PR #361, opened 2018-01-23
Feature/stories images

weberjavi:
This PR updates rss feed feature. It displays latest stories from the WRI blog on the home page.  
In the import process the stories get stored in CW data base with this columns:   
title, description, tags, background_image_url, link, published_at.  
As for now we are just using in the front end **title**, **background_image_url** and **link**  

Some changes have been requested to Christian Aldridge to avoid some xml data sanitazion and manipulation on the import process:   

**Changes requested**
*There are some xml fields that are sending html tags (title and link), I'm sanitizing the title and extracting the href from the link and attaching it to the domain (to generate a URL) but it would be nice to receive the actual content and the URL to avoid possible bugs in edge cases.*

*We are receiving the blog post original image URL, sometimes this image is quite big. Would be nice if you can send instead the resized image that you're using in the blogpost
blogpost: http://www.wri.org/blog/2017/11/how-cities-can-harness-good-and-avoid-bad-new-mobility-movement
original image: http://www.wri.org/sites/default/files/6870785251_9ddaed1a67_o.jpg
resized image: http://www.wri.org/sites/default/files/styles/large/public/6870785251_9ddaed1a67_o.jpg?itok=ilpIkdK2*

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,7 +40,6 @@ gem 'aws-sdk', '~> 2'
 gem 'active_model_serializers', '~> 0.10.0'
 gem 'oj'
 gem 'scenic'
-gem 'simple-rss'
 
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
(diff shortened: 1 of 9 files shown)


weberjavi:
Nice, getting the link is a better approach than hardcode it, it would avoid feature breaking if there are changes on the data source.  
Thank you!!

PR #362, opened 2018-01-23
Feature/responsive setup

weberjavi:
This PR sets the initial breakpoints and relative font sizes to use in the CW responsive version. It is defined to be *mobile first*, that is default styles and layouts will be the mobile version ones and tablet and desktop versions will be wrapped in media queries.  
For scenarios where a change in the HTML structure is needed a ` Responsive` component has been created. It uses [`'react-responsive'`](https://github.com/contra/react-responsive) library.  
Importing it in the component where it wants to be used allows to do stuff like:  
```
<Desktop>
  <p>Render desktop</p>
</Desktop>
<Tablet>
  <p>Render tablet</p>
</Tablet>
```

### DIFF
--- a/app/javascript/app/styles/settings.scss
+++ b/app/javascript/app/styles/settings.scss
@@ -2,13 +2,13 @@
 $font-family-1: 'Lato', sans-serif;
 
 // Font sizes
-$font-size-s: 13px;
-$font-size-sm: 14px;
-$font-size: 16px;
-$font-size-m: 19px;
-$font-size-large: 24px;
-$font-size-x-large: 44px;
-$font-size-xl-large: 54px;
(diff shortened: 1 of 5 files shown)


app/javascript/app/styles/settings.scss
```diff
@@ -2,13 +2,13 @@
...
 
 // Font sizes
-$font-size-s: 13px;
-$font-size-sm: 14px;
-$font-size: 16px;
-$font-size-m: 19px;
-$font-size-large: 24px;
-$font-size-x-large: 44px;
-$font-size-xl-large: 54px;
+$font-size-s: 0.8125rem; // 13px / 16px
```
j8seangel: Good idea!

PR #364, opened 2018-01-24
Feature/mobile menu country select

weberjavi:
#### What this PR does
This PR updates responsive breakpoints, both in SASS settings and `<Responsive>` component, to adapt to a more accurate range of design changes on the UI.  
It also adapts country select component styles and behaviour to design responsive version.  
Design could be found [here](https://app.zeplin.io/project/598b1dc89689e51e1cd1b5ce/screen/5a5dc4e724f129ede08d34c3?)  
#### Comments   
`<Map />` component responsive behaviour is handled through `<Responsive>` component, instead of CSS media queries, to avoid `<Map />` mounting in small devices.

### DIFF
--- a/app/javascript/app/components/search/search-component.jsx
+++ b/app/javascript/app/components/search/search-component.jsx
@@ -16,6 +16,12 @@ class Search extends Component {
     };
   }
 
+  componentDidMount() {
+    if (this.props.autofocus) {
+      this.inputRef.focus();
+    }
+  }
+
   componentWillReceiveProps(nextProps) {
     if (nextProps.input !== this.props.input) {
       this.setState({ search: nextProps.input });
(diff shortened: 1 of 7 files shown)


app/javascript/app/components/search/search-component.jsx
```diff
@@ -16,6 +16,12 @@ class Search extends Component {
     };
   }
 
+  componentDidMount() {
+    if (this.props.autofocus) {
+      this.inputRef.focus();
```
Bluesmile82: This is not working on my computer. Maybe it will be better to focus when you click on the nav item

weberjavi: Auto focus is only implemented in country selected page but not in country select on the nav-nested-menu.  That's because from the mobile menu the navigation will route the user to the page and it will not open the nav-nested-menu. Does it makes sense?

j8seangel: It makes!

PR #365, opened 2018-01-24
Fix/show unit y axis

Bluesmile82:
This PR fixes:
https://basecamp.com/1756858/projects/13795275/todos/335749809#attachments_for_todo_335749809
- Adds y unit to y-axis in ESP graph
image

- Changes format to ''. Decimal or exponential format so the numbers won't show any SI prefix. Also fixes next to zero values in the first tick and changes it to 0. See options here (https://github.com/d3/d3-format)
- Passes an espGraph variable as too much specific code from that graph was affecting the others.

Extra: 
https://basecamp.com/1756858/projects/13795275/todos/335337561
- Fix default indicator selection

### DIFF
--- a/app/javascript/app/components/charts/line/line-component.jsx
+++ b/app/javascript/app/components/charts/line/line-component.jsx
@@ -8,7 +8,8 @@ import {
   YAxis,
   CartesianGrid,
   Tooltip,
-  ResponsiveContainer
+  ResponsiveContainer,
+  Label
 } from 'recharts';
 import TooltipChart from 'components/charts/tooltip-chart';
 import { format } from 'd3-format';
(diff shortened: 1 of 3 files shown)


app/javascript/app/components/charts/line/line-component.jsx
```diff
@@ -146,12 +174,15 @@ ChartLine.propTypes = {
   height: PropTypes.any.isRequired,
   onMouseMove: PropTypes.func.isRequired,
   domain: PropTypes.object,
-  forceTwoDecimals: PropTypes.bool
+  forceTwoDecimals: PropTypes.bool,
+  margin: PropTypes.object,
+  espGraph: PropTypes.bool.isRequired
```
j8seangel: I'd prefer something more reusable that this prop, but is ok for now

PR #369, opened 2018-01-29
Fix/graph remove default cero

Bluesmile82:
https://basecamp.com/1756858/projects/13795275/todos/339797072
- The graph was showing a non-linear scale starting always at 0. Now we should the minimum value as a start

image

https://basecamp.com/1756858/projects/13795275/todos/335337561
- Also, the default subcategory has changed in the data so we needed to update it to GHG Emissions by gas with LULUCF. The selection by default has been extracted to a constant.

### DIFF
--- a/app/javascript/app/components/emission-pathways/emission-pathways-graph/emission-pathways-graph-selectors.js
+++ b/app/javascript/app/components/emission-pathways/emission-pathways-graph/emission-pathways-graph-selectors.js
@@ -43,6 +43,15 @@ const AXES_CONFIG = {
   }
 };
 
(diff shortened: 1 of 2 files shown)


app/javascript/app/components/emission-pathways/emission-pathways-graph/emission-pathways-graph-selectors.js
```diff
@@ -43,6 +43,15 @@ const AXES_CONFIG = {
   }
 };
 
+const DEFAULT_SELECTIONS = {
```
j8seangel: 👏

PR #371, opened 2018-01-29
Hide t in the yaxis and label unless absolute value is selected

Bluesmile82:
- We don't want to show t in the y-axis and the tooltip of the country GHG and GHG graphs
Some code from the ESP branch was copied to avoid conflicts
image


j8seangel:
I've fixed the x-axis as just noticed the scale wasn't linear so we were missing the empty years.

PR #373, opened 2018-01-30
Feature/responsive modal

Bluesmile82:
This PR makes the modal component responsive:

- Refactor of the modal to include the header as a prop in the Modal.
- Use flexbox to make the modal header fix while the content is scrolling
- Include themes
- Adjust z-index as we want the menu to be on the top even if the modal is open

---
There have been a lot of changes and turnarounds so it's better to look directly at the code

image


Bluesmile82:
Ready again! Please take a look or otherwise I will merge

j8seangel:
We got it!! Thanks

PR #378, opened 2018-01-30
feature/allow-cors

pjosh:
Allow CORS on the API


j8seangel:
Looks good to me, but does it worth to include only a whitelist?
And what endpoints do you need?

Bluesmile82:
Looks ok, as is only a GET and we don't expose sensitive information. We will have to change it otherwise

PR #379, opened 2018-01-30
Feature/scenario overview

Bluesmile82:
- Add scenario overview, remove empty fields from the modal
- Add maintainer data from the scenario to the overview section

Extra: 
- Add a options sanitizer to the dropdown to prevent a strange bug that was breaking the graph when going from a scenario page to the ESP page. It will display a warning and an empty option instead of breaking.

image
image


j8seangel:
And good catch with the sanitizer!

PR #380, opened 2018-01-31
Add footer to embed, style and embed-test file

Bluesmile82:
The embed layout needed a footer:

- Add footer with link to embed and style
- Create HTML to test the embed easily. You just have to run a server in that folder, try:
```python -m SimpleHTTPServer 8000```

This will improve with the responsive version. Although we will have to create a specific design if little heights should be used.
image

### CODE REVIEW

File: app/javascript/app/layouts/embed/embed-component.jsx [hunk 1]
```diff
@@ -2,16 +2,25 @@ import React, { PureComponent } from 'react';
...
     const { route } = this.props;
     return (
-      <div className={layout.content}>
+      <div className={[layout.content, styles.embed]}>
         <CountriesProvider />
-        {renderRoutes(route.routes)}
+        <div className={styles.embedContent}>{renderRoutes(route.routes)}</div>
+        <div className={styles.footer}>
+          For more information, visit
+          <a href="https://www.climatewatchdata.org/" parent="_blank">
```
weberjavi: When clicking on the link it is not redirecting to ClimateWatch website.  
Instead of ``parent="_blank"`` shouldn't it be ``target="_blank"``?

Bluesmile82:
Ready again! Thanks for the comment!

j8seangel:
Good embed test example!

PR #381, opened 2018-01-31
Make partners page responsive

Bluesmile82:
About page is mostly responsive, we only needed to change the partners' page disposition:

image

### DIFF
--- a/app/javascript/app/app.jsx
+++ b/app/javascript/app/app.jsx
@@ -5,6 +5,8 @@ import { BrowserRouter } from 'react-router-dom';
 import { renderRoutes } from 'react-router-config';
 import ScrollToTop from 'components/scroll-to-top';
 
+import 'styles/sticky.scss';
+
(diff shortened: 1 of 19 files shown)


weberjavi:
In the ``permissions & licensing``  section the link in the bottom of the page overflows the viewport width allowing the user to scroll horizontally    

screen shot 2018-02-02 at 10 37 20

simaob:
@weberjavi I'm not sure we have the texts where they are missing. I mean we did have something for Vizz but not for the others I don't think.

simaob:
It is sad :(

Bluesmile82:
Quick fix! I'm thinking that sometimes the anchor-nav doesn't make very clear that we have more than two elements to chose from. Maybe we could increase the gradient effect so its obvious

j8seangel:
We've included the sticky anchor nav on the about page and added some other fixes on a general review (not related with this PR itself)

PR #387, opened 2018-02-06
Feature/linkages responsive

weberjavi:
This PR develops ndc-sdg linkages responsive version.  
It addresses layout design changes  and updates some page behaviours to  adapt to inexistent `onHover` event on mobile (goals selection and map tooltip trigger and country navigation).  
Designs can be checked on [zeppelin](https://app.zeplin.io/project/598b1dc89689e51e1cd1b5ce/screen/5a60835e461834cc5fc6cb27).    
 
linkages


app/javascript/app/components/ndc-sdg/ndc-sdg-linkages-map/ndc-sdg-linkages-map-styles.scss
```diff
@@ -1,8 +1,33 @@
 @import '~styles/layout.scss';
+@import '~styles/settings.scss';
 
 .container {
-  height: 100%;
-  position: relative;
+  position: absolute;
+  width: 100%;
+  height: 50vh;
```
Bluesmile82: It would be nice to use only around 30-35vh for the map like in the designs, so we don't have that empty space on the top

app/javascript/app/components/ndc-sdg/ndc-sdg-linkages-content/ndc-sdg-linkages-content-component.jsx
```diff
@@ -17,20 +18,31 @@ class NdcSdgLinkagesContent extends PureComponent {
...
-          targetHover={targetHover}
-          onTargetHover={handleTargetHover}
-        />
-        <NdcSdgLinkagesMap goalHover={goalHover} targetHover={targetHover} />
-        <ModalMetadata />
-      </div>
+      <TabletLandscape>
+        {matches => (
+          <div
+            className={cx({ [layout.content]: matches === true }, styles.grid)}
```
j8seangel: Maybe a better variable name could help to understand what it does, for example, an `isTablet` ? also we don't need to compare with `===true` and it's already a boolean

app/javascript/app/components/ndc-sdg/ndc-sdg-linkages-map/ndc-sdg-linkages-map-styles.scss
j8seangel: Agree, the design seems to be using the 40% of the height.

app/javascript/app/pages/ndc-sdg/ndc-sdg-styles.scss
```diff
@@ -1,13 +1,19 @@
...
-  grid-template-columns: 1fr 2fr;
-}
+  @media #{$mobile-only} {
+    height: 50vh;
+  }
 
-.bg {
-  background-color: $light-gray;
+  @media #{$tablet-portrait} {
+    height: 45vh;
```
j8seangel: To match the map height size, a good idea could be to save it in a variable and share it across components, couldn't it?

app/javascript/app/components/ndc-sdg/ndc-sdg-linkages-content/ndc-sdg-linkages-content-component.jsx
```diff
@@ -17,20 +18,31 @@ class NdcSdgLinkagesContent extends PureComponent {
...
+          >
+            <NdcsSdgsMetaProvider />
+            <NdcSdgLinkagesTable
+              goalHover={goalHover}
+              onGoalHover={handleGoalHover}
+              onGoalTap={handleGoalTap}
+              targetHover={targetHover}
```
j8seangel: And I'm not sure we need this anymore, could you review it?

j8seangel:
I've spotted a new bug not related to this feature but could you check when you click on the empty card? We should handle the error. Thanks!

app/javascript/app/components/ndc-sdg/ndc-sdg-linkages-content/ndc-sdg-linkages-content-component.jsx
weberjavi: Maybe the name of the Tap event is a little bit misleading, because what we are really handling is the click event, but we are triggering this handler on goal click to set the proper data in the map on mobile devices (thats why I call it onTap, to understand that this handler aims touch devices).

app/javascript/app/pages/ndc-sdg/ndc-sdg-component.jsx
```diff
@@ -29,7 +31,13 @@ class NdcSdg extends PureComponent {
           href={location.href}
         />
         <NdcsSdgsMetaProvider />
-        <Header size="small" route={route} className={styles.header}>
+        <Header
+          size="small"
+          route={route}
+          className={cx(styles.header, {
+            [styles.isOpen]: qs.parse(location.search).goal !== ''
```
j8seangel: I don't think that is responsibility of the component, the container is the one who takes care of theses thing passing props to the component keeping it "dumb"
So I'd move this logic to the container and pass it as `isOpen` or any other name prop to the component

app/javascript/app/pages/ndc-sdg/ndc-sdg-component.jsx
```diff
@@ -40,7 +48,11 @@ class NdcSdg extends PureComponent {
             </div>
           </div>
         </Header>
-        <div className={styles.wrapper}>
+        <div
+          className={cx(styles.wrapper, {
+            [styles.isOpen]: qs.parse(location.search).goal !== ''
```
j8seangel: Same here, we are duplicating logic.

j8seangel:
Please review the styles too because now the cards have double lined border between each other
image

j8seangel:
👏

PR #391, opened 2018-02-07
Feature/responsive esp

Bluesmile82:
- [x] Make ESP page Responsive
- [x] Make Model page Responsive
- [x] Make Scenario page Responsive
- [x] Create filters dropdown component
- [x] Make table scrollable

Extra:
Fix prop-type warning


app/javascript/app/components/table/table-component.jsx
```diff
@@ -31,17 +32,20 @@ class SimpleTable extends PureComponent {
 
     if (!data.length) return null;
     const hasColumnSelectedOptions = hasColumnSelect && columnsOptions;
+    const getResponsiveWidth = width => {
```
j8seangel: Could you spend some time to play with the number of columns selected to obtain the ratio?
Then we can obtain more readable tables in mobile, what are your thoughts?

app/javascript/app/components/collapse/collapse-component.jsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  render() {
+    const { handleToggleCollapse, opened } = this.props;
+
+    return (
+      <div>
+        <Button
+          className={cx({ [styles.open]: opened })}
+          onClick={() => handleToggleCollapse(c => !c)}
```
j8seangel: I'd like more to have this function [here](https://github.com/Vizzuality/climate-watch/pull/391/files#diff-181e77e217727451a3abb76c41300e42R4) instead of inside the component.
So to make easier to reuse the collapse component I'd change this function to an onToggleClick callback and I'd do the login in the container.

Bluesmile82:
Fixed collapse destination and changed responsive table. Let me know if works for you

PR #392, opened 2018-02-08
Feature/filter stories by tag

weberjavi:
This PR allows filtering Stories received from WRI RSS feed.   
We are using tags received from the feed to filter which stories are going to be displayed on the home page.  
Selected tags are defined on the request to the API at ``app/javascript/app/components/stories/stories-actions.js``   
On the import to the data base we are importing all stories to avoid having less than five stories on the home page in case that there are not enough stories with the desired tags (actually that is what happens).   
Bandcamp thread can be checked [here](https://basecamp.com/1756858/projects/13795275/todos/333348992)

### DIFF
--- a/app/controllers/api/v1/stories_controller.rb
+++ b/app/controllers/api/v1/stories_controller.rb
@@ -2,12 +2,28 @@ module Api
   module V1
     class StoriesController < ApiController
       def index
-        stories = Story.limit(stories_limit).order(:published_at)
-
+        stories = storiesFilter(params[:tags])
         render json: stories, each_serializer: Api::V1::StorySerializer
       end
 
       private
+
+      def storiesFilter tags
+        taggedStories = taggedStories(tags)
+        return taggedStories if taggedStories.length >= 5
+        moreStories = untaggedStories
+        return (taggedStories + moreStories).first(5) if taggedStories
+        return moreStories
+      end
+
+      def taggedStories tags
+        tagsArray = tags.split(',')
+        Story.where("tags && ARRAY[?]::varchar[]", tagsArray).order(published_at: :desc).limit(stories_limit)
+      end
+
+      def untaggedStories
+        Story.order(published_at: :desc).limit(stories_limit)
+      end
 
       def stories_limit
         params[:limit] || 5
--- a/app/controllers/api/v1/stories_controller.rb
+++ b/app/controllers/api/v1/stories_controller.rb
@@ -2,26 +2,28 @@ module Api
   module V1
     class StoriesController < ApiController
       def index
-        stories = storiesFilter(params[:tags])
+        stories = stories_filter(params[:tags])
         render json: stories, each_serializer: Api::V1::StorySerializer
       end
 
       private
 
-      def storiesFilter tags
-        taggedStories = taggedStories(tags)
-        return taggedStories if taggedStories.length >= 5
-        moreStories = untaggedStories
-        return (taggedStories + moreStories).first(5) if taggedStories
-        return moreStories
+      def stories_filter(tags)
+        tagged_stories = tagged_stories(tags)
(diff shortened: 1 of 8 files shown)


app/controllers/api/v1/stories_controller.rb
```diff
@@ -2,13 +2,31 @@ module Api
...
 
+      def stories_filter(tags)
+        tagged_stories = tagged_stories(tags)
+        return tagged_stories if tagged_stories.length >= 5
+        more_stories = untagged_stories
+        return (tagged_stories + more_stories).first(5) if tagged_stories
+        more_stories
+      end
+
+      def tagged_stories(tags)
```
simaob: I think that these methods should go into the model, instead of the controller.
And can you add some unit specs for these? (tagged_stories and untagged_stories)

app/controllers/api/v1/stories_controller.rb
```diff
@@ -2,13 +2,31 @@ module Api
...
+        tags_array = tags.split(',')
+        # rubocop:disable LineLength
+        Story.where('tags && ARRAY[?]::varchar[]', tags_array).order(published_at: :desc).limit(stories_limit)
```
simaob: you can break the chain before the order, moving order to the next line and leaving the dot at the end of the line. This way you don't need to disable rubocop.

spec/models/story_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe Story, type: :model do
+  describe '#tagged_stories' do
+    it 'should return one story' do
+      FactoryGirl.create(:story, tags: ['NDC'])
```
simaob: to make this a more reliable test I'd create a second story without the tag, this way we can ensure that it's only retrieving the tagged one.

spec/models/story_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it 'should not return any story' do
+      FactoryGirl.create(:story, tags: ['OP23'])
+      expect(Story.tagged_stories('NDC, esp', 5)).to have(0).items
+    end
+  end
+
+  describe '#untagged_stories' do
+    it 'should return no story' do
+      FactoryGirl.create(:story)
```
simaob: This test is not explicit enough in my opinion. Because it's trusting the factory to create the story with tags, it's not transparent that that's the case. Given that we are testing if it will return only untagged stories, I would add the tags information to the FactoryGirl create method.

simaob: And now that I read the test bellow, I'm a bit confused, sorry.

So my real comment is. This test with "0" as an argument to `untagged_stories` is in fact testing the ActiveRecord `limit` method. I'd suggest that the test would create some tagged stories and then try to fetch untagged stories and have none be returned.

spec/models/story_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      expect(Story.untagged_stories(0)).to have(0).items
+    end
+
+    it 'should return one story' do
+      FactoryGirl.create(:story)
```
simaob: In here, I'd create a story with explicit `nil` value on tags and then call the `untagged_stories` method to prove that it would return the right story. For more reliability, given the limit `1`, I'd probably also create a story with tags to show that it just returns the relevant story.

spec/models/story_spec.rb
weberjavi: Maybe the name of the method isn't adequate enough. It should return stories, tagged or not, that not match the tags that have been passed as query params

simaob: Yes, that is a bit misleading. So maybe it could be called `not_tagged_by(my_tags)`

weberjavi:
Thanks @simaob for the pair programming session!! 
I mean, with you pair of hands :P

PR #395, opened 2018-02-09
ESP - Remove indicators,  add get in touch button, custom message for graph

Bluesmile82:
image
- Remove indicators from menu
- Add get in touch button
- Remove all the indicators logic and change url to models/7 instead of models/7/scenarios
- Add custom message when the selected model or indicator is not available:
image


j8seangel:
Could you please review the feedback message when the indicator selected is the default one?
To replicate the bug just go to the [default page](http://localhost:3000/emission-pathways/models) and change the country/region to any other one.

image

Thanks!

Bluesmile82:
That seems to be a racing condition not related to the PR as it was happening before. Only happens when you don't wait until the world data to load and you switch to some countries: Turkey, US. I created a new commit that keeps the loading until we have a selected model so it won't show the message. It loads the correct model after a while

PR #403, opened 2018-02-13
Show a margin for negative values and widen y=0 line

Bluesmile82:
This PR solves the lack of visibility of negative values due to their magnitude in the country GHG graph.

Try with: http://localhost:3000/countries/IND?filter=192%2C189&source=16

Extra:

- Fix sectors that were being filtered in the per capita and per GDP options. Now you only can see the real total.

image


simaob:
What's left to do here?

Bluesmile82:
Ready!

j8seangel:
Looks good to me as has the requested functionality, we could iterate later to include the reference area to show the negative part with a different colour to make it clearer. Good work!

PR #406, opened 2018-02-15
Feature/pinned story

weberjavi:
This PR adds `pinned` filter to the Stories model. That way we will pin the last Story with the `climatewatch-pinned` tag as the main story in the home page.  
Basecamp thread could be checked [here](https://basecamp.com/1756858/projects/13795275/todos/333348992#comment_598780477)

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -5,11 +5,11 @@ class Story < ApplicationRecord
 
     pinned_stories = pinned_stories(1)
     tagged_stories = tagged_stories(tags_array, limit)
-    main_stories = pinned_stories + tagged_stories
+    main_stories = (pinned_stories + tagged_stories).uniq
     return main_stories if main_stories.length >= limit
 
     more_stories = not_tagged_by(tags_array, limit)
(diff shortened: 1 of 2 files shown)


simaob:
Hi Javi, I was having a think and maybe instead of just picking one pinned stories we could fill the "list" with pinned stories sorted by date, and then other tagged stories, and then stories that aren't tagged.

I'm thinking that there might be more than one pinned stories at any given time, and given the definition of pinned from other websites (like twitter or forums) I think it might be a good idea to allow more than one pinned with the "main story" being the most recent pin.

Let me know what you think.

weberjavi:
Sounds fair, but maybe that way we end up with some outdated stories on the front page or forcing to add the pinned tag to all the stories that 'needs' to be displayed on the home page (it could turn in a horror story like when you begin to drop `important!` in a CSS file :)  
I suppose it depends in the interface WRI has to tag the stories (if is an easy task to unpin stories).

simaob:
That's true... maybe let's go with this approach first and then iterate as needed.

PR #414, opened 2018-02-19
Feature/compare graph

Bluesmile82:
Compare countries graph:

This PR creates the compare countries graph section merging the functionality of the ghg graph and the country graph.
- It generates a graph using a new emissions provider that is shared with Ghg-emissions (ghg emissions reducer and actions have been deleted) Actually this is a case where you can see that providers are much cleaner in the implementation than reducers for these uses if you are using reselect.
(If we create the quantifications part and we update this provider we will be also using all the logic of the country GHG emission reducer so we will be able to delete this one as well)

Other fixes included: 
- Adjust min heights for the chart (also in GHG chart)
- Fix selectors
- Fix unit warning
- Fix layout on embed
---
 Extra: 

- Fix dropdowns that were modified by the country selector and fix styles for this component according to the design

image


Bluesmile82:
Thanks for noticing! I will remember to restart the server when I test the feature

Bluesmile82:
Ready with the required changes

j8seangel:
Thanks for the changes, look awesome!

PR #416, opened 2018-02-20
Feature/esp scenario page changes

Bluesmile82:
- Add Unit for indicators
- Order alphabetically by category (It was needed to create a selector as the category is an object and the default sorting of the table wasn't working with it)
- Link to the pathways graph from the name with filters (scenario, model, country/region, category, subcategory and indicator)
- Change the ordering of columns to "Category", "Sub-Category", "Name",  'Unit', 'trend'
- Add first and last columns with respective year and data for the trendlines.

image
---
Extra:
- Order the ESP table alphabetically by category in indicators section
- Fix anchor navbar


j8seangel:
Good work!

I've updated the link to use the `<Link>` component from react-router because using `<a>` tag we were reloading the entire page what sucks in performance terms. 

Thanks!

PR #429, opened 2018-02-27
Fix/timeline tooltip

Bluesmile82:
- Fix timeline tooltip in landscape
image
- Remove repeated border in add button
image


j8seangel:
I've included a small change to keep the border at the top when you scroll down.
Thanks for the quick fix.

PR #434, opened 2018-02-28
Feature/founders logos ordering

weberjavi:
This PR addresses the feature described in [this basecamp thread](https://basecamp.com/1756858/projects/13795275/todos/342076588).  
It adds two more founders to the about page and updates footer behaviour as defined in [this pivotal task](https://www.pivotaltracker.com/n/projects/2083759/stories/155546415).  

logos


j8seangel:
I don't think is the component responsibility to check the actual path and decide by itself when rendering something or not... this is the main issue for reusability and that is why we are using the container to parametrize those kinds of things.

There are also some style changes to match design requests, please review it later.

weberjavi:
I see the point. Sorry for that. Maybe I´m more like a 'dumb' component :(

PR #437, opened 2018-03-01
Feature/grid linkages page

weberjavi:
This PR updates xy-grid mixins to accept several arguments.  
It also adapts sdg-linkages styles to make use of the new grid mixins.  
In order to check this PR please join both home and sdg-linkages page and check the layoutn in different browser widths.  

sdg-grid

### DIFF
--- a/app/javascript/app/components/stories/stories-styles.scss
+++ b/app/javascript/app/components/stories/stories-styles.scss
@@ -77,17 +77,41 @@
 }
 
 .grid {
-  @include xy-grid($direction: 'vertical', $wrap: true);
+  @include xy-grid($wrap: true);
 
   height: 100%;
 
+  @media #{$tablet-landscape} {
+    @include xy-grid($direction: 'vertical', $wrap: true);
+  }
+
   > a {
-    @include xy-cell(3, $gutters: 0);
-    @include xy-cell(50%, $gutters: 0, $vertical:true);
+    @include xy-cell(6, $gutters: 0);
+    @include xy-cell(25%, $gutters: 0, $vertical:true);
+
+    @media #{$tablet-portrait} {
+      @include xy-cell(6, $gutters: 0);
+      @include xy-cell(33.333333%, $gutters: 0, $vertical:true);
+    }
+
+    @media #{$tablet-landscape} {
+      @include xy-cell(3, $gutters: 0);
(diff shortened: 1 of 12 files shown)


app/javascript/app/components/stories/stories-styles.scss
```diff
@@ -77,17 +77,41 @@
...
 
   & :first-child {
-    @include xy-cell(6, $gutters: 0);
-    @include xy-cell(100%, $gutters: 0, $vertical:true);
+    @include xy-cell(12, $gutters: 0);
+    @include xy-cell(50%, $gutters: 0, $vertical:true);
+
+    @media #{$tablet-portrait} {
+      @include xy-cell(12, $gutters: 0);
+      @include xy-cell(33.333333%, $gutters: 0, $vertical:true);
```
Bluesmile82: Arent we trying to avoid this? I think 4 of 12 would be better. It would be great to keep all the measures related to the 12 columns

weberjavi: The problem I found is that the grid created measures have only a couple of decimal places and hence the cells do not span in the 100% of the container resulting in white gaps between them :(

app/javascript/app/styles/layout.scss
```diff
@@ -27,12 +27,26 @@ $grid-padding-gutters: $gutter-padding;
...
-  @include xy-gutters($negative: $inner);
+  @include xy-gutters($gutters: $gutter-size, $negative: $inner);
 
   > * {
-    @include xy-cell(12/$number);
+    @include xy-cell($size: 12/$number, $gutters: $gutter-size);
+  }
+}
+
+@mixin customColumns($columns, $grid-gutters: $gutter-padding, $cell-gutters: $gutter-padding) {
```
Bluesmile82: It would be nice to have the same name for the gutter variables in the two mixins. So have grid-gutters and cell-gutters for both for example. Also, merge the two mixins if possible as they are very similar. But this could be a bit harder.

weberjavi: Agree with the variables naming.  
About merging both mixins maybe could become less explicit, in the classes using them, the kind of grid that we are trying to create. I would go with both of them.

j8seangel:
Closing as the mixins are outdated, please review #435. Won't delete the branch yet in case we need to get some info from it.

PR #441, opened 2018-03-05
Fix/css fixes ndc content

Bluesmile82:
- Map not draggable
- Fix margin between selectors and map and mobile center
- Fix top margin for the table
- Min height for no results
image
image

I also added the CSS-fixes for the home page in this PR:

- Fix margin between sections
image

- Country profile image should end with the section bottom border
image

- Update screenshots to match new styles
- Remove youtube fullscreen because we are using our own and don't work well together
image


Bluesmile82:
Good point! I added the margin also for mobile. Please also review the home page fixes included in this PR

PR #442, opened 2018-03-05
Feature/partners footer

weberjavi:
This PR fixes some css issues and requests on partners footer.  
Fixes are described in [this pivotal ticket](https://www.pivotaltracker.com/story/show/155546415).  
It also addresses the footer requirement suggested in [this basecamp thread](https://basecamp.com/1756858/projects/13795275/todos/344119040). 
  - Adds alphabetical ordering on partners display. 
  - Fixes some footer styling to adapt to mobile screens.  
  - Fixes navbar breaking on small screens due to pathways being added to navbar    

screen shot 2018-03-05 at 13 39 00    

screen shot 2018-03-05 at 14 38 27

### DIFF
--- a/app/javascript/app/components/footer/footer.js
+++ b/app/javascript/app/components/footer/footer.js
@@ -128,16 +128,21 @@ const alphabetically = (a, b) => {
   return 0;
 };
 
+const parsedPath = pathname => {
+  if (pathname === '/ghg-emissions' || pathname === '/ndcs-sdg') { return 'ghg & ndcs-sdg'; }
+  if (pathname.includes('/countries') || pathname.includes('/ndcs')) { return 'countries & ndcs'; }
+  if (pathname.includes('/pathways')) return 'pathways';
+  return 'default';
+};
+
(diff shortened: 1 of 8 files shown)


app/javascript/app/components/footer/footer.js
```diff
@@ -119,11 +136,13 @@ const getLogos = pathname => {
       return [...basePartners];
     case '/ndcs':
     case '/countries':
-      return [...basePartners, ...ndcPartners];
-    case '/emission-pathways':
-      return [...basePartners, ...espPartners];
+      return [...basePartners, ...ndcPartners].sort(alphabetically);
+    case '/pathways/models':
```
Bluesmile82: I think we should use just pathways. Models in this case is for the menu of the table. So it can be /scenarios or /indicators too and still will be the same page

j8seangel:
Good job! 
We'll have to iterate over this changes on the footer styles with the new grid mixins but for now is perfect!

PR #447, opened 2018-03-07
Feature/show country ghg disclaimer

Bluesmile82:
https://www.pivotaltracker.com/story/show/155740305
https://zpl.io/anrRogx

This PR adds a disclaimer to be shown in the country Greenhouse Gas Emissions and Emissions Targets Section only once per browser. So once the disclaimer is shown a variable is added in localStorage to prevent it from showing again.

The disclaimer is also shown at the end of the info modal of this section. As the country page share the same modal Component and it is not possible to select the right modal component rendered in one of the sections, the modal has an extraContent prop with the title of the modal as the key for the extra content. 

To test several times add
```localStorage.setItem('disclaimerShown', 'false');```
to line 7 in disclaimer.js just before the state declaration

---
Extra:

Add ClickOutside to Modal component so it will close if we click outside the modal.

image

image

### DIFF
--- a/app/javascript/app/components/disclaimer/disclaimer-styles.scss
+++ b/app/javascript/app/components/disclaimer/disclaimer-styles.scss
@@ -3,7 +3,7 @@
 .disclaimerBox {
   box-shadow: 0 5px 15px 0 rgba(0, 0, 0, 0.1);
   display: flex;
-  padding: 30px;
+  padding: $gutter-padding;
   position: relative;
   line-height: $line-height-medium;
   font-size: $font-size-xs;
--- a/app/javascript/app/components/disclaimer/disclaimer-styles.scss
+++ b/app/javascript/app/components/disclaimer/disclaimer-styles.scss
@@ -1,0 +1,45 @@
+@import '~styles/layout.scss';
+
+.disclaimerBox {
+  box-shadow: 0 5px 15px 0 rgba(0, 0, 0, 0.1);
+  display: flex;
+  padding: 30px;
+  position: relative;
+  line-height: $line-height-medium;
+  font-size: $font-size-xs;
+  background-color: $white;
+  flex-direction: column;
+
+  @media #{$tablet-landscape} {
+    padding: 20px;
+    flex-direction: row;
+  }
+}
+
+.firstText {
+  color: $theme-color;
+  flex: 1;
+}
+
+.secondText {
+  color: $dark-gray;
+  flex: 1;
+  margin: 10px 0 0;
+
+  @media #{$tablet-landscape} {
+    margin: 0 20px;
+  }
+}
+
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: app/javascript/app/components/disclaimer/disclaimer.js [hunk 1]
```diff
@@ -0,0 +1,26 @@
+import { createElement, PureComponent } from 'react';
+import Component from './disclaimer-component';
+
+class DisclaimerContainer extends PureComponent {
+  constructor() {
+    super();
+    this.state = {
+      hasBeenShown: JSON.parse(localStorage.getItem('disclaimerShown'))
```
j8seangel: Could we move this to a constant?

app/javascript/app/components/modal/modal-component.jsx
```diff
@@ -34,19 +35,21 @@ class CustomModal extends PureComponent {
...
-        contentLabel={contentLabel}
-        shouldCloseOnOverlayClick={shouldCloseOnOverlayClick}
-      >
-        {header}
-        <Button onClick={onRequestClose} className={theme.closeBtn} square>
-          <Icon icon={closeIcon} className={theme.close} />
-        </Button>
-        <div className={theme.content}>{children}</div>
-      </Modal>
+      <ClickOutside onClickOutside={onRequestClose}>
```
j8seangel: We don't need another dependency to attach a listener to click outside because the library already has one, please review this example http://reactcommunity.org/react-modal/examples/should_close_on_overlay_click.html#using-shouldcloseonoverlayclick and update our own version with that feature

app/javascript/app/components/disclaimer/disclaimer-styles.scss
```diff
@@ -0,0 +1,52 @@
+@import '~styles/layout.scss';
+
+.disclaimerBox {
+  box-shadow: 0 5px 15px 0 rgba(0, 0, 0, 0.1);
+  display: flex;
+  padding: 30px;
```
j8seangel: $gutter-padding 👀?

app/javascript/app/components/disclaimer/disclaimer.js
Bluesmile82: A constant won't update the component with the createElement like the state does. If you think that it could let me know and we can fix it tomorrow

app/javascript/app/components/modal/modal-component.jsx
Bluesmile82: 👍 Thanks!

simaob:
@Bluesmile82 is this ready to be reviewed again by José?

j8seangel:
Not yet, just something else to fix.

Bluesmile82:
Ready!

PR #449, opened 2018-03-08
Feature/grid update countries page

Bluesmile82:
This PR updates the grid for countries page and simplifies and fixes the code. The appearance is almost the same as the one we had with some fixed margins and paddings.
Some structure is slightly different to adjust to the real grid, mostly the NDC Overview and LInkages butttons:

image

---
Extra:
- Fix navbar in mobile z-index (it was hiding behind the modal)

### DIFF
--- a/app/javascript/app/components/country/country-ndc-overview/country-ndc-overview-component.jsx
+++ b/app/javascript/app/components/country/country-ndc-overview/country-ndc-overview-component.jsx
@@ -31,11 +31,7 @@ class CountryNdcOverview extends PureComponent {
   renderCompareButton() {
     const { iso } = this.props;
     return (
-      <Button
-        className={styles.exploreBtn}
-        color="white"
-        link={`/ndcs/compare/mitigation?locations=${iso}`}
-      >
+      <Button color="white" link={`/ndcs/compare/mitigation?locations=${iso}`}>
         Compare
       </Button>
     );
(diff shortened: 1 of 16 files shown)


weberjavi:
Although it is not related with changes made in this PR, `descriptionLeft` and `descriptionRight` `classes` are not being used hence outputing a `div` without a `claass`. Maybe would be a good idea adding `className="grid-column-item"` to understand that are columns of the grid?  
In [`country-climate-vulnerability-component`](https://github.com/Vizzuality/climate-watch/pull/449/files#diff-2c800de4f4b289218784c44107e778d2R30)

app/javascript/app/components/country/country-ndc-overview/country-ndc-overview-component.jsx
```diff
@@ -53,13 +57,22 @@ class CountryNdcOverview extends PureComponent {
 
   renderCards() {
     const { sectors, values } = this.props;
+    const renderSubtitle = text => <h4 className={styles.subTitle}>{text}</h4>;
+    const adaptationTitle = (
+      <h4 className={styles.subTitle}>Adaptation contribution</h4>
+    );
     return (
       <div>
```
weberjavi: This `div` is also a column container.  I think is better to add a `class` to identify it as such to avoid eventual deletions in the future.

weberjavi:
Some margin bottom is needed on the `explore NDC content` button.   
image

app/javascript/app/components/country/country-ndc-sdg-linkages/country-ndc-sdg-linkages-styles.scss
```diff
@@ -17,9 +17,15 @@
...
   @include columns(12, $gutters: false);
+  @include xy-gutters($gutter-position: ('bottom'));
+
+  @media #{$tablet-portrait} {
+    @include columns(6, $gutters: false);
+  }
 
   @media #{$tablet-landscape} {
     @include columns(4, $gutters: false);
+    @include xy-gutters($gutter-position: ('bottom'), $gutters: 0);
```
weberjavi: This line avoids a separation between the linkages and the footer.  

image

Bluesmile82:
Ready again! Thanks!

PR #454, opened 2018-03-12
Feature/grid country compare

weberjavi:
This PR addresses grid changes in Country compare page.  
image

image


app/javascript/app/pages/country-compare/country-compare-styles.scss
```diff
@@ -12,3 +9,15 @@
     border-right: 1px solid $theme-border;
   }
 }
+
+.compareWrapper {
```
j8seangel: This starting to be a repetitive task... I had to do the same on all of the other pages so maybe worth it to promote to app level so we don't have to do it over and over again. What do you think?

Bluesmile82:
image
Could you check this gap? The selectors and buttons are aligned in the designs

Bluesmile82:
image
Also we have some extra padding in the mobile selector

app/javascript/app/pages/country-compare/country-compare-styles.scss
weberjavi: It makes sense.  If you agree I'll address it on the ndc's branch I'm working at this moment.

app/javascript/app/components/compare-ghg-chart/compare-ghg-chart-styles.scss
```diff
@@ -1,31 +1,27 @@
...
-
-  grid-template-columns: repeat(2, 1fr);
-  margin-bottom: 10px;
+  margin-bottom: $gutter-padding;
 
   @media #{$tablet-landscape} {
-    @include msGridColumns(1fr, 1fr, 1fr, 1fr, 1fr, 1fr);
-
-    grid-template-columns: repeat(6, 1fr);
+    @include columns((2, 2, 2, 1, 3, 2));
```
Bluesmile82: The Button group seems to have a 2 column size in the design instead of 3. Also could we fill the gap with an offset instead of more columns?

app/javascript/app/components/compare-ndc-content-overview/compare-ndc-content-overview-styles.scss
```diff
@@ -1,26 +1,26 @@
...
 .col6 {
-  @extend %grid;
+  @include columns();
 
-  margin-bottom: 10px;
+  margin-bottom: $gutter-padding;
 
   @media #{$tablet-landscape} {
-    @include msGridColumns(1fr, 1fr, 1fr, 1fr, 1fr, 1fr);
+    @include columns((6, 2, 1, 3));
```
Bluesmile82: I don't really get why we have so many columns here if we only have two elements

app/javascript/app/components/compare-ghg-chart/compare-ghg-chart-styles.scss
weberjavi: Button group does not have a consistent grid lining up on the designs. To avoid buttons to narrow to much i've decided to  give it 3 spaces? Does it makes sense to you.  
image . 
Agree about empty columns and offset.

PR #456, opened 2018-03-12
Fix/compare graph wrong values

Bluesmile82:
The values in the GHG compare graph were not being filtered by the available years of the calculation data (population and gdp). Some refactoring was done in the process
image


j8seangel:
This is what I see 
image
Seems very different, let's review it together tomorrow

PR #460, opened 2018-03-13
Fix Country GHG sectors - Always use AR4 if possible

Bluesmile82:
https://basecamp.com/1756858/projects/13795275/todos/343578930
We were selecting AR2 instead of AR4 by default and the allowed sectors are different for each version so they weren't showing. Now we select AR4 by default.

image


app/javascript/app/components/country/country-ghg-emissions/country-ghg-emissions-selectors.js
```diff
@@ -117,15 +117,18 @@ export const getVersionSelected = createSelector(
   [getVersionOptions, getVersionSelection],
   (versions, selected) => {
     if (!versions || !versions.length) return {};
```
j8seangel: I think returning a null here will improve the cache system of reselect and also avoid to check the `isEmpty` function in next selectors

Bluesmile82:
Ready! Thanks

PR #461, opened 2018-03-13
Feature/Pathways label for empty values

Bluesmile82:
Use 'Not specified' label for empty values in the Pathways tables (pathways, model, scenario).
image

### DIFF
--- a/app/javascript/app/components/table/cell-renderer-component.jsx
+++ b/app/javascript/app/components/table/cell-renderer-component.jsx
@@ -53,7 +53,12 @@ const cellRenderer = ({
   return parseHtml ? (
     <div dangerouslySetInnerHTML={{ __html: cellData }} />
   ) : (
-    cellData || emptyValueLabel || ''
+    cellData ||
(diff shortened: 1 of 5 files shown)


app/javascript/app/components/table/cell-renderer-component.jsx
```diff
@@ -62,8 +62,13 @@ cellRenderer.propTypes = {
   props: PropTypes.shape({
     titleLinks: PropTypes.array, // [ [ {columnName: 'title field name in the table', url:'/destination-url' or 'self'}, ... ] ]
     trendLine: PropTypes.string, // 'field name of the trend line column'
-    parseHtml: PropTypes.bool
+    parseHtml: PropTypes.bool,
+    emptyValueLabel: PropTypes.bool.isRequired
```
j8seangel: emptyValue shouldn't be required if we set the default value as null and should be string typed, should it?

Bluesmile82:
Thanks for the comment. I also added again the styles as all the 'Not specified' from the db will be removed soon

PR #468, opened 2018-03-16
Fix/patways table height

weberjavi:
This PR changes `react-virtualize` based `<Table>` component depending on active columns categories.  
The final approach has been define setter functions for width and heigh values (`setColumnWidth` and `setRowsHeight`) on the container component.   

kapture 2018-03-16 at 16 52 48

### DIFF
--- a/app/javascript/app/components/table/table-component.jsx
+++ b/app/javascript/app/components/table/table-component.jsx
@@ -107,7 +107,6 @@ class SimpleTable extends PureComponent {
                     label={lowerCase(column)}
                     dataKey={column}
                     width={setColumnWidth(column)}
-                    height={50}
                     flexGrow={1}
                     cellRenderer={cell =>
                       cellRenderer({ props: this.props, cell })}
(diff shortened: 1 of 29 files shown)


app/javascript/app/components/table/table-component.jsx
```diff
@@ -84,7 +85,8 @@ class SimpleTable extends PureComponent {
                 width={getResponsiveWidth(activeColumns.length, width)}
                 height={460}
                 headerHeight={headerHeight}
-                rowHeight={rowHeight}
+                // rowHeight={({ index }) => setRowsHeight(activeColumns, data, index)}
```
Bluesmile82: Can we remove this?

app/javascript/app/components/table/table.js
```diff
@@ -35,6 +35,23 @@ class TableContainer extends PureComponent {
...
+      return 150;
+    }
+    return 50;
+  };
+
+  setColumnWidth = column => {
+    if (column === 'year') {
+      return 80;
+    }
+    if (column === 'description') {
```
Bluesmile82: Could you add 'description' to the fullTextColumns in models in the selectors file. As we are displaying it also there it might get some more height instead of the ellipsis

PR #470, opened 2018-03-19
quantification text display

weberjavi:
This PR addresses the issue described in [this basecamp thread](https://basecamp.com/1756858/projects/13795275/todos/338709068). On GHG emissions chart (on Edge browser) reference dot label was not displaying.  That was caused by the lack of support of [`paint-order`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/paint-order) attribute of SVG on Edge browser.  
The approach taken has been to create a flag that conditionally handles the `stroke-width`  of the label text to avoid that the white `stroke` covers the colored `fill`.  In other browses the original design is respected.

### DIFF
--- a/app/javascript/app/components/charts/stacked-area/stacked-area-component.jsx
+++ b/app/javascript/app/components/charts/stacked-area/stacked-area-component.jsx
@@ -92,6 +92,7 @@ class ChartStackedArea extends PureComponent {
     if (!data.length) return null;
 
     const maxData = getMaxValue(data, config);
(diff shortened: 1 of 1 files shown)


weberjavi:
Addressed changes requested.  
Now we have a `isMicrosoftBrowser` function in `utils`    `¿`💻`?`

Bluesmile82:
Maybe its ok to leave it like isEdge or isIE11? Not very important right now though

j8seangel:
Thanks!

PR #471, opened 2018-03-19
Fix/pathways graph scale issue

Bluesmile82:
https://www.pivotaltracker.com/story/show/156008127
- Connect nulls again in all charts
- Fix n/a in 0 values
- Sort Compare graph data to fix some display error (This was already added)
- Use 4 decimals in the tooltip for Pathways chart
- Calculate precision needed within the domain of the values of the Pathways chart
- Change the format to always display the needed precision in the Y-axis ticks
- Apply margin offset needed for that precision
- Add margins to y-domain to represent the little difference between close values

image


j8seangel:
I've found this small bug when we have long axis values.
image

Could you please review it? Thanks!

PR #473, opened 2018-03-21
Feature/add tooltip in chart

weberjavi:
This PR adds tooltips to MY-CW charts.  
It adds a `<Tooltip>` component to be used as content of Recharts tooltip.  
A `tooltip` field has been created inside charts `config` object.  
A `theme` has been created to share styles between ClimateWatch and My-CW chart tooltips.    

kapture 2018-03-21 at 16 17 50


Bluesmile82:
It would be nice to order the options in the tooltip by value as we have in the other charts
image

Bluesmile82:
There is something wrong with the values here. They are not changing when we change the year
image
image

PR #474, opened 2018-03-22
Feature/add labels and titles

Bluesmile82:
This PR corrects titles adds subtitles as labels and also labels for units
https://basecamp.com/1756858/projects/13795275/todos/346005337
 - Update all charts to include the unit label on top of the y-axis (except for pie charts)
 - If the chart has multiple indicators the title should be only the category and the subcategory should be in the Y vertical axis label (in the case of the pie chart is the top label)
image

 - If the chart has only one indicator the title should be category - subcategory and the indicator name should be on the Y vertical axis label
image

 - Don't show any label or axis in the small view
---
Extra: 
- Clicking on Create new visualization was sometimes editing the last visualization: Clear id and description when the payload is null

Question: Should we change 'visualisation' for 'visualization' as its the American English and more used spelling?


app/javascript/app/components/my-climate-watch/viz-creator/viz-creator-selectors.js
```diff
@@ -33,6 +33,7 @@ export const categoriesSelector = state => get(lenses.$categories, state);
 export const subcategoriesSelector = state => get(lenses.$subcategories, state);
 export const yearsSelector = state => get(lenses.$years, state);
 export const timeseriesSelector = state => get(lenses.$timeseries, state);
+export const getTitle = state => state.title || null;
```
weberjavi: Maybe `titleSelector` would be a better name to follow the naming convention of the other selectors, what do you think?

j8seangel:
Could you please review the conflicts and this warning?

`Warning: React does not recognize the `verticalAnchor` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `verticalanchor` instead. If you accidentally passed it from a parent component, remove it from the DOM element.`

Rest LGTD, thanks!!

Bluesmile82:
Conflicts fixed 👍

j8seangel:
Thanks, it works perfectly!
The branch was broken and I included a small fix as I had a previous version of the widget so I've included a small check before to create the labels in the pieChart.

PR #480, opened 2018-03-26
Remove map from embed and fix embed styles

Bluesmile82:
https://basecamp.com/1756858/projects/13795275/todos/345725109
- Remove the map from the embed view
- Add layout to embed content always and some margin-top too
image


j8seangel:
The drop-down wasn't visible in the embed version, I've included the small fix. Thanks!

PR #484, opened 2018-04-02
Feature/country linkages redirect to ndc

Bluesmile82:
This PR links to the related NDC Full document view with the highlighted text from the Country - NDC SDG Linkages dot.

- Add document_type and language to the target_serializer so we can see the related document and link to it from each dot (try: http://localhost:3000/api/v1/ndcs/DZA/sdgs)
- Link from the dot to the related document
- Fix styles and loading in the Country - NDC SDG Linkages

image
image

### DIFF
--- a/app/javascript/app/components/sdg-card/sdg-card-component.jsx
+++ b/app/javascript/app/components/sdg-card/sdg-card-component.jsx
@@ -20,7 +20,8 @@ class SDGCard extends PureComponent {
       icons,
       hover,
       onClick,
-      onMouseEnter
+      onMouseEnter,
+      handleOnDotClick
     } = this.props;
     const cardStyle = cx(
       styles.card,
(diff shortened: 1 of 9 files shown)


app/javascript/app/components/sdg-card/sdg-card-component.jsx
```diff
@@ -58,11 +59,25 @@ class SDGCard extends PureComponent {
                 targetData.targets[target.number].sectors;
               return (
                 <span
```
j8seangel: We are using a span with a javascript event to go to another page... it makes sense in a map for example, but in a static markup element we could use directly an <a> element, couldn't we?
Also is much better for accessibility, so could we move this to a `Link` component from react-router?

j8seangel:
Good job!

PR #489, opened 2018-04-04
Fix/linkages link to ndc full

Bluesmile82:
The Linkages links were sometimes not redirecting to the right document because they hadn't the info about the document_type and language of the target or goal.

- Add document_type and language to the http://localhost:3000/api/v1/ndcs/sdgs_overview response inside locations - iso
- Adapt the map for the new response format
- Change the link to redirect to the right document

Try Mauritania with goal 6 (before it couldn't scroll to the right linkage)

image
image


app/javascript/app/components/ndc-sdg/ndc-sdg-linkages-map/ndc-sdg-linkages-map-component.jsx
```diff
@@ -29,7 +29,7 @@ class NdcSdgLinkagesMap extends PureComponent {
       this.props.goal &&
       this.props.goal.locations &&
       this.props.goal.locations[country.id];
-    const locations = targets && targets.length;
+    const locations = targets && targets.numbers.length;
```
j8seangel: Double check we have numbers before checking the length, please!

PR #491, opened 2018-04-05
Staging - Search component should have the query as a 'value' prop

Bluesmile82:
In NDC search the searched query from the NDC content page was not appearing in the search box
This happened because sometimes we used input instead of value for the search component.
I will create another PR for develop to standardize the prop name as 'value'

image


j8seangel:
We don't need to update it in staging so as it was already fixed on develop an my-cw I'm going to close this.

PR #499, opened 2018-04-09
Fix not aligned tooltip in ESP and use only 3 decimals

Bluesmile82:
The tooltip had some unaligned values on the Pathways tooltip. Also, 3 decimal numbers instead of 4 were requested.
This is based in My CW branch as we didn't have this problem in develop.

image


simaob:
Javi can you check the latest commits and merge if it all looks good? Thanks!

PR #506, opened 2018-04-11
Filter vulnerability section and anchor link if is contained

Bluesmile82:
This PR removes the vulnerability section from the country compare page if it's contained
try http://localhost:3000/contained/countries/compare?locations=USA%2CAND

image

### DIFF
--- a/app/javascript/app/pages/country-compare/country-compare-component.jsx
+++ b/app/javascript/app/pages/country-compare/country-compare-component.jsx
@@ -58,7 +58,7 @@ const CountryCompare = ({ route, anchorLinks, isContained }) => {
           {route.sections &&
             route.sections.length > 0 &&
             route.sections
-              .filter(s => !(isContained && s.hash === 'climate-vulnerability'))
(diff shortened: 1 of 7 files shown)


app/javascript/app/pages/country-compare/country-compare-component.jsx
```diff
@@ -57,11 +57,17 @@ const CountryCompare = ({ route, anchorLinks }) => {
           </Desktop>
           {route.sections &&
             route.sections.length > 0 &&
-            route.sections.map(section => (
-              <section key={section.hash} id={section.hash}>
-                {!!section.component && <section.component />}
-              </section>
-            ))}
+            route.sections
+              .filter(s => !(isContained && s.hash === 'climate-vulnerability'))
```
simaob: Instead of doing the filtering like this by the section hash, could we add it to the routes config a flag like "includeInContained", here:

`app/javascript/app/routes/app-routes/country-compare-sections/country-compare-sections.js`

This way if we need to add more sections and need to exclude them we can use that config option more easily instead of changing that direct comparison.

PR #509, opened 2018-04-12
Feature/compare sector selector

Bluesmile82:
This PR adds the sector filter for the emissions country compare chart:

- Add Multiselect selector for sectors
- Update filter to display correct data
- Force to select all the sectors when the selected calculation is not 'Absolute value'
- Fix z-index in country selector

image

---
Extra:
- Fix misplaced loader
- Fix NaN on country emissions chart tooltip Total (country page, not compare)
Before: 
image


simaob:
Couple of things:

- please sort the sectors list alphabetically;
- why are we disabling the sectors selection when using the PER_GDP and PER_POPULATION calculations?

simaob:
@Bluesmile82 please check my comments, see if we can get this onto staging today.

Bluesmile82:
Ready again! Fixed some loader styles too

simaob:
Tks! =)

PR #510, opened 2018-04-12
Fix/ndc search map documents

Bluesmile82:
This PR fixes the map and text of the NDC Search page:

- The Map reflects the number of countries that match the search
- The text shows the total number of documents in the db depending whether ndc, indc or all is selected
- Add a backend test for the meta response

image
image


simaob:
When I first the first time the "total" value doesn't seem to be filled. It shows a dash instead of the total documents. It works after I switch between the three options.

Bluesmile82:
Good idea!

PR #513, opened 2018-04-17
Update SEI Logo and add description

simaob:


weberjavi:
Nice Vizzuality green redesign ✌️

PR #515, opened 2018-04-17
Update title on save

weberjavi:
Sets placeholder as title on first visualisation save avoiding `empty name` warning


weberjavi:
Updated naming flow, please check the changes.

PR #516, opened 2018-05-07
Fix/scenario page indicator values

Bluesmile82:
Try: http://localhost:3000/pathways/scenarios/49
- The indicator trend data was not ordered by default in some cases
- Shorten a bit the width of some columns

image


app/javascript/app/components/table/table.js
```diff
@@ -45,10 +45,12 @@ class TableContainer extends PureComponent {
   };
 
   setColumnWidth = column => {
-    if (column === 'year') {
+    const narrowColumns = ['year', 'unit', 'first', 'last'];
```
weberjavi: Maybe we can move columns names to a constant as data used to set rows height in `setRowsHeight` function?

simaob:
ready again, Javi.

PR #520, opened 2018-05-08
Added http basic auth for staging

agnessa:


app/controllers/application_controller.rb
```diff
@@ -1,6 +1,11 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
+  if Rails.env.staging?
```
simaob: I think that at this moment we are using the `production` env on staging... though we should probably change it to use `staging`. Otherwise this won't work. (See: https://github.com/Vizzuality/climate-watch/pull/520/files#diff-b194a7142a89214916bc627216ad2017R52).

If we do change it we will have to update all the config files of course to have it!

agnessa: I changed that; I'm not sure which config files still need amending?

PR #526, opened 2018-05-09
Fix/dots categorisation

weberjavi:
This PR addresses some issues on GHG emissions chart:  
-  Changes svg quantifications dots ordering to avoid z-index problem with it's labels.  
`test it here:` http://localhost:3000/countries/TUR
-  Adds functionality to make `null` quantifications values to behave as `range` values (that is to be represented as a line instead of as a dot) . 
`test it here:` http://localhost:3000/countries/ECU  
-  Changed yellow shade used on `quantified targets` to be the same as the base theme yellow (`$theme-secondary`) . 
This last change follows a suggestion that could be checked in the last comment of this [basecamp thread](https://basecamp.com/1756858/projects/13795275/todos/350625874)


weberjavi:
@Bluesmile82 the PR is now ready to review after the `reimport`, thanks!

PR #528, opened 2018-05-09
Fix/quantification label position

Bluesmile82:
- Wrap the text in long labels

image

### DIFF
--- a/app/javascript/app/components/charts/stacked-area/stacked-area-component.jsx
+++ b/app/javascript/app/components/charts/stacked-area/stacked-area-component.jsx
@@ -5,8 +5,7 @@ import min from 'lodash/min';
 import max from 'lodash/max';
 import isArray from 'lodash/isArray';
 import { getCustomTicks } from 'utils/graphs';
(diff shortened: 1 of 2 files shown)


app/javascript/app/components/charts/stacked-area/stacked-area-component.jsx
```diff
@@ -226,26 +225,52 @@ class ChartStackedArea extends PureComponent {
...
+                  strokeWidth={isEdgeOrExplorer ? 0 : 8}
+                  style={{ paintOrder: 'stroke', zIndex: 500 }}
+                  fontSize="13px"
+                  offset={offset}
+                />
+              );
+              const extraLabel = wordWrap(
+                point.label,
+                MAX_LINE_LENGHT
+              ).map((l, i) => extraLabelLine(l, LABEL_OFFSET + i * DY));
```
weberjavi: Is kind of cryptic whats happening in the `offset`  parameter, wouldn't it be possible to set a fixed value?   
What`s the aim of using the `index`?

PR #531, opened 2018-05-14
Fix/negative quantification color

Bluesmile82:
https://www.pivotaltracker.com/story/show/157551941

- Show Negative BAU points in the right color
- Frame Not quantifiable range into data minimum and maximum
image
image

https://basecamp.com/1756858/projects/13795275/todos/332083809
- GHG emissions: Smooth the chart lines in all cases except for the UNFCCC data
image

https://basecamp.com/1756858/projects/13795275/todos/349822479
- Smooth lines in ESP (Using monotone type). It can't be smoother while still being accurate.
image

Extra:
- Huge needed refactor extracting the logic of the stack-area-component, creating selectors and breaking the different parts in the component (last point, quantification points, labels, ...)
- Dockerfile conflicts with sandbox fixed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,8 @@ ENV RACK_ENV $RAILS_ENV
 ENV NODE_ENV $RAILS_ENV
 
 ENV ESP_API https://data.emissionspathways.org/api/v1
+ENV CW_API /api/v1
+ENV GFW_API https://production-api.globalforestwatch.org
 ENV S3_BUCKET_NAME climate-watch-dev
 ENV GOOGLE_ANALYTICS_ID UA-1981881-51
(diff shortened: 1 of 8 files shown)


Bluesmile82:
Thanks @simaob, that is fixed in [Fix domain commit](https://github.com/Vizzuality/climate-watch/pull/531/commits/d849d2117fb4f243164900925848f91df025cf97)

simaob:
Most issues are gone, but there's one small one with mismatch between the graph and the legend, for these countries:

- http://localhost:3000/countries/BTN
- http://localhost:3000/countries/MDV


the legend for "Non quantifiable target", is not showing.

PR #536, opened 2018-05-16
Fix/anchornav active styles

weberjavi:
This PR fixes `anchor nav` component to update selected item styles on scroll. An `activeSection` field has been added to the `store` (it's `actions` and `reducers` have been defined in `anchor-nav` component) and scrolling behaviour is triggered with the help of [`react-waypoint`](https://github.com/brigade/react-waypoint)   

kapture 2018-05-16 at 13 36 13


app/javascript/app/pages/country-compare/country-compare-component.jsx
```diff
@@ -46,13 +47,19 @@ const CountryCompare = ({ route, anchorLinks }) => (
...
             .filter(s => !(isPageContained && s.excludeFromContained))
             .map(section => (
-              <section
+              <Waypoint
+                bottomOffset={'60%'}
+                topOffset={'30%'}
+                onEnter={() => {
+                  setActiveSection(section.hash);
+                }}
+                fireOnRapidScroll={false}
```
Bluesmile82: This still is not working when we select the socioeconomic anchor. Its because the height of the content is too little and it goes back to the second anchor. Can you think about some workaround. Maybe tweaking the topOffset?

weberjavi: @Bluesmile82 ready to review ✋

PR #537, opened 2018-05-16
Use time scale in xAxis

Bluesmile82:
https://www.pivotaltracker.com/story/show/157552161
This PR fixes a problem with the scale of the X axis in the line chart visualizations in My CW

- Use time scale in the X axis

Before: 
image

After:
image
image


juancarlosalonso:
That graph looks awesome :)
Just one thing: why is it showing 2011 instead of 2010?

PR #545, opened 2018-05-25
Feature/vis creator country selector filter

weberjavi:
[basecamp thread](https://basecamp.com/1756858/projects/13795275/todos/348244860)

This PR updates My CW [visualizations creator](http://localhost:3000/my-climate-watch) selectors behaviour.   
The new behaviour should be as follows:  
When creating a visualization, on charts where more than one country or region could be selected (`Compare countries/regions`) when the first `location` is selected only the countries that have some `model` in common with this location would be displayed. Once a second `location` is selected also the available `models` will be filtered to display only the common `models` between the selected `locations`.


j8seangel:
Was amazing to pair programming together again 💪 !

PR #546, opened 2018-05-25
Sandbox

j8seangel:
- A lot of @simaob things
- The deploy fix


j8seangel:
Merging only the deploy fixes here: https://github.com/Vizzuality/climate-watch/pull/547

PR #554, opened 2018-06-01
Feature/viscreator selections cache

weberjavi:
[Basecamp thread](https://basecamp.com/1756858/projects/13795275/todos/349879565) .  

This PR implements a cache to avoid that when is a change on user selection of a given field on Visualisations Creator all the fields should be filled again. 
So for example if a user adds a second (third...) country in a multi-location chart the subsequents fields (models, scenarios, categories...) will display the cached selection.
-  It updates the filter logic on the selectors and moves it to the reducers.
-  It also adds some info tooltips explaining the behaviour of some filtered selectors.
https://basecamp.com/1756858/projects/13795275/todos/354837515


weberjavi:
@j8seangel this PR is ready to review, thanks 😅

simaob:
Closed in favour of: https://github.com/Vizzuality/climate-watch/pull/559/files

PR #556, opened 2018-06-04
Add more nested options on NDC nav

simaob:
This PR moves us closer to the new version of the top menu as proposed here: https://basecamp.com/1756858/projects/13795275/todos/344949987#comment_622956393

For now we have just added the: NDC Comparison, NDC Explorer and NDC Search.


weberjavi:
@simaob let me know if you want me to take this PR from here 🖐

PR #558, opened 2018-06-08
feature/compare-button

pjosh:
Adding a compare button on top of the countries searcher
captura de pantalla 2018-06-08 a las 11 00 34


simaob:
There seems to be a bug here. When I click on the compare button, and then I select a location on the first selector for locations, I get a javascript error that breaks the page. If after that I refresh the page it works. So there might be something not being initialized.

weberjavi:
@pjosh , @simaob I've added some conditions on `componentDidUpdate` of NDCs content overview Provider to avoid the crash.
Everything is looking fine from my perspective.

PR #563, opened 2018-06-11
Feature/mycw user profile improvement

weberjavi:
[Basecamp thread](https://basecamp.com/1756858/projects/13795275/todos/354675028)

This PR adds `name`, `organization` and `area of work` to `users` on MYCW and allows users to update this personal info on [account settings](http://localhost:3000/my-climate-watch/account-settings) tab of MYCW.

Bonus: 
-  This PR also adds the [CrazyEgg](https://www.crazyegg.com/) script to the app `headers` to allow research team to do some user research.

kapture 2018-06-12 at 17 02 16


simaob:
Yeah it's a bit weird that travis is failing it worked locally.

Anyway the feature seems to work, but it would be good to display a message on the interface when the profile has been updated, so that users know that it worked.

db/migrate/20180612074749_add_area_of_work_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddAreaOfWorkToUsers < ActiveRecord::Migration[5.1]
+  def change
+    add_column :users, :areaOfWork, :string
```
simaob: we should use Rails conventions here and call this field: `area_of_work` using snake_case instead of camelCase

agnessa:
@weberjavi for now just manually remove the line `SET idle_in_transaction_session_timeout = 0;` from structure.sql; this is a problem with incompatible versions of postgres.

weberjavi:
@simaob , @Bluesmile82 this PR is ready to be reviewed again. thanks!

simaob:
There seems to be something weird with the text area we are using :/ it skips some characters.

weberjavi:
@Bluesmile82 , @simaob i think it looks better now.

simaob:
Yes much better! =)

app/javascript/app/components/text-input/text-input-component.jsx
```diff
@@ -1,33 +1,87 @@
...
-);
+class InputComponent extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: props.value
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
```
Bluesmile82: It will be interesting to read this article: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html
Also, componentWillReceiveProps will be deprecated at some point (React 17 as it seems) and in react 16.4 getDerivedStateFromProps is preferred. No big deal right now as we are using 16.2. But maybe something to think about in the future

app/javascript/app/providers/login-provider/login-provider-reducers.js
```diff
@@ -4,24 +4,82 @@ export const initialState = {
...
+  return {
+    ...state,
+    user: { ...user, user_id: { ...newUserData } },
+    profileUpdated: false
+  };
+}
+
 export default {
-  [actions.getUserInit]: state => ({ ...state, loading: true }),
+  [actions.getUserInit]: state => ({
```
Bluesmile82: Great new reducers style!

app/javascript/app/providers/login-provider/login-provider-actions.js
```diff
@@ -17,3 +25,15 @@ export const getUser = createThunkAction('getUser', () => (dispatch, state) => {
...
 });
+
+export const saveUserData = createThunkAction(
+  'saveUserData',
+  () => (dispatch, getState) => {
+    const { login } = getState();
+    CWAPI.patch(`my_cw/users/${login.user.user_id.id}`, login.user.user_id)
+      .then(dispatch(profileUpdated()))
+      .catch(e => {
+        console.warn(e);
```
Bluesmile82: It would be nice to show some error too when the profile fails to update. I think its even more valuable than the confirm that everything worked

app/javascript/app/components/my-climate-watch/my-account/my-account-component.jsx
```diff
@@ -1,16 +1,148 @@
...
+      <h1 className={styles.title}>User Profile</h1>
+      <div className={styles.personalInfo}>
+        <TextInput
+          className={styles.input}
+          theme={theme}
+          placeholder={'Add a name'}
+          label={'First name'}
+          inputType={'text'}
+          onBlur={updateUserFirstName}
+          value={firstName}
```
Bluesmile82: We could add a value={firstName || ''} to avoid the warning message. The input doesn't like null or undefined as a value

app/javascript/app/components/my-climate-watch/my-account/my-account-styles.scss
```diff
@@ -0,0 +1,62 @@
...
+.input,
+.dropdown,
+.textarea,
+.optIn,
+.updateButton {
+  margin-bottom: $gutter-padding;
+}
+
+.textarea {
+  height: 200px;
```
Bluesmile82: Could we add some padding-top: 1em; or maybe a little bit less to keep that gap from the top of the input. Maybe we have to check in other browsers but in Chrome it feels a bit weird

PR #568, opened 2018-06-13
Feature/data explorer feature flag

Bluesmile82:
This PR starts the Data Explorer:
- Create feature flag (to try it add FEATURE_DATA_EXPLORER=true to .env)
- Create Data Explorer page, header and menu
- Fix mobile lower menu adding the MyCW and Data Explorer to the main menu except for desktop
- Fix the active highlighting in the tools menu

Note: In the tools-nav-component I use the withRouter to pass the location as it won't rerender properly with the location obtained from the isActive method of the NavLink

image


simaob:
Hi Álvaro, probably worth adding the flag right away to the Jenkinsfile and Dockerfile. =)

PR #569, opened 2018-06-14
Add crazyegg script to app headers

weberjavi:
This PR adds [crazzyegg](https://www.crazyegg.com/) script to app header to allow User Research team to get navigation info from user base.  
I decided to do an independent PR with this tasks to avoid couple it with #563 where this script was already added, that way it will land sooner into `master` branch.


simaob:
It doesn't break, so I guess it's OK =D

PR #572, opened 2018-06-14
Change "no data" to "not applicable" in NDC map legend and tooltip

agnessa:
After a [lot of deliberation](https://basecamp.com/1756858/projects/13795275/todos/347473811#comment_628280529) we're making these changes to the NDC map legend:
- change "No data" to "Not Applicable"
- change the colour of "No Document Submitted" to a less prominent one (and fixed across the maps)

Requires `bundle exec rake indc:import`

### DIFF
--- a/app/javascript/app/utils/map.js
+++ b/app/javascript/app/utils/map.js
@@ -48,9 +48,10 @@ export function getColorByIndex(data, index) {
 
 export function createLegendBuckets(locations, labels, isos) {
   if (Object.keys(locations) === isos) return labels;
-  const notSubmitted = Object.keys(labels).filter(l => labels[l].index === -2);
+  // An index of -2 is applied in the backend to 'No Data Submitted'
+  const notSubmitted = Object.keys(labels).find(l => labels[l].index === -2);
   if (notSubmitted) {
-    const notApplicableKey = notSubmitted + 1;
+    const notApplicableKey = parseInt(notSubmitted, 10) + 1;
     return {
       ...labels,
       [notApplicableKey]: { name: 'Not Applicable', index: 0 }


weberjavi:
@agnessa the `index` option could work. Currently the logic on the front-end side decides which colour to pick by index, so if you pass a known index to 'No Document Submitted' elements we could give them the colour we decide and not add this colour to the map's colour buckets.
The current buckets we are using do not exceed ten colours so any index bigger than 9 will work (maybe bigger than 20 to ensure that we do not have to change that piece if the indicators in the map grow).
What do you think?

agnessa:
Would a negative number work @weberjavi?

weberjavi:
sure @agnessa

agnessa:
No Document Submitted is now at index -1. It's automatically assigned a grey colour across, slightly lighter than "Not Applicable".

screen shot 2018-06-18 at 12 47 15

I think the colour should be more easily distinguishable from "Not Applicable", and ideally they should appear together in the legend - what do you think @weberjavi ?

weberjavi:
@agnessa maybe `-1` is not the best choice. The javaScript method `indexOf` returns `-1` when the element that is being searched in the array does not exists, that could lead to some misunderstanding to the developer reading the frontend code if we search by `index` property with `-1` value.  
Sorry, I should have had this into account. 🙇

weberjavi:
I could branch out from this PR and make the frontend changes and PR back to this branch, to have the frontend stuff in a different branch, what do you think?

agnessa:
Works for me! Thanks!

agnessa:
screen shot 2018-06-19 at 15 03 43

weberjavi:
@simaob I'm going to make a couple of changes suggested by @Bluesmile82. I'll let you know when it is ready

PR #579, opened 2018-06-19
Colours and homepage updates and other small request

simaob:
This PR makes the following changes:

- Update the colours to the new colour palette as per this basecamp thread: https://basecamp.com/1756858/projects/13795275/todos/354355432#comment_629274897
- Hide monkey when no data is available, it will only show now for the 404 page :( as per this basecamp thread: https://basecamp.com/1756858/projects/13795275/todos/352505208#comment_629157648
- Update text on the home page for the different sections as per the same thread as the monkey one.

One thing that would be good to review is my change to the **br** styles to increase it's line-height as they added multi paragraph text to the homepage sections, I'm not sure if this is the right solution, so would be great to get your approval.

Once this is all good and on staging, @faustoperez will then be able to take some new screenshots to update the homepage!

Thanks!

### DIFF
--- a/app/javascript/app/components/intro/intro-styles.scss
+++ b/app/javascript/app/components/intro/intro-styles.scss
@@ -38,6 +38,10 @@
   margin-bottom: 20px;
   line-height: $line-height-medium;
   color: $white;
+
+  br {
+    line-height: 200%;
+  }
 }
 
 .disclaimer {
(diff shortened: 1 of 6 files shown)


app/javascript/app/components/intro/intro-styles.scss
```diff
@@ -38,6 +38,10 @@
   margin-bottom: 20px;
   line-height: $line-height-medium;
   color: $white;
+
```
weberjavi: It actually looks nicer, but I'm not really sure if that was what they were asking for.
That's what Ola said: `On the homepage, the line break just below 'View Country Profiles' needs spacing, to be consistent with the rest.`
There's no `br` under 'View Country Profiles'.
Any way it could stay like this.

simaob: That point I didn't really understand... because of what you said. But I used the BR in the other sections that were updated now.

PR #584, opened 2018-06-19
Instead of shrinking them, hide the dots that dont match

simaob:
Following a design decision to simplify the filtering on the NDC-SDG component of the country page, I went ahead and replaced the shrinking with hiding of the dots when they are not a match.


weberjavi:
@simaob i've changed `opacity` instead of `display` to address a [comment of this thread](https://basecamp.com/1756858/projects/13795275/todos/340674443#comment_629429064) (allowing user to have a grasp of the amount of linkages that are not present).
How do you feel of adding a an info button and a tooltip like in `mycw` selectors to address the first question of the comment?

simaob:
Yeah, they've been asking that for ages so yeah, I think we should do it. Are you happy to find a place for it? Thanks

weberjavi:
@simaob, @Bluesmile82 please review the changes i've made to this PR.
I've updated `Dropdown` and `Multiselect` components to accept `info` and `infoText` props to add an info icon with an explanation.

simaob:
Looks good to me! =)

PR #585, opened 2018-06-20
Feature/data explorer metadata

Bluesmile82:
This PR shows the metadata depending on the selected source:

- Fetch methodology metadata
- 'Raw Data - Methodology' menu with hashes
- Fetch metadata for the different filters depending on the section
- Create selectors and change URL when the selection changes
- Select metadata depending on the selected source

Note: Still no filtering to the data is applied

image

### DIFF
--- a/app/javascript/app/providers/data-explorer-provider/data-explorer-provider-actions.js
+++ b/app/javascript/app/providers/data-explorer-provider/data-explorer-provider-actions.js
@@ -2,20 +2,25 @@ import { createAction } from 'redux-actions';
 import { createThunkAction } from 'utils/redux';
 import { DATA_EXPLORER_SECTION_NAMES } from 'data/constants';
 import isEmpty from 'lodash/isEmpty';
+import { parseLinkHeader } from 'utils/utils';
 
 export const fetchDataExplorerInit = createAction('fetchDataExplorerInit');
 export const fetchDataExplorerReady = createAction('fetchDataExplorerReady');
 export const fetchDataExplorerFail = createAction('fetchDataExplorerFail');
 
-export const fetchInitialMetadataInit = createAction(
-  'fetchInitialMetadataInit'
+export const fetchSectionMetadataInit = createAction(
+  'fetchSectionMetadataInit'
 );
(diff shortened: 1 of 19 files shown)


simaob:
Not sure where the source content is coming from, but there should be a change, that maybe we won't do here only but needs to be a bit broader.

As per this thread: https://basecamp.com/1756858/projects/13795275/todos/348881975#comment_614177959, the way the sources should be organised is:

CAIT (AR2)
PIK - AR4
PIK - AR2
UNFCCC - Annex I (AR4)
UNFCCC - non Annex I (AR2)

But this I believe will require a reorganising of the backend data.

simaob:
Other than that the metadata is showing =)

app/javascript/app/utils/utils.js
```diff
@@ -137,6 +137,40 @@ export const wordWrap = (long_string, max_char) => {
   return split_out;
 };
 
+const unquote = value => {
+  if (value.charAt(0) === '"' && value.charAt(value.length - 1) === '"') { return value.substring(1, value.length - 1); }
+  return value;
+};
+
+export const parseLinkHeader = header => {
```
weberjavi: 😱

app/javascript/app/providers/data-explorer-provider/data-explorer-provider-actions.js
```diff
@@ -40,9 +56,100 @@ export const fetchDataExplorer = createThunkAction(
...
+        }
+      })
+      .catch(error => {
+        console.warn(error);
+        dispatch(fetchSectionMetadataFail());
+      });
+  }
+);
+
+export const fetchMetadata = createThunkAction(
```
weberjavi: Quite a complex fetching staff going on. Hope you didn't lose the nerve 😅

app/javascript/app/utils/utils.js
weberjavi: good old Regex

simaob:
I've created a different story on pivotal tracker to look at the sources more generally, so I think this can be merged now!

PR #588, opened 2018-06-21
Feature/data explorer filters

Bluesmile82:
This PR filters the data fetch and download:
- Parse the source + IPCC version into separate ids for the fetch
- Pick the query and parse it to fetch the filtered data
- Update the download URL to download the filtered data
- Fix some problems with the filters

image


simaob:
We'll need to include the Goal number on the linkages Goals dropdown.

Bluesmile82:
Just fixed some problems and included the goal number to the Goal filter

PR #591, opened 2018-06-22
Add default image to stories

weberjavi:
This PR adds a default image to Stories (in case imported story does not have an image URL)
[Pivotal](https://www.pivotaltracker.com/story/show/158499552)

Credit free selected image: https://unsplash.com/photos/Q1p7bh3SHj8


simaob:
This doesn't seem to be working as it should, please see: https://climate-watch.vizzuality.com/ there's a story with no image :(

PR #599, opened 2018-06-26
Add a new "real" Overview tab and rename old overview to summary

simaob:
They updated the data to include an actual overview category and
moved the old tab that was called overview to Summary tab, and this one
includes the same data present in the country page NDC Content.

This also meant changing the "notOverview" to "notSummary" as the search is now needed on the overview but not on the summary tab.


simaob:
I've added an extra fix here to a link to the country ndcs which was broken.

simaob:
Updated the title.

PR #601, opened 2018-06-27
Fix/chart axis

weberjavi:
[Pivotal](https://www.pivotaltracker.com/story/show/158499735)
This PR fixes some problems related to `y axis`:    
Avoid displaying same values when filtering by `per Capita`  on emissions chart
[Basecamp](https://basecamp.com/1756858/projects/13795275/todos/355575533)  

Add a buffer to `line charts` when smaller values are over `0` (to avoid confusion because ofthe chart line being in contact with the x axis). This solution has been implemented on Country GHG emissions, Historical emissions, MyCW viscreator and Pathways charts.
[Basecamp](https://basecamp.com/1756858/projects/13795275/todos/352513391)


app/javascript/app/utils/graphs.js
```diff
@@ -102,6 +102,19 @@ export const getColorPalette = (colorRange, quantity) =>
...
+  return [setBuffer, dataMax => dataMax];
+}
+
+function setBuffer(min) {
+  if (min <= 0) return 'auto';
+  return min * 0.7;
+}
+
+export function setXAxisDomain() {
+  return [dataMin => dataMin, dataMax => dataMax];
```
Bluesmile82: Do we need functions here? I think 'dataMin' string should be enough

app/javascript/app/components/country/country-ghg-emissions/country-ghg-emissions-selectors.js
```diff
@@ -345,12 +345,14 @@ export const getChartConfig = createSelector(
...
     const tooltip = getTooltipConfig(yColumnsChecked);
     let unit = DEFAULT_AXES_CONFIG.yLeft.unit;
+    let precision = null;
     if (calculationSelected.value === CALCULATION_OPTIONS.PER_GDP.value) {
       unit = `${unit}/ million $ GDP`;
     } else if (
       calculationSelected.value === CALCULATION_OPTIONS.PER_CAPITA.value
     ) {
       unit = `${unit} per capita`;
+      precision = 2;
```
Bluesmile82: I'm not sure if this is what they want. I will expect the axis to be rounded as it was and the tooltip to show decimals. But right now the decimals are always 0 so it won't help either

app/javascript/app/utils/graphs.js
```diff
@@ -102,6 +102,19 @@ export const getColorPalette = (colorRange, quantity) =>
 
 export const darkenColor = color => chroma(color).darken();
 
+export function setYAxisDomain() {
+  return [setBuffer, dataMax => dataMax];
+}
+
+function setBuffer(min) {
```
Bluesmile82: It would be nice to have an example of when these buffers apply. Could we check it when you come to the office?

weberjavi: @alvaro I've addressed your comments in the PR (moved the floating point values from the axis to the tooltip and used the string instead of the function on the domain).   
About the `buffer` it adds some space under the smaller value of the line chart to avoid confusion with the zero value (it adds a 30% percent of the minimal value on the bottom of it).

You can check it here: 
http://localhost:3000/ghg-emissions?breakBy=location&filter=GBR%2CZAF&source=10&version=1

https://climate-watch.vizzuality.com/ghg-emissions?breakBy=location&filter=ZAF%2CGBR&source=63&version=1

Bluesmile82: image
I can't see the buffer working. Lets check it when you are back

weberjavi: It is working (otherwise the lower value of the chart would stick to the `x-axis`), what happens is that because of the great difference between higher and lower chart values the buffer seems too small. Maybe would be a good fix to calculate the buffer taking into account higher and lower values.
Let's discuss the approach having in mind that [this task](https://basecamp.com/1756858/projects/13795275/todos/352513391) does not have a clear definition of whats the intended behaviour.

simaob: If we lack clarity let's not dwell too much on it. Fix the key issues and then when replying to them ask for further clarity if they have it or say that we are happy to discuss in more detail if that helps.

PR #602, opened 2018-06-27
Link to contained route and not directly to ndcpartnership

simaob:
This was a bug introduced when we added more shared modules from the
country page.

We were wrongly linking from various Explore buttons to the
ndcpartnership pages, causing a double render of the page, given that
the links are followed from inside the iframe.

This change makes it so that the links used are for routes inside
Climate Watch, with the prefix of /contained and the flag isNdcp=true to
be used by the share buttons, though maybe that is also not needed.


weberjavi:
@simaob after some back and forth :>/ I think this should be ready.
Indeed the `isNdcp=true` flag is needed. I just removed the `goal` filter from the `ndc-sdg-linkages` link (because we were using the active sector to set the filter and there are no goal matching those filters).
Please check the last commit... after the vizzcamp and the long weekend i'm a little bit dazed 😑

PR #605, opened 2018-07-02
Feature/multiple level dropdown component

Bluesmile82:
PR for the Multi-level dropdown component to select either a group or an element of this group

Try: Data Explorer/ NDC-content / Sectors dropdown
The double group label in the group detail view is due to the data. It should be fixed in the import/backend.

image


simaob:
Could the issue with the doubled data be because of the datasource that we are showing? Maybe users should filter first by Data Source? I think Mengpin replied to a question about the filters on Basecamp, so maybe let's review tomorrow morning with Agnieszka as well.

agnessa:
Regarding the duplicates in sectors, can you please try `bundle exec rake indc:import` - I think this is a case of old data (updated recently)

agnessa:
In general regarding the NDC content explorer, have a look at Mengpin's comments here: https://basecamp.com/1756858/projects/13795275/messages/75502613#comment_631627605 - she says we don't need the data source filter. It also seems to me that category needs to be a 2-level dropdown like sectors, and indicators need to be reloaded based on category. That is probably not part of this PR though.

Bluesmile82:
Fixed the double parent in the group view. It was an internal bug in the multi-dropdown at the end

Bluesmile82:
The removing of the source in NDC and the filtering upon selection will be dealt with in other PRs, but thanks!

PR #609, opened 2018-07-04
Feature/data explorer link params

Bluesmile82:
- Update the link to the modules with parsed params from the filters (Right now just source and goals)
- Download button that links to Data explorer page with the params of the module (they come with the external- prefix and are translated immediately)

image
image


simaob:
Something seems amiss in my local.

I go to http://localhost:3000/ghg-emissions and at first the param for the data explorer is not set. Once I change source to PIK it fills that in, and I click the button and takes me to the Data Explorer page but then it crashes. I get the following error on the console:

screen shot 2018-07-05 at 14 28 50

Bluesmile82:
The GHG emissions is not adding the params to the URL on the default page. That is why you have this two errors. I'll fix that

Bluesmile82:
Ready again!

simaob:
It still breaks for me, if the first page I visit is the GHG Emissions and then I click on the download button.

If I do it the other way around, going first to the data explorer -> ghg emissions -> download button, it works. So there seems to be some issues with the loading/fetching order.

simaob:
Other than that the linking looks good. =)

Bluesmile82:
Thanks for the catch. Ready again! Solved some problems with the parsing too:
- CAIT AR4 is not present in the options but it is in GHG emissions so is parsed to AR2
- All the snake_case params are converted to kebab-case

PR #613, opened 2018-07-05
Fix/pathways not showing data

weberjavi:
[Basecamp thread](https://basecamp.com/1756858/projects/13795275/todos/357890262)

Fixes Pathways chart not displaying data. The bug was introduced when generating chart domain through [recharts API](http://recharts.org/#/en-US/api/XAxis#domain) (domain function was not taking x values and was returning `Infinity`).
http://localhost:3000/pathways/models

It also adds `Y axis` buffer to country compare:
http://localhost:3000/countries/compare?locations=ZAF%2CGBR


Bluesmile82:
Thanks for fixing this! There are some cases where the buffer is a bit too much, like the following. But it can maybe be handled in some other PR.
image

PR #614, opened 2018-07-05
Add Pathways section, filters and download link

Bluesmile82:
WIP - Until the links PR is merged and I can finish the link to pathways part. Also methodology is missing

This PR adds the pathways section to the data explorer:
- [x]  Add section on the router
- [x]  Fetch data, metadata from the production URL
- [x]  Add filters
- [x]  Fix download link
- [x] Add methodology from the different model, scenario, and indicator
- [x]  Fix double fetching
- [x]  Fix link to the module with params

image


simaob:
The links PR is now merged.

simaob:
@Bluesmile82 is this ready for review? can we remove the WIP?

Bluesmile82:
Ready!

Bluesmile82:
Should work now! Region locations are showing an id on the fetch because they don't have a iso_code

simaob:
So the backend needs to adapt to this? @agnessa can the backend handle the id if that is passed? or should we create "fake" iso's?

simaob:
The link back from the Data Explorer to Pathways doesn't seem to be working when selecting a location that has an ISO code. As it doesn't get translated to an id.

Bluesmile82:
Ready again! Thanks for the catch. Too many edge cases 😅

weberjavi:
@Bluesmile82 just a doubt, would make sense to store the bulk data (without filters) and use it when filters are cleared instead of fetching again all the data every time the user clears a selection?

Bluesmile82:
I agree that this is maybe depending too much on the backend but after implementing pagination the load time will be much faster

simaob:
There's another PR with pre-filtering across selectors, so maybe that can be done afterwards!

PR #562, opened 2018-06-11
Feature/Add a download popup

qoobaa:

PR #616, opened 2018-07-09
Feature/download modal update

weberjavi:
This PR builds over #562 
It adds a survey each time a user wants to download some dataset (while the data explorer is not active on production this survey will get triggered when user tries to download a data set from the navigation download button).
TODO:
- [x]   Add user data to spreadsheet (follow [this gist](https://gist.github.com/qoobaa/5938bafd207e29fde46b80dc2398b9dc
) created by @qoobaa )

kapture 2018-07-09 at 9 56 21


weberjavi:
@Bluesmile82 , @simaob ready to review
I've shared a google spreadsheet where data gets stored. This spreadsheet is created with testing purposes, is to be decided where this data is going to live.

Bluesmile82:
This dropdown seems to be a bit cut at the first character. And also the scrollbar closes as in the other one. I'm beginning to think that is because they are inside a modal

image

Bluesmile82:
I think you shouldn't be able to subscribe if you don't enter an email

app/javascript/app/components/modal-download/modal-download-reducers.js
```diff
@@ -0,0 +1,37 @@
...
+const setModalDownloadParams = (state, { payload }) => ({
+  ...state,
+  isOpen: payload.open,
+  downloadUrl: payload.downloadUrl,
+  downloadSize: payload.size
+});
+
+const setRequiredFieldsError = (state, { payload }) => {
+  let isFieldRequiredEmpty = 0;
+  ['firstName', 'lastName', 'organization'].forEach(f => {
```
Bluesmile82: You could also use a some() loop

app/javascript/app/data/constants.js
```diff
@@ -341,6 +341,9 @@ export const SOURCE_IPCC_VERSIONS = [
   { name: 'UNFCCC - AR4', source_slug: 'UNFCCC', version_slug: 'AR4' }
 ];
 
+export const USER_SURVEY_SPREADSHEET_URL =
```
Bluesmile82: Let's move this to the .env

app/javascript/app/components/check-input/check-input-component.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { themr } from 'react-css-themr';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
+import styles from './check-input-styles';
+
+const CheckInputComponent = props => {
```
Bluesmile82: Cool Toggle component 😎

simaob:
@Bluesmile82 and/or @weberjavi it feels to me that the theory of the closing dropdown on the modal, might be worth investigating.

weberjavi:
Yes @simao, i've taken a look into it and seems it is because an automatically setted `tabindex="-1"`, i'll try to fix it on this PR

weberjavi:
@Bluesmile82 addressed your suggestions. 
Remember to add `USER_SURVEY_SPREADSHEET_URL=https://script.google.com/macros/s/AKfycbzgN1G9IdLYO3KqlTC4gzBxR1UTX5bYXu1qRaiRn1oD9qoaq6s/exec` to env constants.

weberjavi:
I haven't been able to reproduce the padding issue in the countries dropdown.

simaob:
Can we close the modal after the user clicks the button?

simao:
@weberjavi Wrong @simao please use @simaob ;)

qoobaa:
429 Too Many Simãos

weberjavi:
😆 sorry.
Two Tiagos and two Simaos, that's tricky (portuguese doppelganger 👯‍♂️ )

weberjavi:
@simaob, @Bluesmile82 this PR is ready to review. I think is better to handle dropdown (on modal) bug in another PR, what do you think?

weberjavi:
Yes @simaob, Kuba has created a [gist](https://gist.github.com/weberjavi/26082683cde640f25ca47e1f95803637) with the steps. I'll create the task on pivotal and ask Ola for them to create a spreadsheet and share the macro with us. Ok?

simaob:
Sounds good. But for the moment we can use our own on staging.

simaob:
Btw, I just remembered that there was a requirement that I'm not sure I wrote on the story:

- a user should only be asked for his details once every 5 days. So if a user downloads one file, it gets asked and if he fills it in, then he can download other files for the next 5 days without being prompted (as long as he is using the same browser of course).

weberjavi:
@simaob updated with localStorage data with 5 days expiration 👆

weberjavi:
@simaob please check the updates on jenkins and docker files 🙏

PR #618, opened 2018-07-09
Feature/data explorer prefilter dropdowns

Bluesmile82:
This PR checks the selected filters to accordingly filter the options of some of the dropdowns:

- [x] Historical emissions - Source filters sectors
- [x] NDC-content: categories filter indicators
- [x] NDC-SDG: goals filter Targets
- [x] Pathways: Model filter scenarios
- [x] Pathways: Categories and Scenarios filter indicators

In Pathways, the categories are not filtered by Model but I can't find a clear relationship in the metadata

Extra:
- Add NDC-content Categories and Pathways categories multiple selector.

Extra Extra: 
- Some fields now can be dependent on other i.e. indicators is dependent on categories on Pathways section so it will be disabled until Category is filled. And if the category is cleared the indicator will be too.

- Now there is an option for the multi-dropdown to avoid being able to select the group parents. I.e. you can't select a category in the Pathways section, only subcategories.

image


Bluesmile82:
WIP until Pathways filters are also added

Bluesmile82:
Ready again. Crazy filtering. Maybe some refactor is needed

simaob:
@weberjavi can you try to have a look at this today?

weberjavi:
On pathways there are some selections that should have data but is not being displayed on the table

image . 

image


This is the request `https://data.emissionspathways.org/api/v1/data/emission_pathways?category_ids[]=5&indicator_ids[]=23&locations=US&model_ids[]=3&scenario_ids[]=182` and the response is empty.

Bluesmile82:
That is an interesting error. I'm not sure why this is happening, even here:
https://data.emissionspathways.org/api/v1/data/emission_pathways?scenario_ids[]=182&indicators_id[]=23
@agnessa, any idea?

agnessa:
About the "no data" issue: is in the first instance where you're filtering by category only you're seeing results for indicator called Electricity|Generation by technology|Biomass w/o CCS. This indicator has id=44. When you select Biomass w/o CCS in the indicators dropdown somehow it filters by a different indicator with id=23 whose full name is Electricity|Capacity by technology|Biomass w/o CCS and that one does not have results. I think the problem is we need a subcategory filter before the indicator filter.

Bluesmile82:
Ready! Fixed the warning. The other error is most probably backend related. I will add a new task in PT to discuss it

Bluesmile82:
WIP - The indicator selector depends on the subcategory selected so it should be disabled until we have one. Also, I will probably have to disable the category selection

Bluesmile82:
Ready again: Check Extra extra in the description!

weberjavi:
The loader is now missing on first data fetch and `no data` message is displayed instead.

kapture 2018-07-12 at 10 23 10

weberjavi:
Another thing that is happening (do not know if it was happening before) is that when we visit for the second time a tab, even though we can see the data on the table another fetch to take the data and metadata is made and the selectors are disabled until the response arrives. 

kapture 2018-07-12 at 10 28 36

Bluesmile82:
Fixed the loading. We are always requesting the data again when we change sections. The filters are deleted when you change the section so you could have different data when you arrive there again. It should be faster with the pagination.

weberjavi:
Would it make sense to allow user to select only the countries that have available data in each section?   

For example in `Emission Pathways` only Argentina, Brazil, Canada, China, Colombia, Indonesia, India, Japan, South Korea, Mexico, Pakistan, Russia and Taiwan have data but the selector allows selecting all the countries and zones giving a lot of chances to end up with `no data` in the table.  

Maybe that is not the place for this comment and is more a design thing? 
@Bluesmile82 , @simaob what do you think?

weberjavi:
Thanks @Bluesmile82!!
Go pagination GO

PR #620, opened 2018-07-10
Fix dropdown scrollbar inside modal

Bluesmile82:
Remove tabindex from all the modals to be able to scroll with the scrollbars inside dropdowns.
Please check if you can find something better than this more than hacky solution

image

### DIFF
--- a/app/javascript/app/components/modal/modal-component.jsx
+++ b/app/javascript/app/components/modal/modal-component.jsx
@@ -5,6 +5,7 @@ import { themr } from 'react-css-themr';
 import Button from 'components/button';
 import Icon from 'components/icon';
 import closeIcon from 'assets/icons/sidebar-close.svg';
+import cx from 'classnames';
 import styles from './modal-styles.scss';
 
 class CustomModal extends PureComponent {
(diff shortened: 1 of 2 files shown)


simaob:
Something weird maybe... on the pages like Pathways or GHG Emissions I can not see the dropdown thingy when looking at select filters... same for Data Explorer. Not sure if it's related with this PR but I thought I'd check them to see if the behaviour was consistent.

Other than that dragging the dropdowns on the my Climate Watch modal seems fine now with these changes.

simaob:
By dropdown thingy I mean the: scrollbar, sorry. When scrolling on the selectors on the Pathways or GHG graphs I can't see the dropdown when I start scrolling (in Chrome) as I can see in My Climate Watch.

simaob:
OK, I think is independent from this PR, but here are some screenshots for information:

screen shot 2018-07-10 at 11 53 02 2
screen shot 2018-07-10 at 11 52 53 2

simaob:
See that the regions dropdown on pathways doesn't have a scrollbar, but the one on My Climate Watch does.

PR #625, opened 2018-07-11
Feature/mycw add model logos

weberjavi:
This PR adds model logo to MyCW visualizations (both on the visCreator and on the embedded modules). 
To see it in action it is mandatory to create a new visualization, because old ones have been already stored in the data base without the logo data.  
[Pivotal](https://www.pivotaltracker.com/story/show/158939295)
[Basecap](https://basecamp.com/1756858/projects/13795275/todos/357603432)  

kapture 2018-07-11 at 17 23 44


simaob:
The feature seems to be working, but I have a question. Why is it that when I edit an existing visualisation, even if I change the model, that the logo doesn't show up? Would that be possible? Otherwise we might be able to run a script to update existing charts on prod.

weberjavi:
@simaob if you only change the model but not the country, because the `state` has models data stored does not request models to the server hence not having the `logo` field (because it is consuming the old data), but if you change the country and therefore fetch models from the server the new data on the response has the `logo` data and it is displayed on the chart legend.
With the 'old' visualizations we are consuming data already stored that has an outdated structure.

PR #627, opened 2018-07-12
Add 2020 to timeline

weberjavi:
This PR adds a dot to country timeline that point to the 2020 commitment of countries to submit new NDC's
[Pivotal](https://www.pivotaltracker.com/story/show/158939770) 

Test it here: http://localhost:3000/countries/ARG

### DIFF
--- a/app/javascript/app/components/country/country-timeline/country-timeline-selectors.js
+++ b/app/javascript/app/components/country/country-timeline/country-timeline-selectors.js
@@ -21,6 +21,12 @@ export const getDates = createSelector(
         });
       }
     });
+    // add year 2020 as due date for new NDCS submissions
+    documents.push({
+      year: '2020',
+      link: null,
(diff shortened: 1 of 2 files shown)


weberjavi:
@Bluesmile82 , @simaob actually the dot link does not point to any page. I think it would be nice to point to the WRI document about [2020 NDC's enhancement](http://www.wri.org/sites/default/files/WRI17_NDC.pdf).  
What do you think?

simaob:
Hi Javi, I haven't checked it yet, but might be good to make it have a different colour to differentiate from the other dots. In regards to the link, maybe ask in the Basecamp thread to understand if that's something they'd be keen on.

weberjavi:
Waiting for a response in [basecamp](https://basecamp.com/1756858/projects/13795275/todos/345801190)

PR #628, opened 2018-07-12
Feature/data explorer pagination

Bluesmile82:
WIP - This PR is blocked and won't work in the Emission Pathways section as we need to expose the headers for https: 
https://www.pivotaltracker.com/story/show/158971822

This PR implements the pagination in the data explorer:
- Install react-paginate
- Use pagination
- Style the pagination + mobile
- Only fetch when the request has a page

The paginated requests are still slow: 
https://www.pivotaltracker.com/story/show/158971770

image

Extra:

- Fix the Pathways location filter (now its using location_ids[])


simaob:
ESP headers change deployed to production.

simaob:
Just noticed that in the case of Pathways the API information needs to include the domain name for pathways, as the API is not through Climate Watch. Should include: https://data.emissionspathways.org/

simaob:
That last comment is not really related with this PR, but as the API one had been merged, maybe it makes sense to handle in this PR. Other than that comment it looks like it's working \o/

Bluesmile82:
Ok, Ready again! Pagination is still slow but it will improve 😄

app/javascript/app/components/data-explorer-content/data-explorer-content-styles.scss
```diff
@@ -54,3 +66,39 @@
...
+
+  li {
+    padding: 0 5px;
+    display: inline-block;
+  }
+}
+
+.active {
+  a {
+    color: $theme-secondary;
```
weberjavi: I'm not being able to find the designs for the pagination, but would make sense to change the `$theme-secondary` colour for `$theme-color`?
Do not take this comment into account if you do not want, is just that I found the selected page number in the secondary colour too loud.

app/javascript/app/providers/data-explorer-provider/data-explorer-provider-actions.js
```diff
@@ -46,15 +44,20 @@ export const fetchDataExplorer = createThunkAction(
           section
         ]}${parsedQuery ? `?${parsedQuery}` : ''}`
       )
+        .then(response =>
```
weberjavi: Not sure if this is happening because of this changes on the data fetching, but when I filter with the selectors and then change tab (or page) when coming back to the filtered tab, the filter is already actyve but the selectors are empty (and the filter is not set on the URL)

_in this example I've selected `Cambridge Architectural Research` visited another tab and went back to `Emission Pathways`_

image

app/javascript/app/components/data-explorer-content/data-explorer-content-styles.scss
weberjavi: Pagination may need some more space to avoid layout breaking in some cases (or allow less selectable pages)

image

Bluesmile82:
Ready again! Fixed some problems with the fetching and page param flow. Also changed the style in the pagination

simaob:
No data seems to be requested when I change page, and the table on Historical Emissions doesn't scroll :/

Bluesmile82:
No table is supposed to scroll, we always have 7 elements per page. That is what we have pagination for. It's true that in Emission Pathways the rows are taller and you can scroll. I can take a look at that.

About the no data. Can you describe what you do when you don't get data? Its working fine for me.

simaob:
Is the 7 elements by design? It seems a little low to give an idea of the type of results. Maybe a more natural number would be 10 or 20 per page (with scroll in this case).

regarding the pagination, I click on any page and the contents of the table doesn't change at all, see some screenshots:

screen shot 2018-07-18 at 15 06 34
screen shot 2018-07-18 at 15 06 20

Bluesmile82:
Ready!

simaob:
Javi this one is waiting your re-review.

weberjavi:
Locations selector on `Pathways` is not working.

kapture 2018-07-19 at 10 36 54

weberjavi:
Beyond that every thing is working fine.

agnessa:
This is likely the cause: https://github.com/Vizzuality/emissions-scenario-portal/pull/270

simaob:
I think this is all good now =) Thanks!

PR #631, opened 2018-07-13
Fixed logo width

weberjavi:
This PR fixes logos `width` on MyCW visualisations (it can be tested on the visualisation edit modal or on the embedded page).
[Pivotal](https://www.pivotaltracker.com/story/show/159006498)

Due to the huge differences on logos formats some of them seem bigger than others. Maybe we could suggest some min and max measures?
@faustoperez what do you think?

image
image
image

### DIFF
--- a/app/javascript/app/components/my-climate-watch/viz-creator/components/charts/legend/legend-styles.scss
+++ b/app/javascript/app/components/my-climate-watch/viz-creator/components/charts/legend/legend-styles.scss
@@ -19,14 +19,19 @@
   box-shadow: none;
 }
 
-.legendLogo {
+.legendLogoContainer {
   display: flex;
   flex-direction: column;
+  align-items: center;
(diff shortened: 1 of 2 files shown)


app/javascript/app/components/my-climate-watch/viz-creator/components/charts/legend/legend-styles.scss
```diff
@@ -19,14 +19,19 @@
...
+  min-width: 110px;
   font-size: $font-size-sm;
 }
 
 .legendLogoTitle {
   padding-bottom: 10px;
 }
+
+.legendLogo {
+  width: 100px;
```
Bluesmile82: We could use 115px to be perfectly aligned to the 'Data provided by:' text. Just being picky 😝but its ok. It's true that some of them seem bigger and are easier to read

PR #632, opened 2018-07-13
Logout: delete token cookie, data and redirect

Bluesmile82:
This PR implements the logout from My CW:
- Delete user_token cookie
- Delete user data
- Redirect to my-climate-watch

image

### DIFF
--- a/app/controllers/api/v1/auth_controller.rb
+++ b/app/controllers/api/v1/auth_controller.rb
@@ -13,7 +13,11 @@ module Api
       # rubocop:enable LineLength
 
       def logout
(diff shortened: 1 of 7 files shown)


weberjavi:
@simaob it is working fine for me. I logout and have to login again to access my visualisations (reloading does not log me in)

Bluesmile82:
Ready again! Now the tokens are deleted in the rails side. I changed a bit the auth part in Rails but I think it is only related to this login/logout

simaob:
Thanks, Álvaro =)

PR #633, opened 2018-07-16
MyCW viscreator chart bugs

weberjavi:
-  Fix `Data provided by:` styles to avoid layout breaking:
[Basecamp](https://basecamp.com/1756858/projects/13795275/todos/357603432)

-  Update `setBuffer` function on `utils/graphs.js` to fix this two bugs:  

_Proper tick values_
[Basecamp](https://basecamp.com/1756858/projects/13795275/todos/358536529)
image

_Display data on charts with `0` min values_ 
[Basecamp](https://basecamp.com/1756858/projects/13795275/todos/358428226)
image
image

- Add `cancel` button to viscreator modal. User would be able to delete visualisations only from the viscard.  
-  Fix styles on viscreator modal on internet explorer


weberjavi:
@simaob you are right it is hidden by design, but it makes sense in case it is a delete button (the visualisation should be created before being able to delete it) but thats not the case with the cancel. I just added it.

app/javascript/app/components/my-climate-watch/my-visualisations/my-visualisations-styles.scss
```diff
@@ -39,9 +39,18 @@
   margin: $navbar-height auto;
 }
 
+.microsoftBrowserModal {
```
Bluesmile82: Thanks for fixing this 👍

app/javascript/app/components/my-climate-watch/viz-creator/components/steps/step-four.jsx
```diff
@@ -85,10 +84,10 @@ class Step4 extends Component {
         <div className={styles.saveContainer}>
           {id && (
```
Bluesmile82: This is why we don't have it in the Create modal, isn't it? It would be nice to have always cancel and save

PR #636, opened 2018-07-18
Fix eslint errors and warnings

aabdaab:
- split too long lines (more than 140 chars)
- add jest recognition for linter
- fix other issues threw by linter

### DIFF
--- a/app/javascript/app/components/about/about-partners/about-partners.js
+++ b/app/javascript/app/components/about/about-partners/about-partners.js
@@ -41,7 +41,8 @@ const partnershipWith = {
         src: climateActionTrackerImage
       },
       description:
-        'joined the Climate Watch development team and provided support in developing concepts for the scope and functions of Climate Watch.'
+        'joined the Climate Watch development team and provided support in developing concepts for the ' +
+        'scope and functions of Climate Watch.'
     },
     {
       link: 'http://climateanalytics.org/',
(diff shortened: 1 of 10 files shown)


app/javascript/app/components/about/about-partners/about-partners.js
```diff
@@ -41,7 +41,8 @@ const partnershipWith = {
         src: climateActionTrackerImage
       },
       description:
-        'joined the Climate Watch development team and provided support in developing concepts for the scope and functions of Climate Watch.'
+        'joined the Climate Watch development team and provided support in developing concepts for the ' +
```
Bluesmile82: We prefer to use always template literals than concatenation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

app/javascript/app/data/SEO.js
```diff
@@ -1,16 +1,23 @@
 export const TITLE = 'Climate Watch: Data for Climate Action';
 export const HOME_PAGE =
-  'Climate Watch is an open online platform designed to empower users with the climate data, visualizations and resources they need to gather insights on national and global progress on climate change, sustainable development, and help advance the goals of the Paris Agreement.';
+  'Climate Watch is an open online platform designed to empower users with the climate data, ' +
```
Bluesmile82: Same, template literals make it easier to update the text

app/javascript/app/components/about/about-partners/about-partners.js
aabdaab: I don't know to what extent should we listen to eslint ;> when you use template literals eslint throws "Strings must use singlequote". I used concatenation as the line was too long. Should I revert those changes?

app/javascript/app/data/SEO.js
aabdaab: same as above, eslint warnings

app/javascript/app/components/about/about-partners/about-partners.js
Bluesmile82: I will revert the changes in these cases. It's better to use string literals in my opinion. You could also disable the es-lint must use singlequote errors. But I'm not sure if it worths it.

PR #640, opened 2018-07-19
Updated node engine version on package.json

weberjavi:
This PR updates `node` engine version on `package.json` config file. 
The versions accepted are all from `v8.0.0` until `v9` (last one not included). Version `8` is the current Long Term Support (LTS) version of `node`.

Feel free to use this PR thread to discuss further about this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "climate-watch",
   "version": "1.4.0",
   "engines": {
-    "node": "8.0.0"
+    "node": ">=8.0.0 <9"
   },
   "scripts": {
     "start": "run-s rails:db:migrate servers",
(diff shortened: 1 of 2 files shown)


simaob:
I think this is sensible as well. I was using 8.9.2 locally, and without this change I had to instal another version of node.

simaob:
Maybe when we upgrade to node 9 (no idea what that would imply) we could be a bit more strict. In the rails side we use a specific version of ruby, but as most people use RVM or RBENV and these respect the `.ruby-version` file, this is easier to manage. with node I do use `nvm`, if everyone uses it then it would be simpler to use a more specific version.

PR #643, opened 2018-07-19
Feature/data download survey feature flag

weberjavi:
Creates a feature flag `FEATURE_DATA_SURVEY` to decide wether to display the User survey previous to data download (both on data downloader and on the download button on top navigation)

It also updates `Dockerfile` and `Jenkinsfile` to avoid showing the survey modal on production.


app/javascript/app/components/download-menu/download-menu.js
```diff
@@ -7,12 +7,56 @@ import { actions } from 'components/modal-download';
...
 import Component from './download-menu-component';
 
+const FEATURE_DATA_SURVEY = process.env.FEATURE_DATA_SURVEY === 'true';
 const { S3_BUCKET_NAME } = process.env;
 
 const server = `http://${S3_BUCKET_NAME}.s3.amazonaws.com`;
 const folder = '/climate-watch-download-zip';
 const url = `${server}${folder}`;
 
+const downloadMenuOptions = [
```
qoobaa: isn't it possible to make a more "generic" array and reuse it in both places?

weberjavi: @qoobaa you're right. It's been a quick fix to allow to update production `branch` with other features. Hopefully non survey stuff will be deleted soon. 
But if you want me I can handle it a cleaner way. It will take some more time as we are passing the options array to the `simpleMenu` component that, despite its name, is not so simple in the inside :>)

simaob: Given that the Data Explorer is nearly finished, I wouldn't bother with this. As you say we will get rid of this soon.

PR #646, opened 2018-07-20
Fix/ndcs map table categories

weberjavi:
This PR fixes the inconsistent data on [NDCs](http://localhost:3000/ndcs-content) map and table categories selector.
[Basecamp](https://basecamp.com/1756858/projects/13795275/todos/359316501)
[Pivotal](https://www.pivotaltracker.com/story/show/159172907)

kapture 2018-07-20 at 13 47 02

### DIFF
--- a/app/javascript/app/pages/ndcs/ndcs-reducers.js
+++ b/app/javascript/app/pages/ndcs/ndcs-reducers.js
@@ -18,6 +18,7 @@ export default {
           ...state,
           data: {
             categories: payload.categories,
+            mapCategories: state.data.mapCategories || {},
             sectors: payload.sectors,
             mapIndicators: state.data.mapIndicators || [],
             indicators: payload.indicators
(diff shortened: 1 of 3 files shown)


app/javascript/app/pages/ndcs/ndcs-reducers.js
```diff
@@ -18,6 +18,7 @@ export default {
           ...state,
           data: {
             categories: payload.categories,
+            mapCategories: state.data.mapCategories || {},
```
j8seangel: Is needed to generate a new array / object on the reducer? 

I mean, this will trigger a new render as it is a new instance so why not extend the data using something like?
```
data: {
  ...state.data,
  theNewParamsThatWeHave...
}
```

and update only the params that we need

aabdaab: +1

PR #647, opened 2018-07-23
Fix/countries selection clear

j8seangel:
This PR hides the clear selection when only one country is selected to avoid a full empty white page.

kapture 2018-07-23 at 13 34 01

React selectize doesn't have a good documentation so I have to dig into the src to found the [props availables](https://github.com/furqanZafar/react-selectize/blob/master/src/index.d.ts#L20) just in case someone else could need them again.


Bluesmile82:
Good job! I think that this is not related but I found that when selecting a country that is not in the store for the first time the socio-economic information doesn't show until we select another one.

image
image

Bluesmile82:
I have added a little commit to fix this bug

j8seangel:
Thanks for the bug fixing!

PR #650, opened 2018-07-24
Speed up years query using a materialised view

agnessa:
Some speed improvements here, mostly to do with the "years" query, which now runs on top of a normalised representation of the historical emissions records table.

`bundle exec rake db:migrate`


agnessa:
@santostiago can you have a look and see if these performance improvements help with the download timeout problem and whether you see further options for improving the speed of this query? Thanks!

PR #653, opened 2018-07-26
Ensure all default selection in ndc map and table

j8seangel:
Duplicating code for now, but would be nice to put this in practice once: https://github.com/reduxjs/reselect#sharing-selectors-with-props-across-multiple-component-instances

Not the best moment for now so here we have the same default indicator selected independently of the map or the table, also was an issue with the sort as was done after the indicator selector it could have diferences across them

kapture 2018-07-26 at 16 16 27


weberjavi:
There is a [already closed bug](https://www.pivotaltracker.com/story/show/158623094) that stated that we had to request for `filter=global` to populate NDCs table. We are not doing it anymore. Is this the intended behaviour?
@simaob

j8seangel:
Yes, the intended behaviour is to show the same indicators than the map. Thanks!

PR #655, opened 2018-07-31
Feature/data explorer change wording

weberjavi:
Changed `Emission Pathways` labels to `Pathways` on Data Explorer component.
Also reworded `Regions` dropdown label to `Countries and regions` 
[Pivotal (pathways)](https://www.pivotaltracker.com/story/show/159405777)
[Pivotal (Countries and regions)](https://www.pivotaltracker.com/story/show/159404553)
test here:
http://localhost:3000/data-explorer/historical-emissions?page=1

image
image

### DIFF
--- a/app/javascript/app/components/data-explorer-content/data-explorer-content-component.jsx
+++ b/app/javascript/app/components/data-explorer-content/data-explorer-content-component.jsx
@@ -13,7 +13,7 @@ import Loading from 'components/loading';
 import Button from 'components/button';
 import ModalDownload from 'components/modal-download';
 import anchorNavLightTheme from 'styles/themes/anchor-nav/anchor-nav-light.scss';
-import { toStartCase } from 'app/utils';
+import { toStartCase, deburrCapitalize } from 'app/utils';
 import cx from 'classnames';
 import ReactPaginate from 'react-paginate';
 import { DATA_EXPLORER_MULTIPLE_LEVEL_SECTIONS } from 'data/constants';
@@ -84,8 +84,8 @@ class DataExplorerContent extends PureComponent {
         (multipleSector(field) ? (
           <MultiDropdown
             key={field}
-            label={toStartCase(field)}
-            placeholder={`Filter by ${toStartCase(field)}`}
+            label={deburrCapitalize(field)}
+            placeholder={`Filter by ${deburrCapitalize(field)}`}
             options={filterOptions ? filterOptions[field] : []}
             value={selectedOptions ? selectedOptions[field] : null}
             disabled={isDisabled(field)}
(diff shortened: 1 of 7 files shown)


weberjavi:
Would be nice to get rid of `emission-pathways` in all the project (both on code and on files/folder structure)

app/javascript/app/components/data-explorer-content/data-explorer-content-component.jsx
```diff
@@ -84,8 +84,8 @@ class DataExplorerContent extends PureComponent {
         (multipleSector(field) ? (
           <MultiDropdown
             key={field}
-            label={toStartCase(field)}
-            placeholder={`Filter by ${toStartCase(field)}`}
+            label={deburrCapitalize(field)}
```
Bluesmile82: Please use the toStartCase util method. This is basically Converting some acronyms like INDC or CO2 to the correct format

app/javascript/app/data/constants.js
```diff
@@ -310,6 +310,13 @@ export const DATA_EXPLORER_SECTION_NAMES = {
   'ndc-content': 'ndc_content'
 };
 
+export const DATA_EXPLORER_SECTION_LABELS = {
```
Bluesmile82: Maybe we can rename DATA_EXPLORER_SECTION_NAMES to URIS or maybe even merge them as they have the same keys. Having Labels and Names together could be confusing

Bluesmile82:
Renaming the column has affected the flow of the dropdown data. Please take a look:

image

weberjavi:
Ok so I've misunderstood the requirement for this task ([change Regions to Countries and regions](https://www.pivotaltracker.com/story/show/159404553)), thought it was just a wording thing:

image

What they probably want is to display also countries on the dropdown options, as we are doing on GHG emissions module. Also to be consistent a rephrasing of the dropdown label of the module should be done.

image

I'll revert those changes and create another task to address them in other PR.
What do you think @Bluesmile82, @simaob ?

app/javascript/app/data/constants.js
weberjavi: If you agree I'll go with the change of the constant to `DATA_EXPLORER_SECTION_URIS`.

Bluesmile82:
Yes, I think you are right. We will have to mimic the dropdown. It's a bigger issue that should be addressed in other PR

app/javascript/app/components/data-explorer-content/data-explorer-content-component.jsx
weberjavi: Do we need to make those substitutions on this dropdown labels? 
Also what we need here is capitalize the string, using `toStartCase` returns `Countries And Regions` while `deburrCapitalize` returns `Countries and regions`

Even do acronym replacements inside a function called `toStartCase` may be hiding the real functionality of the function, shouldn't it be better to compose two functions and do something like `startCase(replaceAll(string, ACRONYM_REPLACEMENTS))`.

app/javascript/app/data/constants.js
weberjavi: ⚡️ done

PR #656, opened 2018-08-02
NDC-Comparison/Show country values in the correct order and dash if empty

Bluesmile82:
- Show empty values of the country in the accordion with a dash ('-')
- There was a bug showing the order of the values for each country. Before the countries without values were not displayed at all so some country values were misplaced. Now the array has empty values so they can be properly arranged.

Before: 
image

After:
image


simaob:
There are other tabs on the compare: mitigation, adaptation, etc. I'd apply the same principle to those as well.

PR #657, opened 2018-08-02
Feature/configurable sorting for data explorer endpoints

agnessa:
Enables `sort_col` and `sort_dir` params for historical emissions, NDC content, NDC-SDG linkages data explorer endpoints.

In case of historical emissions `sort_col` can be a year.

Related: https://github.com/Vizzuality/emissions-scenario-portal/pull/271

### DIFF
--- a/app/controllers/api/v1/data/historical_emissions_controller.rb
+++ b/app/controllers/api/v1/data/historical_emissions_controller.rb
@@ -13,7 +13,7 @@ module Api
                  each_serializer: Api::V1::Data::HistoricalEmissions::RecordSerializer,
                  params: params,
                  root: :data,
-                 meta: {years: @filter.years}
+                 meta: {years: @filter.years, columns: @filter.column_manifest}
         end
 
         def meta
--- a/app/controllers/api/v1/data/historical_emissions_controller.rb
+++ b/app/controllers/api/v1/data/historical_emissions_controller.rb
@@ -13,7 +13,7 @@ module Api
                  each_serializer: Api::V1::Data::HistoricalEmissions::RecordSerializer,
                  params: params,
                  root: :data,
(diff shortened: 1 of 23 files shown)


agnessa:
I haven't considered sorting by value per year. Let me see how hard that is going to be to add.

agnessa:
Actually it looks like sorting by year value is going to be pretty horrible to implement (for historical emissions and emission pathways). So just double checking this requirement. For example in pathways it probably doesn't even make sense to sort by year value, because we could have a different indicator on every row. Here it maybe makes more sense, since everything is converted to MtCO2e, but looks like a bit of work that is definitely not going to make this endpoint faster 😅 @Bluesmile82 @simaob Happy to hear your thoughts on that feature and whether or not it would be useful to merge this PR as well as the respective ESP PR already, since they cover sorting by all "normal" columns. The WIP label is on both of them because of this missing functionality, which could be covered in a separate PR if that helps progress work on the frontend.

Bluesmile82:
I agree that it doesn't make sense for pathways as they are different indicators. In Historical Emissions it does but maybe we can ask the client. I'm not sure the sorting by years was an explicit requirement or we just implied it. What do you think @simaob

agnessa:
Ready for review.

agnessa:
Hang on, one more thing to support https://www.pivotaltracker.com/story/show/159467339 and https://www.pivotaltracker.com/story/show/159467268

agnessa:
I piggybacked the small requests to change how data is returned in this PR in order to avoid further database changes, as these transformations are embedded in the materialised view which serves the search. Values for historical emissions are now rounded to 2 decimal places and if they are blank, they come out as 'N/A'. NB, reimporting historical emissions also refreshes the materialised views.

PR #658, opened 2018-08-02
Feature/data explorer split categories subcategories dropdown pathways

weberjavi:
This PR splits `categories` two level dropdown into two different dropdowns: `categories` and `subcategories`
[Pivotal](https://www.pivotaltracker.com/story/show/159405453)

BEFORE . 

image

AFTER .   
image

Further improvements should be done to allow the display of the proper selections when changing from data explorer to module view and vice versa. Those improvements will be addressed in another PR regarding dropdowns dependencies ([Pivotal](https://www.pivotaltracker.com/story/show/159405550))

It also fixes an small issue with navigation layout triggering horizontal scroll when `FEATURE_DATA_EXPLORER=true`

before .  
image

after .  

image

### DIFF
--- a/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
+++ b/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
@@ -19,6 +19,66 @@ import {
   DATA_EXPLORER_MULTIPLE_LEVEL_SECTIONS
 } from 'data/constants';
 
+const SECTION_NAMES = {
+  pathways: 'emission-pathways',
+  historicalEmissions: 'historical-emissions'
+};
+
+const FILTER_NAMES = {
+  categories: 'categories',
+  subcategories: 'subcategories'
+};
+
+const FILTERED_FIELDS = {
+  'historical-emissions': {
+    sectors: [
+      {
+        parent: 'source',
+        id: 'data_source_id'
+      }
+    ]
+  },
+  'ndc-sdg-linkages': {
+    targets: [
+      {
+        parent: 'goals',
+        parentId: 'id',
+        id: 'goal_id'
+      }
+    ]
+  },
+  'ndc-content': {
+    indicators: [
+      {
+        parent: FILTER_NAMES.categories,
+        parentId: 'id',
+        id: 'category_ids'
+      }
+    ]
+  },
+  'emission-pathways': {
+    scenarios: [
+      {
+        parent: 'models',
+        idObject: 'model',
+        id: 'id'
+      }
+    ],
+    indicators: [
+      {
+        parent: FILTER_NAMES.categories,
+        idObject: 'category',
+        id: 'id'
+      },
+      {
+        parent: 'scenarios',
+        parentId: 'indicator_ids',
+        id: 'id'
+      }
+    ]
+  }
+};
+
 const getMeta = state => state.meta || null;
 const getSection = state => state.section || null;
 const getSearch = state => state.search || null;
@@ -40,7 +100,7 @@ export const getSourceOptions = createSelector(
       !meta ||
       isEmpty(meta) ||
       !section ||
-      section !== 'historical-emissions' ||
+      section !== SECTION_NAMES.historicalEmissions ||
       !meta[section]
     ) {
       return null;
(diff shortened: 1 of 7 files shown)


app/javascript/app/data/constants.js
```diff
@@ -374,6 +375,9 @@ export const DATA_EXPLORER_TO_MODULES_PARAMS = {
     },
     categories: {
       key: 'category'
+    },
+    subcategories: {
+      key: 'subcategories'
```
Bluesmile82: This should be key: 'subcategory' to be parsed correctly

Bluesmile82:
This PR doesn't include:
- Data Explorer Pathways to module parsing
- Download params parsing
- Filtering of Indicator options by subcategory (Maybe we can reuse the dependencies constant)

app/javascript/app/data/constants.js
weberjavi: I'll change it to prepare the next iteration but (as for now) the module (pathways page) is not accepting `subcategory` as URL param.

weberjavi: ⚡️ done

app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
```diff
@@ -138,6 +154,25 @@ export const getLink = createSelector(
   }
 );
 
+function getOptions(section, filter, filtersMeta) {
+  if (section !== 'emission-pathways') return filtersMeta[filter];
```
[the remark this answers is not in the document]
aabdaab: like enums sort of way `EMISSION_PATHWAYS` etc

app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
```diff
@@ -138,6 +154,25 @@ export const getLink = createSelector(
...
+  if (filter === 'categories' || filter === 'subcategories') {
+    return filtersMeta.categories;
+  }
+  return filtersMeta[filter];
+}
+
+function parseOptions(section, filter, options) {
+  if (section !== 'emission-pathways') return options;
```
aabdaab: use `switch` here maybe?

app/javascript/app/components/navbar/navbar-component.jsx
```diff
@@ -1,15 +1,22 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
+import cx from 'classnames';
 import Nav from 'components/nav';
 import ToolsNav from 'components/tools-nav';
 
 import styles from './navbar-styles.scss';
 
+const FEATURE_DATA_EXPLORER = process.env.FEATURE_DATA_EXPLORER === 'true';
```
aabdaab: maybe it's a bit overkill but... wdyt of moving it to `feature-data-explorer` file (we could keep it in `features` folder) and implement a function `isEnabled` ? this way we could standardize features enabling and if the logic grows (this feature is enabled if this && this && this are met) we'd have that in one place.

app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
weberjavi: Well I think this is more related with preferences (as in this small statement a `switch` would not add any performance improvement). 
I use to go into `switch`es when there are more cases,  but if you think it could increase redeability I'll go for it

aabdaab: I meant for readibility reasons, yes. i think having too many ifs is a bad practice, is all

app/javascript/app/components/navbar/navbar-component.jsx
weberjavi: I think is a good idea 👍 , but should it belong to its own PR?

app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
weberjavi: ⚡️ done

aabdaab: 🙏 thank you!

app/javascript/app/components/navbar/navbar-component.jsx
aabdaab: if it's a feature that appears in this PR, I think it can be done in this PR, wdyt?

app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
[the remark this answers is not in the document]
weberjavi: ⚡️ done

app/javascript/app/components/navbar/navbar-component.jsx
weberjavi: It could be but would be nice to also do the same with `FEATURE_DATA_SURVEY` and maybe those changes get hidden in this PR.

aabdaab: ok then, it may be another PR for that 👍

PR #659, opened 2018-08-02
Feature/data explorer ndc indc linkable

Bluesmile82:
This PR links the INDC text from the Data Explorer/ NDC-SDG Linkages table to the correspondent text in the document.

image

### DIFF
--- a/app/javascript/app/components/data-explorer-content/data-explorer-content-component.jsx
+++ b/app/javascript/app/components/data-explorer-content/data-explorer-content-component.jsx
@@ -26,7 +26,7 @@ const FEATURE_DATA_SURVEY = process.env.FEATURE_DATA_SURVEY === 'true';
 class DataExplorerContent extends PureComponent {
   // eslint-disable-line react/prefer-stateless-function
   renderTable() {
-    const { data, firstColumnHeaders, loading } = this.props;
+    const { data, firstColumnHeaders, loading, titleLinks } = this.props;
     if (loading) return <Loading light className={styles.loader} />;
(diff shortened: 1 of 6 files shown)


simaob:
The document type and document language are now passed via the API. Is this what you needed?

weberjavi:
@Bluesmile82 is this ☝️ ready to review? it has some conflicts

Bluesmile82:
Conflicts solved!

weberjavi:
There are some cases in which highlighted text (on NDCs country full page) is not inside the view when joining the page, there are also some cases where there is other text highlighted.  

Test on: http://localhost:3000/data-explorer/ndc-sdg-linkages?ndc-sdg-linkages-goals=3&page=1
Testing on **Burkina Faso** INDC texts. 

kapture 2018-08-07 at 18 13 38

Bluesmile82:
@weberjavi Good catch! It seems that for some of the targets the backend return several linkages.  We need a idx attribute for the ScrollToHighlightIndex component. I'm not sure if we can find this in the backend @simaob @agnessa? Maybe @j8seangel can throw some light too

simaob:
I feel that this is the current behaviour that we have on the NDC-Full Text page anyway, so maybe doesn't need to be sorted in this branch just yet.

simaob:
So if we are not handling the issue of the multiple linkages only here, is the rest good to go?

Bluesmile82:
We can handle it in some other task. But I think we are missing some information from the backend. @weberjavi Do you approve without this change?

PR #661, opened 2018-08-07
Feature/data explorer pathways dropdowns dependency

weberjavi:
This PR displays dropdown options depending on previous selections on Pathways tab of Data Explorer.
[Pivotal](https://www.pivotaltracker.com/story/show/159405550)
Test it on: http://localhost:3000/data-explorer/emission-pathways?page=1

kapture 2018-08-07 at 12 53 23

### DIFF
--- a/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
+++ b/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
@@ -3,6 +3,7 @@ import remove from 'lodash/remove';
 import isEmpty from 'lodash/isEmpty';
 import isArray from 'lodash/isArray';
 import pick from 'lodash/pick';
+import uniqBy from 'lodash/uniqBy';
 import qs from 'query-string';
 import { parseQuery } from 'utils/data-explorer';
 import { findEqual } from 'utils/utils';
@@ -27,6 +28,7 @@ const SECTION_NAMES = {
 
 const FILTER_NAMES = {
   models: 'models',
+  scenarios: 'scenarios',
   categories: 'categories',
   subcategories: 'subcategories'
 };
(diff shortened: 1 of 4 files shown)


Bluesmile82:
There is a crash when we select a model in Pathways and then a location that doesn't have that model. Try: http://localhost:3000/pathways/models?category=10&indicator=388&model=3
Also, check the conflicts, I think is because of the migration.

weberjavi:
☝️ ready 
@j8seangel or @aabdaab please check that out.
The requirements for this task where kind of vague (just _display dropdown options depending on previous selections_), so I have to decide along the way. The intended functionality is:
 - You can freely filter by `location`, `model` and `category`
 - `scenarios`, `subcategories` and `indicators` are disabled until some of the previous dropdowns are selected (as `scenarios` are dependent of a `model` and `subcategories` and `indicators` are dependent of a `category`) 
-  Did not add `location` dependency in any selector because the user could want to filter by `model` or `category` without taking into account the `location`
-  Once a `model` or `category` (or any of their dependents) is chosen if the user selects a `location` probably the filter will end up with _no data_ and user will be forced to clear the filtering to be able to proceed with a successful search.

Please share your thoughts.
@simaob you are also welcome to opine.

🙏

app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
```diff
@@ -220,32 +224,111 @@ export const getLink = createSelector(
...
+function getPathwaysCategoryOptions(query, filtersMeta) {
+  if (!query || !query.scenarios) {
+    return filtersMeta.categories.filter(c => c.parent_id === null);
+  }
+  const selectedScenarioId = query.scenarios;
+  const scenarioIndicatorsIds = filtersMeta.scenarios.find(
+    sc => sc.id === selectedScenarioId
+  ).indicator_ids;
+  const categories = [];
+  scenarioIndicatorsIds.forEach(indId =>
```
aabdaab: I thi

weberjavi:
@aabdaab yeah, I agree about functions being quite big. 
I've been splitting them along the way (the ones I had to deal with), and hope is a job that we could improve on the remaining tasks of the data-explorer.
Maybe is even desirable to move to another folder the ones that do not affect every section of the explorer (as `getPathways...Options` ones)

app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
weberjavi: ⚡️ done 
Thanks, much cleaner that way (also spotted a bug 🐛 ⚔️ )

j8seangel:
👍

PR #664, opened 2018-08-07
Data Explorer - Enable sort via backend

Bluesmile82:
This PR enables the sort via Backend. Before the sorting of the table was done only in the page that we were in. Now we send a request to the backend to sort the table

- Add and fetch by sort params in the url
- Dont sort years in Pathways
- Documentation in the frontend updated

Note: The Pathways years are not sorted in the backend

image


simaob:
@weberjavi can you add this as a story on Pivotal Tracker, as it's not directly related to this frontend task? Then we can merge if everything else is good.

weberjavi:
Fair enough @simaob 👍 
Lets merge then.

PR #665, opened 2018-08-08
Fixed: in the emissions download, the field 'emissions' was sent.

santostiago:
The null values are replaced by `N/A`.

There was as bug with the construction of the csv file. The field `emissions` was being sent. Do you think the way used to fix it is the most appropriate one?

### DIFF
--- a/app/services/api/v1/data/historical_emissions_csv_content.rb
+++ b/app/services/api/v1/data/historical_emissions_csv_content.rb
@@ -6,8 +6,8 @@ module Api
       class HistoricalEmissionsCsvContent
         def initialize(filter, output)
           @grouped_query = filter.call
-          @headers = filter.column_aliases
-          @headers.shift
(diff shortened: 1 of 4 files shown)


simaob:
@agnessa going to merge as the functionality seems to be here. But please can you come back to it and check the solution on removing the emissions key. Cheers!

app/services/api/v1/data/historical_emissions_csv_content.rb
```diff
@@ -6,8 +6,8 @@ module Data
       class HistoricalEmissionsCsvContent
         def initialize(filter, output)
           @grouped_query = filter.call
-          @headers = filter.column_aliases
-          @headers.shift
+          # FIXME: To remove id and emissions: Should this be here?
```
agnessa: probably would be better to disable these columns for download in the configuration bit

PR #667, opened 2018-08-09
Fix/data explorer historical regions to countries

Bluesmile82:
This PR changes the historical emissions dropdown to provide regions and countries. Also changes the slugs in the URL to ids with all the subsequent changes.

- Change the drop-down to multi-select. Note: Not all the dropdowns could be automatically updated, there might be some changes to do still.
- Group the countries and regions
- Fix pagination
- Change slugs in the URL to ids and iso_codes
- Targets drop-down now display description too
- Some refactor

image
image
image


app/javascript/app/components/data-explorer-content/data-explorer-content.js
```diff
@@ -165,23 +209,17 @@ class DataExplorerContentContainer extends PureComponent {
       section,
       filterName
     ).map(k => ({ name: `${section}-${k}`, value: undefined }));
-
+    let parsedValue = value;
+    if (multiple) parsedValue = this.parsedMultipleValues(filterName, value);
+    parsedValue = parsedValue === '' ? undefined : parsedValue;
```
aabdaab: maybe use shortcut like `parsedValue = parsedValue || undefined` ?

Bluesmile82: Actually, it wasn't needed anymore. Thank you for spotting it

Bluesmile82:
Ready again

PR #670, opened 2018-08-16
Feature/data explorer add year filters

Bluesmile82:
This PR adds the year filters:

- Add the start_year and end year filters
- Changes the order of the displayed columns: https://www.pivotaltracker.com/story/show/159760423 (Still some changes and defaults to apply)
- Changes the per page to 200: https://www.pivotaltracker.com/story/show/159760381

image

Extra: 
  - Fix fetch problem when no filters
  - Extract filters part to component and refactor (Next step could be to extract the selectors to a different file)
- In the Pathways module get the subcategories to the URL and parse them to the data explorer through the download button
image

### DIFF
--- a/app/javascript/app/components/data-explorer-content/data-explorer-content.js
+++ b/app/javascript/app/components/data-explorer-content/data-explorer-content.js
@@ -10,21 +10,16 @@ import isEmpty from 'lodash/isEmpty';
 import pick from 'lodash/pick';
 import {
   DATA_EXPLORER_SECTIONS,
-  DATA_EXPLORER_FILTERS,
   DATA_EXPLORER_EXTERNAL_PREFIX,
-  DATA_EXPLORER_DEPENDENCIES,
   DATA_EXPLORER_PER_PAGE
 } from 'data/data-explorer-constants';
 import { ESP_HOST } from 'data/constants';
 import DataExplorerContentComponent from './data-explorer-content-component';
 import {
   parseData,
-  getActiveFilterRegion,
   getMethodology,
   getSectionLabel,
   getFirstTableHeaders,
-  getDependentOptions,
-  getSelectedOptions,
   parseFilterQuery,
   parseExternalParams,
   getLink,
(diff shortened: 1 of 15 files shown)


Bluesmile82:
Added an important fix for data initialization. The data was not fetched at first if no filters were selected

app/javascript/app/components/data-explorer-content/data-explorer-content.js
```diff
@@ -137,6 +103,7 @@ class DataExplorerContentContainer extends PureComponent {
   }
 
   componentDidUpdate(prevProps) {
+    if (!this.dataFetchInitialized) this.dataFetchInitialized = true;
```
j8seangel: 👃 👀

agnessa:
Maybe I can add the backend fix in this branch, so it can be applied within the same PR?

Bluesmile82:
Sure, feel free to do it

agnessa:
I have a (maybe) better idea on how to fix it:
- the `meta -> years` property would contain the list of all years available when all other filters are applied except start / end year - you could use that directly for the dropdowns
- to render the year columns, in case start/end year were provided, you'd need to remove years from outside of the range from `meta -> years`; I can return the currently selected start_year and end_year in the `meta` part to make it easier (if it helps)

The reason is I want to avoid running 2 queries for available years, as it is the longest query for this endpoint (same for queries which retrieve max / min year). Would that work? I think the same applies to historical emissions, right?

agnessa:
It now returns 2 arrays, `years` is the one "not filtered by year range" to be used for the dropdown, `header_years` is only those years to show up in column headers.

```
      "years":[
         1850,
         1851,
         ...
      ],
      "header_years":[
         1921,
         1922,
         ...
      ]
```

Bluesmile82:
Ready again!

Bluesmile82:
Thanks for the comments! Ready again!

app/javascript/app/components/data-explorer-content/data-explorer-filters/data-explorer-filters.js
```diff
@@ -117,14 +108,17 @@ class DataExplorerContentContainer extends PureComponent {
...
+    const dependentKeysToDeleteParams = DATA_EXPLORER_DEPENDENCIES[section]
+      ? getDependentKeysToDelete(section, filterName).map(key => ({
+        name: `${section}-${key}`,
+        value: undefined
+      }))
+      : [];
+
+    const parsedValue = multiple
+      ? this.parsedMultipleValues(filterName, value)
+      : value;
```
weberjavi: Not sure if there is any case in which not forcing empty string (`''`) to `undefined` will trigger any error, that's why I though a function would be needed. But if this is not the case feel free to merge it.
Thanks!! 🙏

PR #671, opened 2018-08-16
Feature/my cw login social

Bluesmile82:
As the email-password login was not working this PR changes the MY CW login page to have the three social login buttons. This is workingn but redirect through GFW login, we still need some accounts created by the client to change it to the climate accounts login.

- Add buttons with social login and style
- Little button refactor

image

### DIFF
--- a/app/javascript/app/components/button/button-component.jsx
+++ b/app/javascript/app/components/button/button-component.jsx
@@ -20,11 +20,7 @@ const Button = props => {
   } = props;
   const classNames = cx(className, styles.button, {
     [styles.square]: square,
-    [styles.transparent]: color === 'transparent',
-    [styles.yellow]: color === 'yellow',
-    [styles.white]: color === 'white',
-    [styles.plain]: color === 'plain',
-    [styles.red]: color === 'red',
+    [styles[color]]: !!color,
     [styles.noBox]: noBox,
     [styles.disabled]: disabled || (!onClick && !link && !href),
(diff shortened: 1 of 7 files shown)


app/javascript/app/components/button/button-component.jsx
```diff
@@ -20,11 +20,7 @@ const Button = props => {
   } = props;
   const classNames = cx(className, styles.button, {
     [styles.square]: square,
-    [styles.transparent]: color === 'transparent',
-    [styles.yellow]: color === 'yellow',
-    [styles.white]: color === 'white',
-    [styles.plain]: color === 'plain',
-    [styles.red]: color === 'red',
+    [styles[color]]: !!color,
```
j8seangel: 👏

app/javascript/app/data/constants.js
```diff
@@ -2,11 +2,13 @@ const { GFW_API } = process.env;
 const { CW_API } = process.env;
 const { ESP_API } = process.env;
 
-export const LOGIN_URL = `${GFW_API}/auth/login?applications=climate-watch&token=true&callbackUrl=${location.origin}${CW_API}/auth/login`;
+export const loginURL = social =>
```
j8seangel: Seems weird to have a function inside a file called constants, maybe we could keep the constant here and then add a getLoginUrlBySocial helper

Bluesmile82:
Can you check it again? I think the spinner can be added in some other PR

PR #675, opened 2018-08-20
Fix/data explorer regions dropdown

Bluesmile82:
The data explorer/ GHG emissions/ Regions and countries dropdown was not working as expected:
- Dots were not appearing in the selected options
- Some strings in the URL were showing the label instead of the value

This was fixed and:
- Now only when we have one region or country selected we show the name as the dropdown label

Important:
- The selection is now cleared when we select a new region to match the GHG dropdown behavior. 
- The link to GHG won't automatically transfer the regions and countries, this will be done in a new PR.

image

### DIFF
--- a/app/javascript/app/components/data-explorer-content/data-explorer-filters/data-explorer-filters.js
+++ b/app/javascript/app/components/data-explorer-content/data-explorer-filters/data-explorer-filters.js
@@ -81,8 +81,17 @@ class DataExplorerContentContainer extends PureComponent {
     ];
   };
 
-  parsedMultipleValues = (filterName, value) =>
-    value.map(filter => filter.value).toString();
+  parsedMultipleValues = (filterName, value) => {
+    const selectedValue = value[value.length - 1];
+    if (
+      selectedValue &&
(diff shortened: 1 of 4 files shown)


app/javascript/app/components/data-explorer-content/data-explorer-filters/data-explorer-filters.js
```diff
@@ -87,20 +87,11 @@ class DataExplorerContentContainer extends PureComponent {
     const removing = oldFilters && value.length < oldFilters.length;
     const selectedFilter = !oldFilters
       ? value[0]
-      : value
-        .filter(x => oldFilters.indexOf(x) === -1)
-        .concat(oldFilters.filter(x => value.indexOf(x) === -1))[0];
+      : value.filter(f => !oldFilters.map(o => o.label).includes(f.label));
```
weberjavi: This logic should be changed, now it is not matching any element inside of `value` object when removing a selection.

Bluesmile82:
Updated to match Pathways module!

PR #677, opened 2018-08-22
Fix the build

agnessa:


weberjavi:
Merging this fix as Travis is not complaining
Thanks!

PR #678, opened 2018-08-23
Fix/link to module data explorer

weberjavi:
This PR moves modules to Data Explorer links generation logic from `ButtonGroup` to `utils/data-explorer.js` file.
Now `generateLinkToDataExplorer` is used on each module to create a `link` prop that is passed from the module to `ButtonGroup` with the generated `URL`.

### DIFF
--- a/app/javascript/app/components/emission-pathways/emission-pathways-graph/emission-pathways-graph.js
+++ b/app/javascript/app/components/emission-pathways/emission-pathways-graph/emission-pathways-graph.js
@@ -98,27 +98,34 @@ class EmissionPathwayGraphContainer extends PureComponent {
     }
 
     const { search, filtersSelected } = this.props;
-    [
+    this.updateUrlWithNewParams(search, filtersSelected);
+  }
+
+  updateUrlWithNewParams = (search, filtersSelected) => {
+    const possibleParams = [
       'model',
       'category',
       'subcategory',
       'indicator',
       'currentLocation'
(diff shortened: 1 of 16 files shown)


app/javascript/app/components/emission-pathways/emission-pathways-graph/emission-pathways-graph.js
```diff
@@ -113,8 +105,17 @@ class EmissionPathwayGraphContainer extends PureComponent {
       'indicator',
       'currentLocation'
     ].forEach(f => {
-      if (!search[f] && filtersSelected[f]) {
-        this.updateUrlParam({ name: f, value: filtersSelected[f].value });
+      if (!search[f]) {
+        if (f === 'currentLocation' && filtersSelected.location) {
+          this.updateUrlParam({
```
Bluesmile82: This probably needs be refactored. Calling several times updateURL will only get the one on the URL. Its better if we aggregate them in an array and only call updateUrlParam once as in data-explorer-content.js:
```
data-explorer-content.js:
    const paramsToUpdate = Object.keys(validParams).map(key => ({
      name: key,
      value: validParams[key]
    }));
    this.updateUrlParam(paramsToUpdate, true);
```

weberjavi:
@Bluesmile82 ☝️ ready again

Bluesmile82:
Working great! Thanks for the changes

PR #682, opened 2018-08-24
Feature/data explorer order and defaults

Bluesmile82:
This PR implements the default filters selection and ordering of columns and filters:
image

- Orders columns following the document: https://www.pivotaltracker.com/story/show/159760423
- Orders filters
- Default filters in URL
- Refactor and logic to avoid double renders when we have external parameters or not all the default filters in the URL
- Be able to clear the filters from the default value so they show every option possible

Note: If the filters are wrong or not available the selected filter will be an empty string so the field remains unfiltered without errors or stack overflow


Bluesmile82:
Ready again! Solved conflicts, clear defaults feature included and year defaults too

Bluesmile82:
Thanks for the catch. Also fixed a problem in the year filters

weberjavi:
Thank you!
Everything working fine 👍

PR #683, opened 2018-08-24
NDC-SDG Linkages backend changes

agnessa:
- adds statuses, climate_responses and types_of_information to meta
- merges goal number + goal title into "sdg", target number + target title into "sdg_target
- removes document_type and language from response
- reorders columns in csv


app/serializers/api/v1/data/ndc_sdg/ndc_sdg_serializer.rb
```diff
@@ -4,19 +4,15 @@ module Data
...
           attribute(:status) { object['status'] }
+          attribute(:sector) { object['sector'] }
           attribute(:climate_response) { object['climate_response'] }
           attribute(:type_of_information) { object['type_of_information'] }
-          attribute(:sector) { object['sector'] }
-          attribute(:target_number) { object['target_number'] }
-          attribute(:target) { object['target'] }
-          attribute(:goal_number) { object['goal_number'] }
-          attribute(:goal) { object['goal'] }
-          attribute(:document_type) { object["document_type"] }
```
Bluesmile82: We need the document_type and language in the table to be able to redirect to the correct indc_text document. Maybe include it only in the table data JSON and not the CSV and it will be filtered so it won't appear in the frontend table

Bluesmile82:
The new filters (status, climate_response, and types_of_information) in NDC-SDG Linkages are not working in the backend. Do they need to be added or there is already something there but not documented? 
Try: http://localhost:3000/data-explorer/ndc-sdg-linkages?ndc-sdg-linkages-status=Present&page=1

aabdaab:
@Bluesmile82 Please check if there's anything else missing in this PR

Bluesmile82:
Thanks for the changes! It works fine

PR #686, opened 2018-08-27
Add Top Emitters region to regions dropdown

Bluesmile82:
This PR adds the Top Emitters option to the regions in Historical Emissions

Note: As this is a non-backend region it is needed to add all the countries along with the TOP region

image

### DIFF
--- a/app/javascript/app/data/data-explorer-constants.js
+++ b/app/javascript/app/data/data-explorer-constants.js
@@ -1,3 +1,5 @@
+import { TOP_EMITTERS } from './constants';
+
 export const DATA_EXPLORER_BLACKLIST = [
   'id',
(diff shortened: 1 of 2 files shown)


app/javascript/app/data/data-explorer-constants.js
```diff
@@ -1,3 +1,5 @@
+import { TOP_EMITTERS } from './constants';
```
weberjavi: Would it be better to import the `constants` file with `data/constants` instead that with the relative import? 
(just to avoid crashes in case the `data-explorer-constants` or `constants` files get moved to another folder).

Bluesmile82:
Thank you for the comment. Fixed and conflicts too. Merging...

PR #687, opened 2018-08-28
Feature/narrower table columns

weberjavi:
[Pivotal task](https://www.pivotaltracker.com/story/show/159960983)
This PR changes Data explorer table columns width on [Pathways section](http://localhost:3000/data-explorer/emission-pathways?page=1). 
With the new width some `years` columns could be seen without scrolling to the right giving the user a hint that more years are hidden in this side of the table.

image


app/javascript/app/components/table/table.js
```diff
@@ -50,6 +57,9 @@ class TableContainer extends PureComponent {
   };
 
   setColumnWidth = column => {
+    if (this.props.hasStrechedColumns) {
```
Bluesmile82: What do you think about a forcedColumnsWidth prop with a width like 100 instead of the hasStretchedColumns boolean. It will be more flexible

Bluesmile82:
Maybe we could also change the width of the historical emission table. Just to keep them similar

weberjavi:
Haven't forced column width on GHG because there's only one year (2014) with the default selections and would look weird.
To define a forced width in any other section just add the section and the desired width to `DATA_EXPLORER_TABLE_COLUMNS_WIDTH` on `data-explorer-constants.js`.

PR #688, opened 2018-08-28
Feature/data explorer multi scenario

Bluesmile82:
This PR uses multiselect for Pathways / scenario selector so its possible to select several scenarios and match the Pathways module.

image

Extra: Fix bug in dependent selection with cleared dropdowns. The forced dependent selections were not being activated if the URL had empty string values.

### DIFF
--- a/app/javascript/app/components/emission-pathways/emission-pathways-graph/emission-pathways-graph-selectors.js
+++ b/app/javascript/app/components/emission-pathways/emission-pathways-graph/emission-pathways-graph-selectors.js
@@ -4,6 +4,7 @@ import isUndefined from 'lodash/isUndefined';
 import uniqBy from 'lodash/uniqBy';
 import uniq from 'lodash/uniq';
 import groupBy from 'lodash/groupBy';
+import isArray from 'lodash/isArray';
 import sortBy from 'lodash/sortBy';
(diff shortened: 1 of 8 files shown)


app/javascript/app/components/emission-pathways/emission-pathways-graph/emission-pathways-graph-selectors.js
```diff
@@ -160,18 +161,19 @@ export const getScenariosOptions = createSelector(
...
+    if (!scenariosSelected) {
       return scenarios;
     }
-    const scenarioSelectedParsed = scenarioSelected
-      .split(',')
-      .map(s => parseInt(s, 10));
+    let scenarioSelectedParsed = isArray(scenariosSelected)
+      ? scenariosSelected
+      : scenariosSelected.split(',');
+    scenarioSelectedParsed = scenarioSelectedParsed.map(s => parseInt(s, 10));
```
weberjavi: Seems weird using a `map` to mutate the mapped object. Wouldn't it make sense to name the first `let scenarioSelectedParsed` differently (e.g. `const selectedScenariosArray`) (as what you're doing is checking if it's an array and converting it to one if it is not) an then `map` over it assigning the new array to `const scenarioSelectedParsed`?

PR #693, opened 2018-09-06
Fix importing historical emissions sectors

agnessa:
When importing sectors for historical emissions also consider the data source.

Requires re-running: `bundle exec rake historical_emissions:import`


simaob:
@weberjavi ping.

PR #694, opened 2018-09-06
Removed fake 'Economy-wide' sector from API response

agnessa:
The sector used to be returned as 'Economy-wide' where the sector was empty, but that was confusing because filtering did not work correctly when applying the real 'Economy-wide' sector filter. Now it's back to blank when sector is not provided; it should be part of the data if it's needed.


simaob:
@weberjavi up for review it seems.

weberjavi:
There are some conflicts to be resolved thought :>/

PR #696, opened 2018-09-10
Feature/ghg emissions as engine

agnessa:
This demonstrates how to use engines with extracted backend code. In this PR, locations & ghg emissions backend logic was removed from global platform and instead is pulled from a gem.


config/initializers/climate_watch_engine.rb
```diff
@@ -0,0 +1,15 @@
+ClimateWatchEngine.s3_bucket_name = Rails.application.secrets.s3_bucket_name
```
simaob: Aren't these only required when importing this datasets? do we need them to be always initialised when Rails starts?

agnessa: It's in the initializer, because this is the way to pass parameters to the engine. The importer is now part of the engine and the path to the file needs to be configurable; maybe it could be done using a service decorator in the host application rather than initializer, but that would make engine integration more cumbersome I think.

simaob: OK! =) It seems to make sense and I trust it does =)

simaob:
with the merge of the Rails upgrade we now might want to rebase this work?

agnessa:
yes, will do

agnessa:
rebased

simaob:
Thanks, let's wait for Marta's review and then she can merge if she's happy with it.

agnessa:
The refresh is called on two AR classes which represent materialised views; that needs to happen after data has been imported into the tables. These materialised views are used to speed up the search in the data explorer, which is why I did not include them & related logic in the gem.

agnessa:
Rails upgrade has been applied (engines needed upgrading as well)

PR #697, opened 2018-09-10
Upgrade rails to 5.2

aabdaab:
This PR upgrade rails to 5.2 version. 
It's needed as we want to use `ActiveStorage` for uploading files directly to S3. We could probably do that with a gem but `ActiveStorage` seems like a safe way to do that.
See ActiveStorage in the release notes -> https://guides.rubyonrails.org/5_2_release_notes.html#major-features


aabdaab:
comments addressed, ready again!

PR #699, opened 2018-09-13
Establish second database for shared resources

aabdaab:
One of the requirements from client is to have shared login credentials across country platforms. This PR proposed a second shared DB in a rails app which will hold the shared resources, like admins.

* Add this gem for adding a second database to the app: https://github.com/customink/secondbase
* Gem 'secondbase' requires adding config to database.yml
* Gem 'secondbase' creates the db on rake db:create command along with the default database
* Using 'secondbase' gem saves us a lot of boilerplate and makes the integration a lot easier than creating boilerplate for the second base by yourself


agnessa:
Did you have a look at https://github.com/customink/secondbase ? I'm not sure if it's compatible with Rails 5.2 (last update was for 5.1), but if it was it would save us a lot of boilerplate (making it easier to integrate & maintain the connection in country platforms as well) - last time I used it it worked well. There's also https://github.com/ankane/multiverse, which I haven't used but looks more recent. We could also, if those gems don't fit, turn this code into our own gem for easier integration in multiple apps.

aabdaab:
@agnessa I agree with adding the `secondbase` gem, it makes the integration really easy. Check out the latest commit; I removed the boilerplate from before and added `secondbase` instead along with the config for it in `database.yml` file. Ready to review again!

simaob:
Safe to merge? :D

simaob:
I have asked Tiago Garcia to setup a new database for this. These were named: `climate_watch_shared_db_staging` and `climate_watch_shared_db_production`. So make sure to update the deploy scripts to use them. If you need more details, please check with Tiago! =)

k8s/staging/deployment.yaml
```diff
@@ -62,6 +62,11 @@ spec:
                 secretKeyRef:
                   name: dbsecrets
                   key: CW_POSTGRES_URL_STAGING
+          - name: SHARED_POSTGRES_URL
+            valueFrom:
+              secretKeyRef:
+                  name: dbsecrets
+                  key: climate_watch_shared_db_staging
```
rrequero: Change for CW_SHARED_POSTGRES_URL_STAGING

PR #703, opened 2018-09-25
Enable uploading to S3, add pages for index and upload for country platforms

aabdaab:
This PR introduces ActiveStorage + AWS gem for uploading files to S3
- AWS gem is needed as ActiveStorage doesn't support uploading to specific folders in S3 bucket
- Add a dropdown menu for each platform divided into sections
- Add uploading files for specific datasets
- Add support for downloading single files
- Add support for zipping and downloading multiple files
- Integrate importers -> create workers for importers, add sidekiq
- For test purposes right now, everything uploads to `/test` folders, however importers are operating on already uploaded files to correct folder in S3.
- Add ability helper for conditionally showing dropdown with corresponding platform sections (for climate watch global show only dropdown for cw global, for india platform only show dropdown for india platform, you get the idea by now)

To test this locally, make sure that you have aws credentials in your `.env` file.
Run the migrations.
Run the rake task `rake db:admin_boilerplate_create` to create boilerplate for platforms, sections and corresponding datasets.
Seed the db with new AdminUser (there's already an example in `db/seeds.rb`)
Run `redis-server` and run `bundle exec sidekiq`
Then, go to `/admin` and log in to see the admin panel. 

This PR does not include content management (texts, translations).

Note that:
Each section (apart from Ndc Texts) is exactly the same:
- It has a table with its datasets to which we can upload associated datafile (one dataset - one datafile). The Upload button creates an association with ActiveStorage (for recognizing which datafiles are linked to which datasets) and uploads datafile to respective folder in S3.
- If datafile is uploaded, we can download that file (it's pulled from S3 and we again rely on the association with ActiveStorage to know which datafile it is)
- We can download all the files linked to the section (i.e. all the datafiles for Historical Emissions) -> zipped
- We can run the importers linked to the section (button `Run Importers`) - it triggers a worker, we can check the status of it in Sidekiq ----> TO IMPROVE

With all of the specifities of ActiveAdmin I couldn't think of an idea to make the pages generic (they all share the same functionalities but they redirect to different pages upon success etc (paths). If you've got an idea how to avoid repeating functionalities code, let me know. ActiveAdmin has its own DSL and even moving part of code out of `page_action` is not that straightforward. I tried concerns but failed to make them work. So, give me your thoughts ☺️


simaob:
Remember to run:

`rails db:admin_boilerplate_create`

To test it all, otherwise sections will be missing.

app/admin/global_country_platform/locations_members.rb
```diff
@@ -0,0 +1,87 @@
+ActiveAdmin.register_page 'Global Cw Platform Locations Members' do
```
agnessa: could the pages for locations and locations members be merged, or would that require merging the importers?

app/admin/global_country_platform/adaptation.rb
```diff
@@ -0,0 +1,87 @@
...
+      download_path: admin_global_cw_platform_adaptation_download_datafiles_path,
+      download_single_data_file_path: admin_global_cw_platform_adaptation_download_datafile_path,
+      import_path: admin_global_cw_platform_adaptation_run_importer_path
+    }
+  end
+
+  page_action :upload_datafile, method: :post do
+    dataset = Admin::Dataset.find(params[:dataset_id])
+
+    if params[:datafile].content_type != 'text/csv'
```
agnessa: this breaks when you hit "upload" without selecting a file, not a major issue but best avoid a 500

config/sidekiq.yml
```diff
@@ -0,0 +1,14 @@
...
+staging:
+  :concurrency: 2
+demo:
+  :concurrency: 2
+sandbox:
+  :concurrency: 2
+production:
+  :concurrency: 2
+:queues:
+  - database
```
agnessa: I'm having trouble running the background jobs, could be that I just don't know how to use active job properly, so can you check I'm doing this right: I started sidekiq with this config file and clicked "run importer", I can see in the web panel that the job is enqueued in the "default" queue, but that queue is not being processed by sidekiq as it's pulling from "database" queue.

config/sidekiq.yml
```diff
@@ -0,0 +1,14 @@
+verbose: true
+logfile: ./log/sidekiq.log
+development:
+  :concurrency: 2
+staging:
+  :concurrency: 2
+demo:
```
agnessa: sandbox & demo are Trase-specific and can be removed safely

app/admin/global_country_platform/adaptation.rb
aabdaab: thanks for catching that! I'll fix it now

config/sidekiq.yml
aabdaab: yep, I was peeking on Trase, thanks

app/admin/global_country_platform/locations_members.rb
aabdaab: I divided them as we have different importers for Locations and LocationMembers. So yes, it would require merging the importers

app/admin/global_country_platform/adaptation.rb
aabdaab: fixed!

config/sidekiq.yml
aabdaab: Again, thanks for catching that. I was doing some changes in the workers yesterday and I accidentally deleted that line `sidekiq_options queue: :database`.
Now, it's going to the correct queue.

app/admin/global_country_platform/adaptation.rb
```diff
@@ -0,0 +1,64 @@
+ActiveAdmin.register_page 'Global Cw Platform Adaptation' do
```
santostiago: When creating the gem, all these files could be generated using a template.

aabdaab:
Changes addressed, ready again!
Although I didn't add a validation for not being able to edit your own account; I think admin might want to change his password for example.

PR #708, opened 2018-10-23
Add laws and policies module

aabdaab:
This PR adds Laws and Policies module to country pages.
It adds new Slide Cards component and Laws and policies component as a parent.
It also modifies Card and Dropdown components but without any changes on existing usages of these components.

image

**Notes**:
- API is already hooked up
- Modal metadata is not there yet
- Intro text is not provided yet

Design can be found here:
https://app.zeplin.io/project/598b1dc89689e51e1cd1b5ce/screen/5bbc8acbd010065fdf0b91af

Few things are weird here as from the design I assumed there will be only two targets per card, although it's not the case sometimes. 
The flow is as follows: NDC Content card is always fixed per sector, there's only one per sector.
Targets are filtered 
-> by country
  -> then by sector, 
    -> then grouped by source (source is the title with link on the top of the card)
      -> and then by type 

The number next to sector dropdown is all targets altogether for this country (without NDC counting). The same goes for number on the LSE logo.

image

This is taken from target attribute `scope`.

So, in conclusion. The card represents targets from the source, and card's content is segregated by types, so we can have a situation as follows:
FIRST ROW:
Targets Type
Base Year Target

SECOND ROW:
Targets
Energy Efficiency
9% reduction in energy use

THIRD ROW:
Targets
Energy Efficiency
9% reduction in total energy consumption

and all three rows are related to this one target type and then we can have more targets in that same card.


Bluesmile82:
Can you remove the lorem ipsum until we have the final text?

simaob:
I believe the sentence next to the sector selector should update with your selection. The total laws and policies are on the card on the right. So when changing the filter it should reflect what is showing bellow, see in the screenshot it should say: _1 National Policy available for Brazil_.

screen shot 2018-10-25 at 12 01 03

simaob:
I see a small z-index issue with the sectors dropdown:

screen shot 2018-10-25 at 12 03 18

simaob:
> Can you remove the lorem ipsum until we have the final text?

I think it's fine for now, Álvaro. They will send it between today and tomorrow and for now it will only be on staging. And I told them we'd keep the placeholder for now.

Bluesmile82:
Not sure why but the country is undefined here: 
image

aabdaab:
`The total laws and policies are on the card on the right.` That's what I thought in the beginning but then had doubts and changed it to total laws for all the sectors. I thought it's confusing as it says for {Country} without specifying a sector but I guess you're right.

simaob:
When there's only 1 match on the right hand-side and then you move to an option that has no matches, there's a weird visual change, see screenshots bellow:

screen shot 2018-10-25 at 12 06 06
screen shot 2018-10-25 at 12 05 08

Can we keep the gray background always behind the cards, so that if there is only one, it would show the full space covered as well?

simaob:
> `The total laws and policies are on the card on the right.` That's what I thought in the beginning but then had doubts and changed it to total laws for all the sectors. I thought it's confusing as it says for {Country} without specifying a sector but I guess you're right.

True, maybe we can adjust the text by adding "for the selected sector" at the end.

app/javascript/app/components/card/card-component.jsx
```diff
@@ -1,28 +1,80 @@
...
 
 class Card extends PureComponent {
   // eslint-disable-line react/prefer-stateless-function
   render() {
-    const { title, subtitle, children } = this.props;
+    const {
+      title,
+      subtitle,
+      children,
+      theme,
```
Bluesmile82: 😎

app/javascript/app/components/country/laws-and-policies/laws-and-policies-component.jsx
```diff
@@ -0,0 +1,165 @@
...
+              <span
+                className={styles.logoText}
+              >{`See all ${allLawsTargets.length} ${countryName} national policies on Climate Change Laws of the World`}</span>
+            </a>
+          </div>
+        </div>
+        <div className={styles.actions}>
+          <Dropdown
+            className={styles.dropdown}
+            key="filter1"
```
Bluesmile82: Why do we need a key here?

app/javascript/app/components/country/laws-and-policies/laws-and-policies-selectors.js
```diff
@@ -0,0 +1,141 @@
...
+const CARDS_IN_ROW = 2;
+
+const getIso = state => state.iso || null;
+
+const getData = state => state.lawsAndPolicies || null;
+
+const getActiveSector = state => state.activeSector;
+
+export const getSectors = createSelector([getIso, getData], (iso, { data }) => {
+  const sectors = (data && data[iso] && data[iso].sectors) || [];
```
Bluesmile82: An empty array won't cache the null result. We can try to use null instead if its possible

app/javascript/app/components/country/laws-and-policies/laws-and-policies-selectors.js
```diff
@@ -0,0 +1,141 @@
...
+);
+
+export const getLawsAndPolicies = createSelector(
+  [getIso, getData, getSectors, getActiveSector],
+  (iso, { data }, sectors, currentSector) => {
+    if (!data) return null;
+    const activeSector =
+      (!isEmpty(currentSector) && currentSector.sector) || sectors[0];
+
+    const allTargets = (data && data[iso] && data[iso].targets) || [];
```
Bluesmile82: We use this several times so could be extracted to a different selector

app/javascript/app/components/country/laws-and-policies/laws-and-policies-styles.scss
```diff
@@ -0,0 +1,116 @@
...
+  position: relative;
+  overflow: auto;
+
+  @media #{$desktop} {
+    overflow: visible;
+  }
+}
+
+.fixedCard {
+  flex: 0 0 33%;
```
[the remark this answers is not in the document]
Bluesmile82: Such an annoying puppy

app/javascript/app/components/slide-cards/slide-cards-component.jsx
```diff
@@ -0,0 +1,137 @@
...
+import rightArrow from 'assets/icons/right-arrow.svg';
+import backArrow from 'assets/icons/left-arrow.svg';
+
+import cx from 'classnames';
+import styles from './slide-cards-styles.scss';
+
+class SlideCards extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { currentIndex: props.cards && 0, currentDot: 0 };
```
Bluesmile82: Won't will this always be 0? (props.cards && 0)

app/javascript/app/components/country/laws-and-policies/laws-and-policies-component.jsx
aabdaab: we don't really, I'll remove it

app/javascript/app/components/country/laws-and-policies/laws-and-policies-selectors.js
aabdaab: 👍 although the empty array has its advantages as you don't have to do additional checks for map, find, etc. functions

app/javascript/app/components/country/laws-and-policies/laws-and-policies-selectors.js
aabdaab: 👍

PR #720, opened 2018-11-02
Fix caching

aabdaab:
- Change the cache key to country iso
- Move caching to the service instead of controller


aabdaab:
I added don't merge yet label, I'm gonna try to debug with caching enabled on develop

simaob:
I've pushed a commit with updated info-window slugs, including an extra one besides the laws and policies one.

aabdaab:
It correctly goes into cache on develop, I say let's check on staging :P

PR #723, opened 2018-11-08
Show total matches of sector in dropdown

aabdaab:
- Set economy-wide as a default section
- Add text in the dropdown

image


aabdaab:
@simaob fixed 👷 
although two questions about it: the number displaying next to the sector in dropdown only includes the laws and policies now. Should it also include NDC? but then it might look confusing when next to the dropdown there is different number displayed (next to the dropdown we also have only laws and policies number without NDC)

aabdaab:
I removed the matches amount text from the dropdown options and marked the options in bold and black for the ones that have some content 
image

PR #733, opened 2018-11-21
The only issue with the .csv order was the year ordering

santostiago:
It was reversed.


agnessa:
I fixed the whitespace violations. Looks good to me, not sure if anything left to do here.

PR #735, opened 2018-11-27
Adjusts to the newly acquired knowledge about the LSE API

simaob:
The API has some source entries without IDs, so this changes our
grouping method to use title.

### DIFF
--- a/app/javascript/app/components/slide-cards/slide-cards-component.jsx
+++ b/app/javascript/app/components/slide-cards/slide-cards-component.jsx
@@ -63,9 +63,9 @@ class SlideCards extends Component {
                   contentFirst
                   title={{
                     link: card.source.link,
-                    title: `Targets ${card.source.title !== ''
-                      ? `found in ${card.source.title}`
-                      : ''}`
+                    title: card.source.title
+                      ? `Targets found in ${card.source.title}`
+                      : 'Targets'
                   }}
                   theme={{
                     contentContainer: styles.cardContentContainer,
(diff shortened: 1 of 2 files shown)


simaob:
> as titles are empty sometimes (http://localhost:3000/countries/MEX#laws-and-policies) maybe we can solve that on FE checking if title is truthy and give the card a different title: https://github.com/Vizzuality/climate-watch/blob/develop/app/javascript/app/components/slide-cards/slide-cards-component.jsx#L66

It seems that it's a bug on their API (see latest email from Phil), so maybe this won't happen again. But we can put a safeguard in place anyway. What do you think?

aabdaab:
We could have a generic title just in case

app/javascript/app/components/slide-cards/slide-cards-component.jsx
```diff
@@ -63,9 +63,9 @@ class SlideCards extends Component {
                   contentFirst
                   title={{
                     link: card.source.link,
-                    title: `Targets ${card.source.title !== ''
-                      ? `found in ${card.source.title}`
-                      : ''}`
+                    title: `${card.source.title !== ''
+                      ? `Targets found in ${card.source.title}`
+                      : 'Targets'}`
```
aabdaab: you don't need to check if it's empty like this: ${card.source.title !== '' 
empty string is falsy in js so you can shortify it to: card.source.title ? `Targets found in .....` : 'Targets'

PR #741, opened 2018-12-13
Feature/ghg redesign

Bluesmile82:
GHG redesign:
- Use cw-components chart and multiselect
- Add Percentage chart type option
- Multilevel sector (There are 3 levels of sectors we are showing only the first two ones)
- Break by prepared for metric but we don't have the population and GDP data for the Regions so we just break by model right now.
- Merge source and version dropdown
- All selected added to sectors and gases

Extra: Fixed reducers problem with some instalations

---
- Only line chart type can be selected if we have intersecting regions selected. Info hover added
- One region selected => Expanded into countries and 'Others' (only if more than 10 countries), More than one region selected or region and country => region back to collapse
- The legend options can't be deleted if one region is expanded
- The non-others countries on the expanded region are selected based on the biggest last year value
- Add metrics for region options in Break by dropdown. The regions are now calculated with the countries aggregated values but at some point when we have the values it will be done with the region wb endpoint values.
- Color function fixes

TODO: 
- There is a known bug with the 'Others' of one expanded region in the stacked area view. There is a PR in components for that

image


simaob:
Let's not merge this branch until it's fully done. I will merge it to sandbox for showcasing on staging.

PR #743, opened 2018-12-18
Feature/user case slider

weberjavi:
This PR adds `How can Climate Watch help you` section.
Remember to run `yarn install` as 'react-slick' dependency has been added.
Designs: 
https://projects.invisionapp.com/d/main#/console/16156416/335004355/preview
slides:
https://projects.invisionapp.com/d/main#/console/16156416/335004754/preview
https://projects.invisionapp.com/d/main#/console/16156416/335004752/preview
https://projects.invisionapp.com/d/main#/console/16156416/335004753/preview

kapture 2018-12-18 at 17 53 43


weberjavi:
Updated `cw-componets` version on this PR ☝️ with the [carousel fix](https://github.com/ClimateWatch-Vizzuality/climate-watch-components/pull/100)

aabdaab:
image

^ The circles are clipped, I think you need to add some margin-left

app/javascript/app/pages/home/user-cases-section/user-cases-section-component.jsx
```diff
@@ -0,0 +1,177 @@
...
+            }}
+            beforeChange={(current, next) => this.leftSlider.slickGoTo(next)}
+            {...settings}
+          >
+            <div className={styles.slideWrapper}>
+              <h3 className={styles.slideTitle}>
+                How can Climate Watch help you?
+              </h3>
+              <ol className={styles.slideList}>
+                <li>
```
aabdaab: hmm, maybe we could extracts the texts to a table and loop through them, it got kind of crowded here. True also for the `li`s beneath

kfilip94:
- I think you can fix spacing here, text line height is different and "How can Climate Watch help you?" section width is too big.
untitled
- Also, I see icons are a little bit bigger here, with smaller icon border, 
- space between top border and icon is smaller, 
- font color for inactive icons isn't this same:
untitledd
Two design issues, maybe it is worth to ask Fanny or Fausto:
- Maybe it'll be good to add a hover effect for icons?
- We should make this page more responsive, for lower resolutions, maybe set one column layout or hide the picture?

weberjavi:
@aabdaab , @kfilip94 comments addressed ☝️ 
Thanks for your reviews!! 🙏

app/javascript/app/pages/home/user-cases-section/user-cases-section-styles.scss
```diff
@@ -42,63 +55,75 @@
...
+      display: inline-block;
+      text-shadow: 0 2px $white;
+
+      &::before {
+        content: "";
+        width: 100%;
+        height: 2px;
+        background-color: $theme-color;
+        position: absolute;
+        bottom: 3px;
```
aabdaab: I think you can set it to `1px` otherwise in some cases the underline overlaps with the text: 
image

weberjavi: @aabdaab I'm not being able to reproduce this issue (could it be some cache related thing?)
In which browser are you testing?

weberjavi:
@aabdaab changed approach (instead of a `pseudoelement` I'm using `text-shadow` and `box-shadow` combined in the anchor styles), please check if it's working now ☝️

aabdaab:
It depends on the resolution; on smaller resolution than full hd the problem remains on both firefox and chrome. You can achieve that effect on long letters using simple `text-decoration: underline` as chrome webkit added it as a default, but then again, it would have that effect only in chrome. So, both solutions have their caveats.

kfilip94:
Thank you for changes, looks good :+1: 
One thing I'm considering is the image behavior when width is near 1000px:
screenshot from 2018-12-20 10-41-04

 I think we can remove `` tag witch `src` and set parent background image like:
   `background-image: url(image.jpg);`
    `background-position-x: center;`
    `background-size: cover;`
It should fit image to the parent div, like that:
screenshot from 2018-12-20 10-48-23
But this is my suggestion, let me know what do you think about it. :)

weberjavi:
Ready again @aabdaab , @kfilip94 ☝️

app/javascript/app/pages/home/user-cases-section/user-cases-section-component.jsx
```diff
@@ -51,13 +51,22 @@ class UserCasesSectionComponent extends Component {
               fade
             >
               <div>
-                <img src={politicianImg} alt="Political building" />
+                <div
```
kfilip94: Maybe loop here will be a good idea :)
Do we need this <div> parent for each child?

weberjavi: Sorry 🤦‍♂️ 
Yeah the parent is needed because of the way the slider works under the hood.

kfilip94: I thought so, thanks :)

PR #744, opened 2018-12-18
Add latest updates to the admin panel + create an endpoint

aabdaab:
This PR adds latest updates to the admin panel and creates and endpoint for the frontend to get the last 3 updates.
Update has a `category` (i.e Feature, Data), `description`, `link` and timestamps. For the frontend we return `date` attribute which is in fact `updated_at` field.


app/serializers/api/v1/update_serializer.rb
```diff
@@ -0,0 +1,10 @@
+module Api
+  module V1
+    class UpdateSerializer < ActiveModel::Serializer
+      attribute :category
+      attribute :link
+      attribute :description
+      attribute :updated_at, key: :date
```
simaob: should we format the date here? or is it OK for the front to do it? (We don't need the time)

aabdaab: true, no need for FE to do it

aabdaab:
Ready; I updated the base branch as well

simaob:
@kfilip94 and @weberjavi : when pulling from your base branch don't forget to run `rails db:migrate`!

PR #747, opened 2018-12-20
Feature/new home footer

weberjavi:
This PR adds the site map to home footer
image


kfilip94:
Map section looks good, thanks :+1: I have two questions: 
- Should country platform links be connected to staging sites? Because now they are not
- I think this section should be a part of the footer what means it should be present on every page of CW site.

Also, there are some changes in `Partners` section, maybe worth to do it in this PR :
- Partners title should be placed on the line,
- Partners logos should be in color,
- Last section in the footer is a bit different (`Climate Watch © 2018`, ` Powered by ...`)
screenshot from 2018-12-21 10-48-49

weberjavi:
@kfilip94  ready again ☝️

PR #748, opened 2018-12-21
Feature/subscribe section

kfilip94:
This PR adds the `Sign for the Updates` section. There is no backend side yet so it's only the front end component
screenshot from 2018-12-21 17-47-40
After subscribe:
screenshot from 2018-12-21 17-47-53
There are still a few things that need to be done before production release like:
- add email validation for input,
- whole backend communication


app/javascript/app/pages/home/home-component.jsx
```diff
@@ -48,6 +49,9 @@ const Home = () => (
     <div>
       <HighlightedStories />
     </div>
+    <div>
```
aabdaab: I don't think it's needed, can you check?

aabdaab: otherwise looks nice 👍

kfilip94: I know, thanks

PR #751, opened 2019-01-08
Feature/ghg redesign gwp update

simaob:
This PR picked up on Álvaro and Agnieszka's work of the following PRs:

- https://github.com/Vizzuality/climate-watch/pull/741
- https://github.com/Vizzuality/climate-watch/pull/662

Merged it together and is now updating the frontend to get rid of references to GWP, tasks there are:

- [x] Update GHG emissions module to use only the existing data sources, and not merge them with GWP;
- [x] Update Country Profile to use data sources without GWP;
- [x] Update Data Explorer to not use GWP and use just the pure Data Sources;
- [x] check what's up with the performance of loading emissions data;
- [X] Make sure compare page is still working;

- [x] Update GHG emissions module to filter sectors, regions, gases based on the relation to data sources
- [X] Make sure both embed and contained versions of all the pages work as well
- [X] Make sure that Data Explorer and GHG modules are linked with interchangeable params: Make multiselect: Regions, Sectors, Gases
- [X] Solve Sectors mismatch with Data explorer and filter data in Data explorer depending on Source (to discuss)


Bluesmile82:
This is clean again. Maybe we should do a feature flag and merge it with develop before addressing the feedback. Is growing a lot

simaob:
Alright. What would the feature flag encompass? all the changes? it feels a bit scary and hard to fence it. But I do think we should merge.

simaob:
We might be able to manage it so that we release home, sectors and this all at the once when they are all ready.

simaob:
Thoughts, Álvaro ^?

simaob:
@Bluesmile82 regarding the colours, having talked with Fausto we can update them to use the ones on the UI kit: https://projects.invisionapp.com/d/main#/console/16156416/337855930/preview

simaob:
I think there might be a small issue, not sure if it's with the data. On the Data Explorer for GHG emissions, CAIT Source, Bunker fuels shows as a parent sector with Bunker fuels as a subsector? Is that because the way the data is encoded or is it a bug?

Bluesmile82:
Might be a bug. Checking

PR #753, opened 2019-01-10
Fix/home page icons styles

weberjavi:
This PR updates webpack version to `v.4.28` to properly order the styles assets avoiding some `CSS` cascading issues (`cw-component` own styles where overriding the `theme` styles).

It also updates some files inside `config/webpack`.
To mimic `production` environment locally (to check nothing is going to break in production) this steps should be followed:
- On `config/secrets.yml` use the same `secret_key_base` in `production` that in `development`.
- On `Gemfile` move the `dotenv-rails` gem outside the `:development`, `:test` group and run `bundle install`
- Run `RAILS_ENV=production bundle exec rails assets:precompile` to create the `public` folder from which the `production` environment will consume the assets.
- Run the server like this `RAILS_SERVE_STATIC_FILES=true RAILS_ENV=production bundle exec rails server`


simaob:
Since you are offline already, I did it myself. Closing this PR in favour of this new one: https://github.com/Vizzuality/climate-watch/pull/754

PR #760, opened 2019-01-21
Add emissions pathways graph for Agriculture profile

aabdaab:
Add Future Pathways graph for Agriculture profile.
Basically copied the code for the emissions pathways and cleaned it up from selecting category & subcategory.

image


simaob:
> Is it possible to replace `_` signs with spacebars for region labels in selector?
> screenshot from 2019-01-22 12-24-51

I was checking and we have the same issue on the main Pathways page, so I'd leave it to fix at a later stage. Given that it's probably more of a change to the original API.

PR #761, opened 2019-01-21
Add Countries in their NDCs - Map & Table

aabdaab:
Add the Map && Table part of the section `Countries in their NDCs`.

**Map part**
image

**Table part**
image


kfilip94:
Good job Marta!
I think you should put a counted number of countries instead of placeholder `(X)` on section description.
Also, I'm not sure about dropdown `Category` and a map legend: should they be named consistent with the design (`Adaptation actions`, `yes, no, no data` on the map legend) or like in `/ndcs-content`?
screenshot from 2019-01-22 10-58-42

aabdaab:
We agreed that finally the legend should be whatever it is on the NDC Content section; so no simplyfying to `No` or `Yes`

aabdaab:
I'm gonna remove the `vulnerability_agriculture` from the available indicators as it should be only _Adaptation actions_ and _Mitigation actions_

simaob:
This is looking good to me, the only issue I see is the same we have on the main NDC table, when we sort the elements on the table, the links of the countries are not sorted. So Afghanistan is always the first link. This bug is in this Task as well: https://www.pivotaltracker.com/story/show/162235610

So maybe if we could fix it and apply the fix on the main table, that would be great!

aabdaab:
I fixed the links issue in this PR but only for this table; in the main table there is some complicated logic for the title links so I will have to look at it more closely later, maybe after I'm done with the Agriculture Profile, ~early next week.

PR #765, opened 2019-01-28
Fetching data about gas and source of ghg from meta provider

kfilip94:
This small PR fetches data from GHG meta provider and select ids of 'All GHG'  gas and 'CAIT' source and then fills GHG filter object with this metadata.


kfilip94:
@weberjavi Thank you! :+1: Right now I'm working on another PR with chart filters, I'll solve this problem there :)

PR #768, opened 2019-01-29
Feature/home page rebased with develop

simaob:
This PR brings the new homepage to an earlier version of develop. We had to do the rebase again because something was amiss.

This `preserve_develop` is a version of the develop branch before the previous attempt. if this works, I'll replace the existing develop with this one and folks in the project should run:

```
git checkout develop
git reset --hard origin/develop
```

To sync back.


simaob:
I think we actually don't need this anymore.

PR #770, opened 2019-01-30
Agriculture production consumption trade

aabdaab:
This PR adds Bar Chart for presenting meat consumption, production and trade. It's added to Country Context section for Agriculture page.

image

**TO DO**
Stacked data for other countries is yet to be added as I'm waiting for this task to be done: https://www.pivotaltracker.com/story/show/163569429

### DIFF
--- a/app/javascript/app/components/sectors-agriculture/countries-context/meat-data/meat-data-component.jsx
+++ b/app/javascript/app/components/sectors-agriculture/countries-context/meat-data/meat-data-component.jsx
@@ -1,8 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import MeatConsumptionProvider from 'providers/agriculture-meat-consumption-provider';
+import MeatWorldConsumptionProvider from 'providers/agriculture-world-meat-consumption-provider';
 import MeatProductionProvider from 'providers/agriculture-meat-production-provider';
+import MeatWorldProductionProvider from 'providers/agriculture-world-meat-production-provider';
 import MeatTradeProvider from 'providers/agriculture-meat-trade-provider';
+import MeatWorldTradeProvider from 'providers/agriculture-world-meat-trade-provider';
 import { Card, Icon, Chart, Dropdown } from 'cw-components';
 import NoContent from 'components/no-content';
 import infoIcon from 'assets/icons/info';
@@ -12,7 +15,8 @@ import styles from './meat-data-styles.scss';
 
 const cardTheme = {
   card: styles.card,
-  data: styles.data
+  data: styles.data,
+  contentContainer: styles.contentContainer
 };
 
 const MeatData = ({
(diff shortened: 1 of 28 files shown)


kfilip94:
Hey Marta, I checked this :meat_on_bone: section and it looks great :+1:, just few things to fix below:
- Is it possible to set a white background for filter dropdowns and legend elements?
- Could you change content padding, marked below? I think these spaces are a bit different on the design
screenshot from 2019-01-30 14-32-51
- Also I think you can improve width scaling of filter dropdowns, especially for lower resolutions because  categories list is almost unreadable
screenshot from 2019-01-30 14-49-16
I'll check the code now

app/javascript/app/components/sectors-agriculture/countries-context/meat-data/tooltip/tooltip-component.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const countryColor = get(config, 'theme.yCountry.stroke') || '';
+  const othersColor = get(config, 'theme.yOthers.stroke') || '';
+  const unit = get(config, 'axes.yLeft.unit') || '';
+
+  return (
+    <div className={styles.tooltip}>
+      <div className={styles.category}>{category}</div>
+      <p className={styles.info}>
+        <Tag
```
kfilip94: I think you can extract <Tag /> component to separate method `renderTag()` and pass name, value and color properties.

app/javascript/app/components/sectors-agriculture/countries-context/meat-data/meat-data-component.jsx
```diff
@@ -0,0 +1,141 @@
...
+              />
+            )}
+          </div>
+        </div>
+        <div className={styles.yearData}>
+          <span>{selectedYear && selectedYear.value}</span> data
+        </div>
+      </Card>
+    )}
+    {selectedCountry &&
```
kfilip94: I think you can save params as a one object (before `return (...)`):
`const params = selectedCountry && selectedYear && { country: selectedCountry.value, year: selectedYear.value };`
and then use `params` object as a guard here:
`{params && [ <MeatConsumptionProvider... ]}`

simaob:
Ready again?

kfilip94:
I'm checking this chart again and I think I found something worth to discuss:
- most of the countries here have a marginal contribution to global meat production/consumption/trade:
screenshot from 2019-01-31 11-58-03
We don't see here any data for selected country because it's too small to be visible on this chart in relation to other countries.
So in most of the cases bar chart doesn't show us valuable information, only tooltip does.
- Also there are cases when we don't have a data for selected country at all, so maybe we should set 'no data available' message instead of showing bar chart without country data?
screenshot from 2019-01-31 12-13-45

kfilip94:
Marta, thanks for changes! :+1: For me it looks good, we can discuss this comment above later

PR #772, opened 2019-01-31
Add countries number to tooltip meat data

aabdaab:
- Fix agriculture page crashing after changes in the metadata API for meat consumption, production and trade
- Add number of countries to the tooltip to `Other countries` for each indicator (hover on the bar chart to see it)


kfilip94:
it works

PR #773, opened 2019-01-31
Fix hash section navigation in sectors/agriculture

kfilip94:
This tiny PR adds missing styles to section hash in sectors-agriculture component and fix cropping sections after click on sections menu
**Before:**
screenshot from 2019-01-31 15-20-34

**After:**
screenshot from 2019-01-31 15-41-37


kfilip94:
Just wait a sec before merge, because it's not 100% fixed yet, for resolutions between mobile and landscape still doesn't work perfectly :disappointed:

PR #775, opened 2019-02-04
Feature/cc by indicator

weberjavi:
This PR adds `select by Indicator` to Countries Context section on Agriculture Profile.
[InVision](https://projects.invisionapp.com/d/#/console/16383011/339735320/preview)
[Pivotal task](https://www.pivotaltracker.com/story/show/163084689)
This is a first version with out the _top ten_  bar chart.

kapture 2019-02-04 at 11 37 51

Maybe some thinking is needed about data ranges on the map (no color buckets where given beyond `fertilizer consumption` indicator)


aabdaab:
Hey! I found two things about the **tooltip**:
1) I think we shouldn't use SI prefix for the percentage values:
screenshot 2019-02-04 at 11 01 32

2) On the design we've got also the name of the indicator in the tooltip, like so:
image

kfilip94:
I think it would be worth to check tooltip styles for maps in `Understand Countries’ Contexts` and  `Countries’ Actions in their NDCs`; according to the design, they should be white, without border radius and bottom arrow, overall it looks nice :+1:

weberjavi:
@aabdaab ☝️ addressed comments

PR #777, opened 2019-02-05
Feature/agriculture filter by metric

kfilip94:
This PR adds `Break by metric` filter, now we can filter data by GDP (per million $) and by capita
Design: [here](https://projects.invisionapp.com/d/main#/console/16383011/339735320/preview)
screenshot from 2019-02-05 11-07-14


app/javascript/app/utils/defaults.js
```diff
@@ -0,0 +1,5 @@
+export const METRIC_OPTIONS = {
```
weberjavi: Wouldn't it be better to store this values in `data/constants`, as in `utils` we are mainly exposing functions?

simaob: Kamil, after this change is done, we are good to merge!

kfilip94: @weberjavi Thanks for spotting this, now I see we already have `METRIC_CONSTANT` in `data/constants` so I used this one :)

PR #779, opened 2019-02-06
Feature/top ten countries by indicator

weberjavi:
This PR adds top ten countries chart to Countries Context by indicator section on Agriculture profile.
Pivotal task:
https://www.pivotaltracker.com/story/show/163084689
Designs:
https://projects.invisionapp.com/d/main#/console/16383011/339735320/preview

kapture 2019-02-06 at 17 31 17

### DIFF
--- a/app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
+++ b/app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
@@ -107,6 +107,7 @@ const ContextByIndicatorComponent = ({
                     <li key={c.value} className={styles.countryData}>
                       <span
                         data-label={c.label}
+                        data-value={c.valueLabel}
                         style={{
                           width: `${c.chartWidth}%`,
                           height: '12px',
@@ -114,7 +115,6 @@ const ContextByIndicatorComponent = ({
                           display: 'block'
                         }}
                       />
-                      <p className={styles.value}>{c.valueLabel}</p>
                     </li>
                   ))}
                 </ul>
--- a/app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
+++ b/app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
@@ -81,7 +81,7 @@ const ContextByIndicatorComponent = ({
           </div>
         </div>
         <div className={styles.visualizationsContainer}>
-          <div>
+          <div className="layout-container">
             <Map
               paths={paths}
               tooltipId="cc-map-tooltip"
--- a/app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
+++ b/app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
@@ -81,19 +81,21 @@ const ContextByIndicatorComponent = ({
           </div>
         </div>
         <div className={styles.visualizationsContainer}>
-          <Map
-            paths={paths}
-            tooltipId="cc-map-tooltip"
-            onCountryClick={() => {}}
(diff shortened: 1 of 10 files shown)


aabdaab:
The countries chart is looking good 👍 
I found a bug in the map tooltip though:
image
When there's no data it should say `No data` (in some cases it's working correctly).

aabdaab:
Also, on smaller screens I'm experiencing this bug with tooltip staying when scrolling away from the map, I'm not sure if it's just me, could you check on your side?
tooltipbug

app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
```diff
@@ -49,60 +50,105 @@ const ContextByIndicatorComponent = ({
...
+          />
+          <div className={styles.buttonGroupWrapper}>
+            <ButtonGroup
+              className={styles.btnGroup}
+              buttonsConfig={buttonGroupConfig}
+            />
+          </div>
+        </div>
+        <div className={styles.visualizationsContainer}>
+          <div>
```
aabdaab: I think we don't need that extra div here

app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
```diff
@@ -49,60 +50,105 @@ const ContextByIndicatorComponent = ({
...
+            <Map
+              paths={paths}
+              tooltipId="cc-map-tooltip"
+              onCountryClick={() => {}}
```
aabdaab: maybe if there's no explicit action to do in this case we could pass `undefined` instead of an empty function?

kfilip94:
I don't know why but for e.g. `Male employment` in 2002: Chad (82.3%) has shorter bar then Rwanda where value is 82%: 
screenshot from 2019-02-06 17-42-31

weberjavi:
> The countries chart is looking good 👍
> I found a bug in the map tooltip though:
> image
> When there's no data it should say `No data` (in some cases it's working correctly).

So this happens in countries that where not sovereign in the year selected (It happens on the USSR, Yugoslavia, Ethiopia and Sudan), they only have available data or `no data` from their independence year onwards.
How should we handle this?
Any proposal?

aabdaab:
Javi, so the issue here is that for some of the `iso`s you get no data for a given year, right? Then maybe we could iterate through the paths, after updating them with what's in the API, and filling the ones that are empty with the label `No data`

simaob:
A couple of other small tweaks, if you don't mind me barging in: 

- sort indicators' dropdown alphabetically;
- sort years' dropdown from most recent to latest;

Tks!

app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
weberjavi: It is used as container for Map and Legend, I've added a placeholder `layout-container` class to be explicit that they are a `column` layout element.

app/javascript/app/components/sectors-agriculture/countries-context/context-by-indicator/context-by-indicator-component.jsx
weberjavi: ⚡️

weberjavi:
☝️ addressed comments.
There's one thing left behind though. The map tooltip not being hide after scroll on mobile.
After trying to make it work through `react-tooltip` event handlers (`event`, `eventOff` and `globalEventOff`) and failing I've just adjusted the `delayHide` property, but as scrolling should fire the `hide` function it just makes the tooltip to remain there for two more second after the `scroll` event is over.
Further investigation would be needed to address this bug.
Do you agree that we handle it in another PR?

simaob:
Yes, I think it's OK to handle that in another PR.

simaob:
I've added it.

PR #780, opened 2019-02-06
Fix/ghg feedback

Bluesmile82:
This is the first part of changes adressing GHG feedback:

- Icons in icon dropdown larger (updated to cw-components 1.37.1) - Not working - Needs fix
- Change download button tooltip text to: 'View or download raw data'
- Add 't' as a suffix for the unit
- Disable stacked chart and stacked 100% chart when sectors + subsectors or All GHG + other gases- Rephrase the info icon in the chart type and show only if there is a conflict


weberjavi:
🤔 not being able to see dropdown icons with the new size.
I've updated dependencies `yarn install`, am I missing something?

Bluesmile82:
The icon styling is not working in CW but it works in the components showcase. I'm checking it. Any ideas? I wonder if this is related also with the multi level dropdown component not working in CW.

Bluesmile82:
Ok, let's leave the Icon bug for later. I have resolved the conflicts. Let me know if it's Ok to merge

PR #785, opened 2019-02-08
Fix/ghg feedback tasks

Bluesmile82:
- Default to ALL GHG when gases dropdown is cleared
- Fix metadata info button (Show UNFCCC metadata for both UNFCCC options)

Note: The subscript not showing in gases is a conflict with React Truncate component but I didn't have time to solve it. Maybe we could find an alternative but it has to have 'truncation' of several lines


simaob:
It's looking good! =)

If anyone else is going to test it, make sure you re-import your data for historical emissions:

`rails historical_emissions:import` (it takes a bit of time).

One thing I've noticed and not sure is registered in the Open Stories: we need to reset the sectors selection when changing sources, as the sectors are different per source.

simaob:
Hmm, it seems I found something weird, there's an extra item in the tooltip that doesn't appear in the legend:

screenshot 2019-02-08 at 18 07 29

weberjavi:
☝️ taking care of that one

PR #786, opened 2019-02-11
Fix/share links with section

weberjavi:
This PR adds the section hash to the URL to be consumed by the share button on Agriculture profile page.
[Pivotal task](https://www.pivotaltracker.com/story/show/163790009)

### DIFF
--- a/app/javascript/app/components/sectors-agriculture/countries-actions/ndcs-map/ndcs-map-component.jsx
+++ b/app/javascript/app/components/sectors-agriculture/countries-actions/ndcs-map/ndcs-map-component.jsx
@@ -35,7 +35,6 @@ const renderButtonGroup = (clickHandler, downloadLink) => (
       },
       {
         type: 'share',
-        shareUrl: '/embed/ndcs',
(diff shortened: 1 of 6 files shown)


simaob:
One thing that maybe can be fixed on this PR, the map for "Countries actions on their NDCs" has the share button of the original NDC component, and is copying the embed for that. Let's remove that embed link fro the Agriculture page, please!

kfilip94:
@simaob btw. should we add buttons group to `Countries’ Actions in their NDCs/Table` section?
we don't have a design for that but probably buttons should be there, right?

simaob:
Hmmm we probably should, but we don't have it here: https://climate-watch.vizzuality.com/ndcs-content/table either... so maybe let's leave it as is for now. It wasn't on the designs before, and they approved it.

kfilip94:
thanks :+1:  btw2. I see we have this same case with embed links in Drivers of emissions/Future pathways section

simaob:
Let's get rid of it there as well! =)

weberjavi:
No more embed links ☝️

PR #787, opened 2019-02-11
Feature/explore emissions button

weberjavi:
This Pr adds functionality to `Explore Emissions` and `Explore Pathways` button on Agricuture's `Drivers of emissions` section.
[Pivotal task](https://www.pivotaltracker.com/story/show/163790016)
[Designs](https://projects.invisionapp.com/d/main#/console/16383011/339735319/preview)


kfilip94:
Good job :+1:
I found one thing: when we click on `Future Pathways` and then, before the data will be loaded, click on `Explore pathways`, pathways page will open but query params will be null/undefined:
`http://localhost:3000/pathways/model?currentLocation=undefined&category=null&indicator=null&model=null`
ezgif-1-8202aec42a58
Should we set default search?

weberjavi:
Ready to review ☝️ 
Thanks @kfilip94 for spotting the bug 🔎 🐛 
Added the default link.

kfilip94:
Thanks for fixing this :+1: But there is one more thing :cry:
For countries where we don't have any data about pathways, after click on explore emissions, the default link is set as `/pathways/model` but I think it would be good to pass country even if rest of the params is empty.
e.g. we don't have here category and indicator data for Lithuania, but we have location and model:
`location: {label: "Lithuania", value: 146} , 
category: null, 
indicator: null, 
model: {label: "Global Change Assessment Model"}`

weberjavi:
@kfilip94 ready again ☝️

PR #788, opened 2019-02-12
Feature/responsive agriculture page

kfilip94:
This PR implements responsive design for agriculture profile page (demo below)
[Pivotal task](https://www.pivotaltracker.com/story/show/163628149)

**Drivers of emissions**
ezgif-1-6bd66b829752
**Countries Context**
ezgif-1-b50b76382751
**Countries’ Actions in their NDCs & Resources for Action**
ezgif-1-f0a8b4a74670

### DIFF
--- a/app/javascript/app/components/sectors-agriculture/countries-actions/countries-actions-styles.scss
+++ b/app/javascript/app/components/sectors-agriculture/countries-actions/countries-actions-styles.scss
@@ -16,20 +16,21 @@
 
 .descriptionWrapper {
   display: flex;
+  flex-direction: column;
   margin: {
     top: 20px;
     bottom: 50px;
   }
 
   justify-content: space-between;
-  align-items: center;
+  align-items: flex-start;
 }
 
 .description {
   color: $theme-color;
   font-size: $font-size;
   max-width: 600px;
-  margin-right: 20px;
+  margin-bottom: 30px;
 }
 
 .switchWrapper {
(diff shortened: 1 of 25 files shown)


weberjavi:
So, every thing is responsive but as we do not have designs for this I have some proposals :)
- Maybe would be nice to stack the `dropdowns` (two in each row taking the full width of the page) when the screen is smaller than `900px` as we are doing on [GHG Emissions](http://localhost:3000/ghg-emissions)

image

-  Also I think would be better if the tab switcher gets into its own row on smaller than `900px` screens, to make the text breath some more and avoid the tab title to wrap in two lines.

image

- And maybe have the `dropdowns` and the tabs fill the whole `width` of the screen on mobile, each one on its own row?

image

@faustoperez would love to hear your thoughts ☝️

faustoperez:
Hey @weberjavi you're right, all tabs change to `dropdown` on responsive, taking the whole `width`, like so:
image

Here's a link to the Invision screen: https://invis.io/SDPG3CSA2Q9#/337855878_GENERAL_UI_COMPONENT_Mobile

Thanks! 🤙

kfilip94:
@weberjavi @faustoperez Thank you, I'm working on it :man_factory_worker:

weberjavi:
@kfilip94 let me know when it is ready ☝️

kfilip94:
@weberjavi ready again :arrow_up:

weberjavi:
👁 🐽 👁

kfilip94:
@weberjavi :arrow_up: ready again

PR #790, opened 2019-02-13
Fix/tab change on agriculture drivers of emissions

weberjavi:
This PR avoids tab switch when selecting any value on Drivers of emissions (`Future Pathways`) `dropdown`.
Task: https://www.pivotaltracker.com/story/show/163941397

It also updates custom message to specify that the selection does not have matches under `Agriculture` category on the selected model.


kfilip94:
LGTM :ok_man:  :+1: thanks

PR #793, opened 2019-02-19
Feature/countries contexts update socioeconomics card

weberjavi:
This PR updates Socio-economic indicators to display gender gap on agriculture sector
[Pivotal task](https://www.pivotaltracker.com/story/show/164073773)
[BaseCamp thread](https://basecamp.com/1756858/projects/13795275/messages/83139911#comment_681196797)
[Test locally](http://localhost:3000/sectors/agriculture?contextBy=country&country=AFG&countryYear=2009&indicatorYear=2015)

image

### DIFF
--- a/app/javascript/app/components/sectors-agriculture/countries-context/context-by-country/context-by-country-component.jsx
+++ b/app/javascript/app/components/sectors-agriculture/countries-context/context-by-country/context-by-country-component.jsx
@@ -87,30 +87,52 @@ const ContextByCountryComponent = ({
                   dangerouslySetInnerHTML={{ __html: c.text }}
                 />
                 <div className={styles.cardContent}>
-                  <div className={cx(styles.textHtmlWrapper, styles.legend)}>
-                    {c.legend &&
-                      c.legend.map(i => (
+                  {c.legend && (
+                    <div className={cx(styles.textHtmlWrapper, styles.legend)}>
+                      {c.legend.map(i => (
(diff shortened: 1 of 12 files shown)


kfilip94:
Hello, I found a few bugs here:

**1.** When we change `year` filter from year with no data to year with data:
th3ht-fkhpt
**2.** When we change Countryfilter  to e.g. `Andorra`:
omjw3-w0fsa
Error in browser console shows this line in _countries-contexts-selectors.js_:
screenshot from 2019-02-19 16 29 22

**3.** For countries and years where we don't have any data, I think it'll be better to display info message like "No data found" instead of displaying chart with 0% values.
screenshot from 2019-02-19 15 54 05

weberjavi:
Hey @kfilip94 ready again ☝️ 
I've refactored a bit and added a loader (also modified the provider to avoid fetching when changing countries)

kfilip94:
@weberjavi Thanks for changes, great!
I found another bug, probably additional guard will solve this:
- When you'll try to change country to `Cooks Islands`, the page will crash with error:
screenshot from 2019-02-21 14 53 02
in this line:
screenshot from 2019-02-21 14 49 24

Also, I think you can add a comma here, after a country name: `No data for Andorra, please select another country`

weberjavi:
☝️ ready again
@kfilip94 thanks a lot for spotting this crashes 🙏

PR #796, opened 2019-02-26
Show ABOUT nav with nested menu

simaob:
WRI wanted to make the FAQ more easily discoverable, so we thought that we could update the About entry on the navbar to be a nested navigation.

This PR adds an extra file that includes the routes to be nested under that option. I have tried to follow similar pattern to what we have for the NDCs, though for the about the core routes are in a separate file, unlike the NDCs. But hope this is OK!


simaob:
@weberjavi or @kfilip94 can you guys have a quick look at this? =) Thanks!

PR #798, opened 2019-02-26
Feature/ghg module more changes

tsubik:
Changes to GHG module

- new aggregatedBy field for sectors to indicate sector aggregates/totals,
- displaying selected filters conflicts and possible resolutions instead of disabling options,
- improved importer performance (on my machine down from 700s to ~180s), there is still room for improvements, with `activerecord-import` gem it was 100sec, but using it would require time and thinking how to preserve per row validations, but that is doable,

Task: https://www.pivotaltracker.com/story/show/164011839

image

```
bundle exec rake db:migrate
bundle exec rake historical_emissions:import
```


tsubik:
I'm going to double test it after lunch as I updated cw-components.

simaob:
This is looking great, Tomasz! =) I really like the status message, some comments and small tweaks and some crashes I've seen:

- with source CAIT if I select sector "Total Excluding LUCF" and change the Break by to "Sector" I'm getting an error in the console and the page turns blank, so some guard is probably missing [it doesn't seem to happen if there's a second sector selected];
- is it possible to separate the Totals/Aggregated options at the top of their selector with "Aggregated" or something like that. Similar to what we have with REGIONS and COUNTRIES separation.
- replace "conflicted" with "conflicting" (just spell check in this one!);

Gonna test it further!

simaob:
On first load, if I change from Break by region to Break by sector the sectors are not "exploded", I still see the aggregated sector.

tsubik:
I created a separate task for adding this `optgroup` feature to multileveldropdown because this control does not have it.

simaob:
Right, makes sense! Anyway is not the most urgent thing for sure. =)

tsubik:
Something is wrong with aggregation I have other Total for different break-by when I select 2 countries and some sectors.

tsubik:
I had an error in importer and "Total excluding..." aggregated sectors from other sources (the name is the same). You can reimport and try again this CAIT error, I cannot reproduce it now.

Typo fixed.

The data for CAIT is weird, here is Algeria (Total incl is 90Mt, but Total of aggregated sectors is 590Mt :confused: )
image

tsubik:
I made "Others" as one line for expanded regions.

I have to fix "1 selected" not sure why it works like that if that is a component issue.

image

simaob:
> I had an error in importer and "Total excluding..." aggregated sectors from other sources (the name is the same). You can reimport and try again this CAIT error, I cannot reproduce it now.
> 
> Typo fixed.
> 
> The data for CAIT is weird, here is Algeria (Total incl is 90Mt, but Total of aggregated sectors is 590Mt 😕 )
> image

lovely 🤦‍♂️ well if their data is wrong they will need to fix it!

simaob:
Nice, thanks for the fixes! =)

- [x] crash on Total Excluding
- [x] others line

tsubik:
Hmm, MultiSelect component in Data Explorer (gas selector) works in the same way (showing "1 Selected" on initial load). Looks like it is not something new for multiselects, it could be improved, but maybe as a separate task to not expand and delay this PR.

tsubik:
But I found another issue :confounded: should not `World` be a region? But that could be fixed by changing `locations_grouping` csv file.

simaob:
So for the moment, World is treated as a country, but they will probably change it for a region and add all the location members. If it's easy we can do it ourselves straight away!

tsubik:
Sure, I see `World` is marked as a Region :thinking: so maybe some extra code is making it as a country? I thought that will only require changes in csv files, but deleting code is fine too.

simaob:
oh... I bet there's an hack somewhere :-o

PR #804, opened 2019-03-08
Fix/more ghg feedback

tsubik:
Fixing for GHG emission module

- EU28 is a region now
- adding reset icon for regions/gases/sectors dropdowns
- fixing not visible search input, regions dropdown
- new conflict when break by GDP, or CAPITA, only line chart is allowed
- fixing some double counting issues when region has also precalculated data (WORLD, EU28) and we have to reach for all countries data as we want also expand that region if only one selected
- making sure GDP, capita calculations are right for all regions. If WB bank does not have that data, like for World we expanding the region to all underlying countries and taking that data for calculations.
 
[PIVOTAL](https://www.pivotaltracker.com/story/show/164425169)

We have to reimport locations.
```
bundle exec rake locations:import
```

Still waiting for the data or information which one AR should we remove from PIK data source file, so emissions there are doubled until we reimport the new file.

Hopefully, all is right then :see_no_evil:

### DIFF
--- a/app/controllers/api/v1/locations/countries_controller.rb
+++ b/app/controllers/api/v1/locations/countries_controller.rb
@@ -3,10 +3,15 @@ module Api
     module Locations
       class CountriesController < ApiController
         def index
-          countries = Location.where(
-            location_type: 'COUNTRY',
-            show_in_cw: true
-          ).order(:wri_standard_name)
+          countries = Location.
+            where(
+              location_type: 'COUNTRY',
+              show_in_cw: true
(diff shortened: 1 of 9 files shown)


simaob:
Tomasz, when I run the locations import... do I need to reimport all the data as well? Are they linked by location_id or just by the ISO? (I don't remember)

tsubik:
I think I changed only location type for EU28 to `GROUP` so location import. You can always run `location_members` import, but I think I made changes there last week.

PR #819, opened 2019-04-05
Feature/ghg final feedback

tsubik:
[PIVOTAL](https://www.pivotaltracker.com/story/show/165115735)

- filter names changes
- optgroups for sector filter
- unit for per capita/GDP
- fix trim of empty data for years, leaving that in the middle (ex. PIK AFG per capita)
- table with chart data
- updating colors for GHG chart

With colors could be more "fun". Bigger samples or with text are easily distinguishable, but with our little dots and lines on the chart sometimes not so much.

fun-colors-2
fun-colors


tsubik:
I'm also not sure if we have to format the values in the table the same way as on the chart and tooltip? "2.3Gt", "10Mt" or leave it as it is as a raw number.

simaob:
One small change needed on the label of "Gasses" you're missing an **s**, on the filters dropdown.

PR #822, opened 2019-04-15
Feature/implement embed doe pie chart

kfilip94:
This PR enables embed link for pie chart component in agriculture emissions page:
[PIVOTAL](https://www.pivotaltracker.com/story/show/165120395)
uzcob-kp2g5


kfilip94:
@tsubik Good point but I think it should be changed in CW components, to something like:
`innerRadius: PropTypes.oneOfType([
  PropTypes.string,
  PropTypes.number
]),`
so I'll merge this PR now, I'll take care of this warning later

PR #823, opened 2019-04-15
Feature/new partner logo

kfilip94:
This PR adds `UNEP Collaboration Centre` partner logo to footer and partner page
[PIVOTAL](https://www.pivotaltracker.com/story/show/165344505) | [BASECAMP](https://basecamp.com/1756858/projects/13795275/todos/385560632)

**FOOTER:**
Screenshot from 2019-04-15 11 50 28

**ABOUT/CW PARTNERS:**
Screenshot from 2019-04-15 11 47 11


kfilip94:
I think we can but only on partners page. This logo is very long and the height is short, so it's hard to scale it on footer where we don't have free space

simaob:
Yeah, that makes sense. Let's make it a little wider on the About page if possible. And keep it smaller on the footer as it doesn't fit as you said.

kfilip94:
@simaob fixed :man_factory_worker: 
Screenshot from 2019-04-15 15 25 53

PR #824, opened 2019-04-16
Fix/data explorer filter sector and gases for selected data source

tsubik:
Data explorer was lacking the frontend filter of sectors and gases for the selected data source.

As the data explorer is using its own endpoints and I didn't want to mess with the logic of filtering filters which was already there or use other endpoints to get related gases and sources for the data source, so I decided to add related records ids (locations, gases, sectors) to data source data explorer metadata endpoint `api/v1/data/historical_emissions/data_sources`.

[Pivotal](https://www.pivotaltracker.com/story/show/165343045)

### DIFF
--- a/app/controllers/api/v1/data/historical_emissions/data_sources_controller.rb
+++ b/app/controllers/api/v1/data/historical_emissions/data_sources_controller.rb
@@ -4,10 +4,16 @@ module Api
       module HistoricalEmissions
         class DataSourcesController < ApiController
           def index
-            render json: ::HistoricalEmissions::DataSource.all,
+            render json: data_sources,
                    adapter: :json,
                    each_serializer: Api::V1::Data::HistoricalEmissions::DataSourceSerializer,
                    root: :data
+          end
+
+          private
+
+          def data_sources
+            Api::V1::Data::HistoricalEmissions::DataSourceWithRelatedRecordsSearch.new.call
           end
         end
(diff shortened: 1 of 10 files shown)


simaob:
I think I'm seeing a small issue with this. If I change the sector selection, and then change data-source, I think that the filtering is working, but is not reflected on the dropdown. It seems to default to "All selected", but there isn't the little dot on that option of the dropdown.

tsubik:
Ok, I will take a look. Don't know data explore the mechanism good, but I think I saw that you can set a default value for filters. I think what happens is that those defaults are only valid for CAIT source, if you change source then it is not setting the default but empty value because default one not existing for that data source.

simaob:
Yeah, that's probably it. :/

tsubik:
Going to :eyes: at that.

tsubik:
Alright, done. It wasn't a default per source thing, but I had to fix MultiSelectDropdown component to clear not the selection but the displayed label. We don't have a default selections for other sources, only first time default selection for all filters. Making the default per sources would take more effort and maybe it's not worth it.

I also changed other component we are using on that page, MultiSelect, to be the one from cw-components (without 1 selected issue and with tooltip with selected values).

simaob:
Great!=) Thanks, Tomasz. I'll review it this afternoon and merge it if all good =)

PR #835, opened 2019-04-25
Feature/ghg module download chart csv

tsubik:
- Download GHG chart's table data as CSV [story](https://www.pivotaltracker.com/story/show/165343036)
- updating cw-components to not show the Energy sector as a parent sector without any children for PIK datasource (Data Explorer)


tsubik:
Right, we encountered this at the meeting the other day. I'm not sure what we should do. The "breaking feature" when only one aggregate selected is not natural for me. Maybe if we have an option to enable/disable it we could display some message to the user or we could auto deselect the checkbox of that feature in this case. The user would have a chance to know what has happened.

simaob:
Yes that's true. Let's see what they say. Maybe something for the next phase though.

PR #836, opened 2019-04-25
Fix/my cw pages

simaob:
This PR fixes the issues that were preventing the proper use of MyClimate Watch. It does that by:

- updates format of reducers and actions files for LoginProvider;
- fixes bug with current_user controller action not exposing the json data correctly when it was in a Hash format. (not sure why, didn't have a lot of time to dig into it);

### DIFF
--- a/app/controllers/api/v1/my_cw/users_controller.rb
+++ b/app/controllers/api/v1/my_cw/users_controller.rb
@@ -4,7 +4,7 @@ module Api
       class UsersController < MyCwController
         # Gets the current user
         def current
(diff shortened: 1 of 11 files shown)


tsubik:
Great, now it works. Although, when I click on "Create a new visualization" nothing happens. I have a bunch of warnings in the console.

image

simaob:
Those warnings are there from the start I think. No idea what they mean :(

And regarding the creating visualisations, I think I might need to update some other reducers, or probably all of them. I'll have a look!

PR #845, opened 2019-05-09
Feature/tidy up agriculture importer

simaob:
This PR breaks the Agriculture Profile importer into smaller importers to make it quicker to update data, and also to see if it becomes more manageable for the server...

### DIFF
--- a/lib/tasks/agriculture_profile.rake
+++ b/lib/tasks/agriculture_profile.rake
@@ -29,7 +29,7 @@ namespace :agriculture_profile do
 
   desc 'Import all agriculture profile data from remote .csv file'
   task import: [
-    'agriculture_profile:metadata',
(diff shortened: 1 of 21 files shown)


lib/tasks/agriculture_profile.rake
```diff
@@ -1,8 +1,37 @@
...
+  desc 'Import agriculture profile country contexts from remote .csv file'
+  task import_contexts: :environment do
+    TimedLogger.log('import agriculture country contexts') do
+      ImportAgricultureContexts.new.call
+    end
+  end
+
+  desc 'Import all agriculture profile data from remote .csv file'
+  task import: [
+    'agriculture_profile:metadata',
```
santostiago: import_metadata

PR #854, opened 2019-07-03
Feature/ndcs enhancements

kevinlustig:
The WRI team has let me know they're satisfied with the state of development on the new NDC Enhancements page and would like to review it in the staging environment. Please review and let me know any issues or errors with my implementation - happy to adjust as needed to play nicely with your work. Appreciate your help getting familiar with and working with your excellent platform!

Note that this uses the FEATURE_NDCS_ENHANCEMENTS environment flag for including/excluding the relevant routes, similar to the FEATURE_AGRICULTURE flag, per advice from Simao.


app/javascript/app/components/ndcs/ndcs-enhancements-table/ndcs-enhancements-table-selectors.js
```diff
@@ -0,0 +1,111 @@
...
+export const tableGetSelectedData = createSelector(
+  [getIndicatorsParsed, getCountries],
+  (indicators, countries) => {
+    if (!indicators || !indicators.length || !indicators[0].locations)
+      return [];
+
+    const refIndicator = indicators[0];
+
+    return Object.keys(refIndicator.locations).map(iso => {
+      if (refIndicator.locations[iso].label_id !== 237) {
```
simaob: @kevinlustig : what does this `label_id` refer to, what would you expect to have on the id `237`? I'm running the code locally and the page is not showing anything, so I'm wondering if we have a data mismatch, as I've re-imported the data locally several times I am probably far away from id 237.

app/javascript/app/components/ndcs/ndcs-enhancements-viz/ndcs-enhancements-viz-selectors.js
```diff
@@ -0,0 +1,282 @@
...
+        }
+      };
+    });
+    const emissionsIndicator = indicators.find(
+      indicator => indicator.value == 'ndce_ghg'
+    );
+    for (let l in indicator.locations) {
+      const location = indicator.locations[l];
+      var type =
+        location.label_id == 235
```
simaob: As above I'm worried that hard-coding these ids might be causing the issue where I don't see any data.

app/javascript/app/components/ndcs/ndcs-enhancements-viz/ndcs-enhancements-viz-selectors.js
```diff
@@ -0,0 +1,282 @@
...
+//Table data methods
+
+export const tableGetSelectedData = createSelector(
+  [getIndicatorsParsed, getMapIndicator, getCountries],
+  (indicators, indicator, countries) => {
+    if (!indicators || !indicators.length || !indicators[0].locations)
+      return [];
+
+    return Object.keys(indicator.locations).map(iso => {
+      if (indicator.locations[iso].label_id !== 237) {
```
simaob: Another instance of this.

kevinlustig:
Hi Simao,

No worries on the timing! I just wanted to be sure I was submitting the
updated work in a way that was visible and easy to review for your team.

The hardcoded IDs are based on a misunderstanding on my part of how the IDs
get assigned. I assumed those would be consistent across implementations,
but that's obviously not the case.

My goal here is to do certain data tasks with the possible values of the
primary indicator field we're examining on this new page. We have a single
"multiple choice" data field that is powering most of what's shown on that
page. We need to auto-assign a default field value to most countries,
generate summary figures based on the different field values (e.g., there
are 10 countries with value A, 8 countries with value B, etc.), and filter
what's shown in the table based on a limited set of "acceptable" values.

I'm using the value IDs for all of these purposes. However, if the IDs are
not consistent, we'll need to use something else that is. The labels do not
seem to be a good option, as they are quite likely to change in the future
(indeed, the data labels have been changed multiple times in the course of
this work, and are about to change again.)

If there's a slug or other key that's associated with indicator *values *rather
than just the indicators themselves, I'm not sure how to retrieve that -
these appear to come through the API with only a label and an ID. They also
have an index value, but I don't think we want to tie this functionality to
legend ordering and have to refactor it just so they can reorder the legend
at some point in the future.

Any suggestions?

Thanks,
Kevin

.env.sample
```diff
@@ -1,22 +0,0 @@
...
-IPSTACK_GEOLOCATION_API_KEY=
-POSTGRES_URL=
-S3_BUCKET_NAME=
-USER_SURVEY_SPREADSHEET_URL=
-IPSTACK_GEOLOCATION_API_KEY=
-
-
-FEATURE_AGRICULTURE=
-
-
```
aabdaab: please, bring back the `.env.sample` file, you can modify it by adding the feature flag you added `FEATURE_NDCS_ENHANCEMENTS`

aabdaab:
Hey Kevin,
As Simão pointed out, you're relying on a `label_id` that is not persisted across the versions of data. If I understand correctly you're refering to these fields from the API:
image
As I'm browsing through the API `/api/v1/ndcs?category=overview`, not all of indicators.locations have got `label_id` property in the object. Was that the reason you used the ids? Could you filter by `value` from that same object?

.env.sample
kevinlustig: Will restore - this was removed in error.

kevinlustig:
> As I'm browsing through the API `/api/v1/ndcs?category=overview`, not all of indicators.locations have got `label_id` property in the object. Was that the reason you used the ids? Could you filter by `value` from that same object?

The issue I'm facing is that 'value' is likely to change. I'd rather not put the WRI team in a position where if they make a change to a value label in the future, the functionality has to be reworked.

Is there any other persistent way to identify the possible values of a field?

aabdaab:
Could you walk me through what exactly should be the result of the check you're performing with the label ids? We might need to add something on the backend but first I need to understand what is the desired output. Link to designs will also be helpful, a screenshot of what is under `237` id.

kevinlustig:
I'm doing three things with the label_ids.

For simplicity, let's say these are our values:

{
value:"No Information", label_id:x,
value:"Value A", label_id:y
value:"Value B", label_id:z
}

As I've learned, the three label_ids could be anything at all in a given
implementation, so let's ignore those.

1. If any location does *not* have Value A or Value B, assign it No
Information. Almost all locations at this point have no value for this data
set, so to save the WRI team from having to go through and add a default
value to every country, we auto-assign it here. For this, we must be able
to identify the "No Information" value in order to assign it, but we can't
rely on its label being "No Information" permanently.
2. Summarize the number of countries that have Value A and Value B. We are
counting these and displaying the totals, as well as calculating the
overall percentage of global emissions represented by these countries
(using another field.) For this, we must be able to identify the "Value A"
and "Value B" values, but we can't rely on their labels being "Value A" and
"Value B" permanently.
3. If any location does *not* have a value of No Information, show it in
the table. Countries with No Information are relevant and need to be
displayed on the map - however, those countries are irrelevant for the
table as they do not have any other relevant data to display. The needs are
the same as #1.

The tricky part is that Value A might at some point become (these are
totally made up) "Property A" or "Country Has A" or even "Value C." We
cannot rely on all or part of that value to remain the same. It will be
functionally the same - that is, it should appear on the map and in the
summary data in essentially the same way - but its label may change.

One possibility seems to be to generate a slug when the values are first
created that persists even if the label is changed. So, Value A has a
key/slug of "value_a," and that remains "value_a" even if the label is
changed to "Property A," "Value C," etc. However, I have no idea if this is
feasible given your data methodology. Obviously, field values would have to
have some kind of persistent identity in the data structure, and I don't
know if they even do, or if they're generated from the field data/metadata
every time a data update is made.

Another option is to just tell WRI they can't change these labels once the
functionality is live. This is suboptimal for obvious reasons, but the LOE
to address this issue may not be worth it.

aabdaab:
Hm, can you filter the result by the indicator id? There's no endpoint like that yet but I see that in our DB there's an association between labels and indicators, we could have an endpoint exposing data like so:
```
{
data: [
{
id: 234,
indicator_id: 241,
value: "Signed Agreement"
},
{
id: 235,
indicator_id: 241,
value: "Intent to withdraw"
},
{
id: 236,
indicator_id: 241,
value: "Joined Agreement"
},
{
id: 237,
indicator_id: 242,
value: "First NDC Submitted"
},
{
id: 238,
indicator_id: 242,
value: "INDC Submitted"
},
{
id: 239,
indicator_id: 242,
value: "No Document Submitted"
},
{
id: 240,
indicator_id: 243,
value: "Yes"
},
{
id: 241,
indicator_id: 243,
value: "No"
},
{
id: 242,
indicator_id: 243,
value: "No NDC submited yet"
},
{
id: 243,
indicator_id: 271,
value: "Single-year target"
},
{
id: 244,
indicator_id: 271,
value: "Multi-year target"
},
{
id: 245,
indicator_id: 271,
value: "No Document Submitted"
}
```
It would complicate things a bit, cause indicator ids also cannot be relied on, so you'd have to find their id by name or so. Would that be helpful?

kevinlustig:
The issue is that I need to target specific values from the indicator. It
matters in our implementation whether a location has (to use your example)
a value of "Single-year target" or "Multi-year target". That's what I'm
attempting to filter on, not the indicator.

aabdaab:
Yes, I can see the issue there. I'm afraid we've got two solutions here, both not ideal. Because the data is being managed by WRI team, having an additional field, like a slug or something, would have to be maintained by them. Here's a sneak peak at the current csv file which we're importing to make associations for the labels in the database:
image
As you can see, we've got only two columns here. We could add a third one, in our database, `slug` generated based on the current label values but that would also mean, we'd have to somehow persist it across all of the data imports, and if the label values change as well as ids, we hit the same issue as you did; how can we determine to which label object should we assign that slug?
So, the only solution I see here, would be having the slug column in the file itself that would have to be maintained by WRI team.
OR rely on the labels not changing; although you indicated that these changed a few times already. I guess this has to be discussed with WRI to reach an agreement, how can we safely depend on that labels: label values or adding a slug maintained by them.

kevinlustig:
Yeah, I was wondering if we were in that situation.

Personally, I'm in favor of an optional third column wherein they can
provide a key/slug for values that would give them the flexibility to
change labels while preserving functionality based on field values -
presumably this could be left blank for almost all of the fields that are
currently in the system. However, I take it this has not come up yet across
Climate Watch, so it's obviously a niche need. Moreover, it's your
platform, so I defer to you on this question.

As you note, if we don't add this feature, they will need to leave the
labels as-is going forward or else someone will need to rework the code
when they want to make a change. I'm happy to communicate this to them or
you can, as you would prefer.

aabdaab:
I think having a third `slug` column with an optional value would be a great idea. Like you said, it's not pretty nor safe to depend on the labels and it is a limitation for them if they cannot change it. I'd say, have a go and talk with WRI and if they agree on that solution, we can change our importer to count the new `slug` column. 🙌

kevinlustig:
Okay, discussing with the WRI team.

kevinlustig:
I got the all clear on this from the WRI team. Can you send them the metadata spreadsheet in the format you'd like them to use (e.g., with the additional column) as well as any requirements for the slug format? They can fill in the needed values and reupload the spreadsheet whenever you are ready for it.

aabdaab:
Hey, Kevin, I just send them a message on Basecamp and there's a #858  open with acknowledging that slug in the dataset importers along with the endpoints. I'll let you know once I hear from them.

aabdaab:
Hey Kevin, is it ready for re-review?

kevinlustig:
WRI has requested additional changes, so we're on hold for the moment.
(However, the new slug handling seems to be working very well.)

I will cancel this PR and issue another one when the additional updates are
ready.

PR #863, opened 2019-08-16
Fix Indc::ValueSerializer label_slug attribute

aabdaab:
Check with `/api/v1/ndcs?category=overview&location=ARM` endpoint for example.

image


kevinlustig:
Looks good, Marta - direct check of the API endpoint works for all countries, specific countries, and as part of loading the NDCs page. Think this is fixed.

aabdaab:
Great, I'll merge it then so you can rebase with develop!

PR #864, opened 2019-09-02
Fix negative values for ghg emissions OTHERS

aabdaab:
This PR tries to solve the issue of having negative values in GHG emissions, 
`Others` column: https://basecamp.com/1756858/projects/13795275/todos/396887203


aabdaab:
@tsubik If you could take a look at it, as I saw you were working on a different bug regarding that `Others` column

tsubik:
Hey @aabdaab I found what is the problem with `Others` column value. There could be many data points for the expanded region, like on the example multiple different gases selected and the problem is that it takes only one with `find` method :/. Take a look at this PR https://github.com/Vizzuality/climate-watch/pull/866

PR #866, opened 2019-09-02
expanded region data can have multiple data points for different sect…

tsubik:


tsubik:
Marking it as WIP, need to properly test it tomorrow morning.

tsubik:
Ok, I think it works fine now.

PR #869, opened 2019-09-11
Feature/ndcs enhancements

kevinlustig:
After a number of additional changes, as well as remediating issues we previously discussed with data labels/slugs, this is now ready for review and deployment to staging again. 

One major change is that there are now two new pages rather than one that are very similar, but have important and specific differences that make it somewhat tricky to combine them. All things being equal, some kind of parent component would make sense for handling the common aspects of both. But, as this whole section of the site (NDCs) is in the process of being wiped out and remade anyway and these pages are meant as a low-budget stop-gap, I have taken the path of least resistance here. Let me know any issues with my approach. 

Note that this now uses the FEATURE_NDCS_ENHANCEMENTS and FEATURE_NDCS_LTS feature flags.


aabdaab:
checking ☝️

aabdaab:
Hi Kevin!
Unfortunately, the new options on the menu are not working for me. Both `2020 NDC TRACKER` and `LONG-TERM STRATEGY TRACKER` are causing a page crash.
image

These are the steps I took:
- `rake db:migrate`
- `rake db:import`
- set the flags in `.env`
```
FEATURE_NDCS_ENHANCEMENTS=true
FEATURE_NDCS_LTS=true
```
- `yarn start`

Am I missing something?

aabdaab:
Something to change as well; when we don't have the flags set to true, don't display empty options:
image

kevinlustig:
> Hi Kevin!
> Unfortunately, the new options on the menu are not working for me. Both `2020 NDC TRACKER` and `LONG-TERM STRATEGY TRACKER` are causing a page crash.
> image

Marta, I encountered this issue as well - it has to do with the CW Components library. CW Components requires a higher version of React to function properly (16.4), but the main CW project allows an install at React 16.2. React 16.2 can't handle "createRef" in the CW Components table component. See, e.g., https://github.com/highcharts/highcharts-react/issues/84. If you upgrade your React version, the error should go away. It might also be worthwhile to update the package file for the overall CW project, but I didn't want to mess with any global aspects of the project.

kevinlustig:
> Something to change as well; when we don't have the flags set to true, don't display empty options:
> image

Good catch, thanks. Fixed.

kevinlustig:
Thanks, Marta! Good call on the other approach for the menu items. Much
more generalized. I just went with quick and dirty.

FYI, WRI has a couple of further outstanding tweaks they want me to make
and they are going to do a review of this once it's on staging and will
probably have feedback. So, I will be back for a further deployment
probably in the next few days (which will hopefully be the last one!)

simaob:
Added env vars to the Jenkinsfile and will now merge and deploy to staging, should be there in the next 30 minutes.

PR #891, opened 2019-11-15
Feature/lts map styling

Bluesmile82:
- Style map buttons
- Text position

Extra:
- Name z-index in settings so the list stops growing
- Fix title of buttons when disabled

image

### DIFF
--- a/app/javascript/app/components/cookie-consent/cookie-consent-styles.scss
+++ b/app/javascript/app/components/cookie-consent/cookie-consent-styles.scss
@@ -3,13 +3,13 @@
 .container {
   position: fixed;
   bottom: 0%;
-  z-index: 15;
+  z-index: $z-index-cookie-consent;
   width: 100%;
   padding: 10px 15px;
-  background-color: #0A1941;
+  background-color: #0a1941;
   font-family: $font-family-1;
   font-size: 14px;
-  color: #F4F5F6;
+  color: #f4f5f6;
   display: flex;
   flex-direction: column;
   justify-content: center;
@@ -21,16 +21,16 @@
 }
 
 .link {
-  color: #F4F5F6;
+  color: #f4f5f6;
   text-decoration: underline;
 }
 
 .button {
   margin-left: 5px;
   width: 106px;
-  background-color: #FFC735;
+  background-color: #ffc735;
   text-transform: none;
-  color: #0A1941;
+  color: #0a1941;
   margin-top: 15px;
   cursor: pointer;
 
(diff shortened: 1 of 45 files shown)


app/javascript/app/components/cookie-consent/cookie-consent-styles.scss
```diff
@@ -21,16 +21,16 @@
...
-  color: #F4F5F6;
+  color: #f4f5f6;
   text-decoration: underline;
 }
 
 .button {
   margin-left: 5px;
   width: 106px;
-  background-color: #FFC735;
+  background-color: #ffc735;
```
kfilip94: Please use `$theme-secondary` constant here from `settigns.scss`

app/javascript/app/components/cookie-consent/cookie-consent-styles.scss
```diff
@@ -21,16 +21,16 @@
...
   text-transform: none;
-  color: #0A1941;
+  color: #0a1941;
```
kfilip94: Please use `$theme-color-intense` constant here from settigns.scss

app/javascript/app/components/cookie-consent/cookie-consent-styles.scss
```diff
@@ -3,13 +3,13 @@
 .container {
   position: fixed;
   bottom: 0%;
-  z-index: 15;
+  z-index: $z-index-cookie-consent;
   width: 100%;
   padding: 10px 15px;
-  background-color: #0A1941;
+  background-color: #0a1941;
```
kfilip94: Please use $theme-color-intense constant here from settigns.scss

app/javascript/app/components/cookie-consent/cookie-consent-styles.scss
```diff
@@ -3,13 +3,13 @@
...
   font-family: $font-family-1;
   font-size: 14px;
-  color: #F4F5F6;
+  color: #f4f5f6;
```
kfilip94: Please add this color to settings.scss or maybe replace it with a similar one that we already have, maybe `$light-gray`?

kfilip94:
Small issue with a z-index when share menu is expanded:
Screenshot from 2019-11-18 10 28 19

kfilip94:
The bottom edge of South America is clipped and the helper text overlaps the map, is this the desired behavior?
mim3q-kij33

Also, I noticed that there are a few significant differences between this page and [the design](https://projects.invisionapp.com/d/main#/console/18371781/381684670/preview) - are we planning to align them later, in the next iterations of this component?

Bluesmile82:
We cannot really put the text next to the buttons if it's not absolute but I can add some background to it so it doesn't overlap. The map position needs to be aligned probably after we add all the different elements. 

We have been discussing the design differences. The problem is that if we replicate the design as it is we will have to change the theme in all the site and we are going to wait to do this. We will have to discuss it with the client

kfilip94:
@Bluesmile82 nice job :+1:  I left a few comments above :point_up_2: 
I'm not sure if all of them are tied with this PR, but it'd be good to fix them here or add them to Pivotal

PR #892, opened 2019-11-15
Add progress component. Add legend with data

Bluesmile82:
- Add progress component
- Add legend with data and styles

image


kfilip94:
@Bluesmile82 It looks great, nice job with these progress bars :+1: 
The only thing that could be improved IMO is the padding in these two pie-charts:
Screenshot from 2019-11-18 12 02 09

Each side has different padding, also this `0 %` it's not in the middle of the circle. If you could play with position of this chart and text and with the padding, it would be great :+1:

Bluesmile82:
Thanks! I really didn't pay attention to these two first cards. They will be done in the next PR

PR #894, opened 2019-11-21
Feature/lts first cards

Bluesmile82:
- Add the first two cards with data
- Add tooltip and correct the position so it always lets the center label show
- Fix map height
- Add hand cursor icon

Known bugs: 
- For finance: the last two indicators: the donut is not showing. This is because of a mismatch between indicators names. There is a task for that already in PT

Note: The data for the emissions is coming from the emissions data that we have for the last year on CAIT, ALL GHG and Including 

image


simaob:
having a look!

app/javascript/app/components/ndcs/lts-explore-map/donut-tooltip/donut-tooltip.jsx
```diff
@@ -0,0 +1,41 @@
...
+  const chartTop = chartReference.getBoundingClientRect().top;
+  const referenceTop = reference.getBoundingClientRect().top;
+  let top = chartTop - referenceTop + coordinate.y;
+
+  // Avoid covering the label on the center
+  const left = coordinate.x > 40 ? coordinate.x + 80 : coordinate.x;
+  if (top < 340 && left < 190) top += 80;
+  return ReactDOM.createPortal(
+    <div className={styles.tooltip} style={{ left, top }}>
+      {`Parties with ${legendItemName} represent ${percentage} % of global emissions`}
```
kfilip94: A small detail, please remove the space between a number and %:
```suggestion
      {`Parties with ${legendItemName} represent ${percentage}% of global emissions`}
```

kfilip94:
There is a blank card where there is no data: in `Finance and support/Conditionality of adaptation finance` and  `Finance and support/Conditionality of mitigation finance`:
Screenshot from 2019-11-22 16 33 39

Maybe we should add an `no data` info message, or hide this wrapper completely?

app/javascript/app/components/ndcs/lts-explore-map/lts-explore-map-selectors.js
```diff
@@ -198,81 +168,88 @@ export const getLegend = createSelector(
...
-          opts: {
-            color: getColorByIndex(indicator.legendBuckets, label.index),
-            suffix: '%',
-            label:
-              'of global emissions represented by these countries (2014 emissions data)'
-          }
+export const getEmissionsCardData = createSelector(
+  [getLegend, getMapIndicator, getMapIndicator, getIndicatorsData],
+  (legend, indicator, selectedIndicator, indicators) => {
+    if (!indicator || !legend || !selectedIndicator) {
```
kfilip94:
```suggestion
    if (!indicator || !legend || !selectedIndicator || !indicators) {
```

Bluesmile82:
The blank chart is known bug in the description. I think we should have data for all indicators. Actually these are not the final LTS indicators and the problem is more a mismatch between the names in different places so I wont remove the chart if it doesn't have data right now. Lets wait for the final data

kfilip94:
I read this section and I forgot about it, sorry for that :disappointed:

PR #900, opened 2019-11-25
Feature/contained links

Bluesmile82:
This PR changes every ```/contained``` link from the NDCs partnership to open them in new tabs. This includes:

- NDC Linkages - Links in map
- NDC Content - Links in map
- NDC Content - Links in table
- NDC Content - Search dropdown link

Extra:

Fixes dropdown search button

image


simaob:
Hey Álvaro, is this not up to date with the footer changes?
I noticed that the "Download links" on the toolbars isn't opening on a new page, that would another one to open like that! Thanks!

Bluesmile82:
You are right! Just rebased and added the download link target

simaob:
> You are right! Just rebased and added the download link target

and also to update to link to the full route, instead of the contained!

Bluesmile82:
Ready again!

PR #901, opened 2019-11-25
Use only country name for tooltip

Bluesmile82:
There is no design for the tooltip but it was decided it should be only the country name.

image


kfilip94:
On desktop, tooltip looks and works nicely:+1: There is a typical problem with ReactTooltip library on mobiles - after clicking the tooltip doesn't disappear, so the best would be to disable it for mobiles.

Also, there is another glitch on mobiles: share buttons and search input overlap, I know that this is not connected with this PR but I shouldn't be hard to fix.
puwsw-whcl7

After that it'll be ready to merge ;)

Bluesmile82:
Thanks! For the mobile problem, we have a delayHide attribute that was in 3000 ms. I changed it to 1000. I also fixed the styles

image

PR #903, opened 2019-11-26
Feature/ndcs explore page

Bluesmile82:
Create new NDCS explore page:
- Copy the code of the LTS explore page adding the data from the NDCS indicators

Note: We are missing some indicators as you can see on the table. Also as we don't have the emissions indicator the donut chart is not showing. We will have to fix the slugs when we get those indicators from the API (After data import)

Also, the first card information should come from the backend as in the LTS page.

Extra:
- Fix LTS header title and description
- Change menu names as in the design

image

image

image


app/javascript/app/components/ndcs/ndcs-explore-map/donut-tooltip/donut-tooltip.jsx
```diff
@@ -0,0 +1,41 @@
...
+    </div>,
+    reference
+  );
+};
+
+DonutTooltip.propTypes = {
+  content: PropTypes.object
+};
+
+export default DonutTooltip;
```
kfilip94: Here you're duplicating `donut-tooltip` (`ndcs-explore-map` and `lts-explore-map`), and it seems they are exactly the same. Please use only one source of this component, move it directly to the `components` library or `components/ndcs`

kfilip94:
Hey, @Bluesmile82 could you clarify this PR a bit? I see you just copied & pasted a lot of files but in many cases, they are the same, the only difference is with the data that you feed them. Do we have a good reason for not splitting them into reusable parts?

Bluesmile82:
About the copying instead of reusing: I agree that there are components as the donut that should be reused. I will change that. But I'm not very sure if we will have to change some other things so first copying the code and then, when we are sure, reusing the parts I think will be the best choice.

Bluesmile82:
I made some changes:
- Reuse share components (donut tooltip, ...)
- New version of CW components and table theme

There is a bug in the dynamicRowsHeight height for the cell. Its in PT

kfilip94:
@Bluesmile82 thanks for your latest changes, good job with this :+1: Could you please explain what do you mean saying `I'm not very sure if we will have to change some other things` - it's because the design is not finally defined or there are "holes" in the data set? I understand your point and if some parts are not defined and they will change soon, I agree to merge it, but as a reviewer, I just need to understand the reason better ;)

Bluesmile82:
Sorry for being so obscure 😆 Is just about the data. I'm not sure how the first chart is going to work in each page and if they are going to be more differences on the table. I prefer to abstract it later. Also as they are two different pages It might not be the best choice to reuse too much code in my experience

PR #905, opened 2019-11-29
Feature/ndc country changes

Bluesmile82:
- Change dropdown title to View full text
- Change button text and position: Fix for tablet and mobile also with and without search
- Add back button: Back button text was a bit difficult to get as react-router doesn't provide a previous location prop. So the process is the following:

1. In app.js store the previous pathname and search into the sessionStorage. We don't want it cleared on reload.
2. Retrieve this data on the ndcs country and assign the label depending on the pathname with some exceptions filtered by a regex array.
3. If the previous location is one directed from the same page, an error or similar, clear the location and dont show the back button.

Extra:

- Fix Lybia crashing on country compare: http://localhost:3000/countries/compare?locations=LBY

image


kfilip94:
:eyes: checking

kfilip94:
There is something wrong with back button :( When we click at the country, then we will pick NDC/INDC from the dropdown and then we click back in the browser, the `Back to LTS Explore` is not there and dropdown is out of the page layout:
czr0d-g7cv0

Bluesmile82:
This back button has some limitations. The problem is if you come from a page that is directly linked from the ndc country page. You should get the previous back link. You are right about the dropdown. Its out of the grid. Will check that

app/javascript/app/pages/ndc-country/ndc-country-component.jsx
```diff
@@ -13,26 +15,117 @@ import Dropdown from 'components/dropdown';
...
+      return true;
+    }
+    return false;
+  });
+  return lastPathLabel || (updatedPathname && toStartCase(updatedPathname));
+};
+
+const shouldClearPath = pathname => {
+  const clearRegexps = [/\/ndcs\/country/, /\/error-page/, /\/ndcs\/compare/];
+  if (clearRegexps.some(r => pathname.match(r))) {
```
kfilip94: This bit throws an error after copying & pasting the country link:
ah23u-x5xqh
image

app/javascript/app/app.jsx
```diff
@@ -1,11 +1,12 @@
 import 'babel-polyfill';
 
-import React from 'react';
+import React, { useEffect, useState } from 'react';
```
kfilip94: Hooks! :heart_eyes:

kfilip94:
Hey @Bluesmile82, it looks great to me, good job with this `Back` button :+1: Check my comments above and let me know when you'll apply these changes ;)

Bluesmile82:
This PR is now added to `ndc-country-changes` branch

PR #908, opened 2019-12-04
Fix/lts explore table

Bluesmile82:
- This PR fixes the height and width of the columns on the table
- Removes fetching from the table as we already call it on the map and page.
- Fixes the fetching in LTS. It was failing when we came from NDCs explore directly to LTS explore

image


kfilip94:
checking :eyes:

kfilip94:
Hey, `Explore Long-Term Strategies (LTS)` page doesn't show any data, is there something new that I should import to the database or it's a bug?
Screenshot from 2019-12-05 09 37 12

PR #915, opened 2019-12-10
Feature/lts first card

Bluesmile82:
This PR fixes several feedback problems:

- 'Counter' card (First card in LTS explorer) : Now is showing LTS document submission count always
- Updates NDCs explore text on top of the map
- Sort and fix legends and donut chart for LTS and NDC explorer (Now we have a 'Not applicable' or 'No information' at the bottom that has the count of countries without data). There are still a couple of little bugs with the donut chart that are in a new bug task on PT.
- Default for LTS and NDCs categories and indicators
- Apply new data
- Links to LTS country from table
- Filter sources of NDC explore

To try: Remember to run:` bundle exec rails indc:import`

Note: Some selectors have been duplicated as they will produce more bugs in the way we are using selectors. We may do a refactor in the future to pass only the state instead of individual properties.

image

### DIFF
--- a/app/javascript/app/components/ndcs/lts-explore-map/lts-explore-map-selectors.js
+++ b/app/javascript/app/components/ndcs/lts-explore-map/lts-explore-map-selectors.js
@@ -215,24 +215,27 @@ export const getEmissionsCardData = createSelector(
     if (!emissionsIndicator) return null;
 
     const emissionPercentages = emissionsIndicator.locations;
-    const totalEmissions = Object.values(emissionPercentages).reduce(
-      (a, b) => a + parseFloat(b.value),
-      0
-    );
+    let summedPercentage = 0;
     const data = legend.map(legendItem => {
       let legendItemValue = 0;
       Object.entries(selectedIndicator.locations).forEach(entry => {
-        const [locationIso, { value: legendItemName }] = entry;
+        const [locationIso, { label_id: labelId }] = entry;
         if (
-          legendItemName === legendItem.name &&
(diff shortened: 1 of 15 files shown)


simaob:
I think I see a small issue on the ndcs-explore. -4%?
Screenshot 2019-12-10 at 20 08 27

kfilip94:
:eyes: checking

Bluesmile82:
I'm working on the table issue. Not trivial. I will go through the NDCs calculation too but maybe we can do it in a different PR as the priority is LTS.

About the copying of the code: I think we can refactor and abstract some code. But something I learned is that abstracting too early is an anti-pattern. It is not only about the speed of development but these are independent pages and they can have different particularities so abstracting all of them could be counterproductive. I agree that there are some parts that could be abstracted. But premature abstraction is a danger:

https://twitter.com/_ericelliott/status/1150100101928509440?lang=eu

kfilip94:
@Bluesmile82 I agree with you, but we're copying the whole page with the same styles and often selectors, not a single method, I'm worrying that refactor never going to happen, and we'll end up here: https://twitter.com/watkinseandre/status/1150170031944273922?s=20 :D

In terms of `Explore NDCS` calculation bug, if @simaob is fine with merging this as it is, do it :wink:

Bluesmile82:
I will review it and try to find a middle ground. I'm going to do a new PR in cw-components. I'll let you know

kfilip94:
Thank you :+1:

Bluesmile82:
Ready to merge!

PR #916, opened 2019-12-12
Feature/lts country page summary

kfilip94:
This PR implements two features: new styles for data cards, aligned with the new design and a Summary page for LTS-country page. Keep in mind that the data wasn't available for all the slugs.
[PIVOTAL - summary page](https://www.pivotaltracker.com/story/show/170209756) | [PIVOTAL - new  styles for cards](https://www.pivotaltracker.com/story/show/170005132)

65hsx-cyk7k

- I extracted the `NdcCountryAccordionProvider` to the `providers` folder as a separate provider, in the next PR I can use this version in `NDCS-country-accordion` component

PR #919, opened 2019-12-17
Feature/move card theme to styles folder

kfilip94:
This tiny PR moves the new theme for card component to `styles/themes/card` folder, according to the [@Bluesmile82 comment](https://github.com/Vizzuality/climate-watch/pull/916#discussion_r357559996). It also adds the box-shadow to the card-light theme, as it is done on the design.
[PIVOTAL](https://www.pivotaltracker.com/story/show/170005132)

### DIFF
--- a/app/javascript/app/styles/themes/card/card-light.scss
+++ b/app/javascript/app/styles/themes/card/card-light.scss
@@ -1,0 +1,25 @@
+@import '~styles/settings.scss';
+
+.card {
+  color: "pink";
+  border-radius: 6px;
+  border-color: transparent;
+  box-shadow: 0 2px 20px 0 #E8ECF5;
+  margin: 0.9375rem;
+}
+
+.contentContainer {
+  background-color: #ecf1fa;
+  display: flex;
+  align-items: center;
+  padding: 0 30px;
+  min-height: 70px;
+}
+
+.title {
+  font-weight: $font-weight;
+}
+
+.data {
+  background-color: white;
+}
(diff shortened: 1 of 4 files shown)


app/javascript/app/styles/themes/card/card-light.scss
```diff
@@ -4,8 +4,8 @@
   color: "pink";
   border-radius: 6px;
   border-color: transparent;
-  box-shadow: 0 2px 20px 0 #E8ECF5;
-  margin: 0.9375rem;
+  box-shadow: 0 2px 20px 0 #e8ecf5;
```
Bluesmile82: It would be nice to use some color variable that we have on settings or a box-shadow variable. Maybe this is a new box-shadow standard so we can add a variable for this

kfilip94: Hey @Bluesmile82, I added two new colors to `settings.scss` file, I think for now there is no need for adding `box-shadow` as a constant because it seems to be the only place that we're going to use that :)

Bluesmile82: So I think I prefer adding a box-shadow constant than new colors. I don't know if this colors are used somewhere else. This box-shadow will probably be used somewhere else when we change the designs

Bluesmile82: Maybe we need the color constants too. I'm just afraid of too many colors in the settings. I have Polychromofobia

kfilip94: I didn't find this box-shadow on Invision in other places, nor in the project. Do you think it's still worth it to put it there?

Bluesmile82: Ok, let's leave it as it is then

PR #920, opened 2019-12-18
Fix search filtering order

Bluesmile82:
This PR fixes the problem with the search in LTS - NDCs Explorer and LTS tracker
The order of filtering was not correct on the selectors so some data that was not shown was taken into account on the search.

Note: There are still some cases on some links that the href matches with the search but the text doesn't. I don't think we can do much about this as we are receiving the HTML directly on this case. We could parse it but that would be a different task

image

### DIFF
--- a/app/javascript/app/components/ndcs/lts-explore-table/lts-explore-table-selectors.js
+++ b/app/javascript/app/components/ndcs/lts-explore-table/lts-explore-table-selectors.js
@@ -8,7 +8,7 @@ import { getMapIndicator } from 'components/ndcs/lts-explore-map/lts-explore-map
 const getCountries = state => state.countries || null;
 const getCategories = state => state.categories || null;
 const getIndicatorsData = state => state.indicators || null;
-const getQuery = state => deburrUpper(state.query) || '';
+export const getQuery = state => deburrUpper(state.query) || '';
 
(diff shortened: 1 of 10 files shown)


app/javascript/app/components/ndcs/lts-explore-table/lts-explore-table-selectors.js
```diff
@@ -167,3 +151,19 @@ export const getFilteredData = createSelector(
...
+  (data, query) => {
+    if (!data || isEmpty(data)) return null;
+    return data.filter(d => {
+      let match = false;
+      Object.keys(d).forEach(col => {
+        if (deburrUpper(d[col]).indexOf(query) > -1) {
+          match = true;
+        }
+      });
+      return match;
```
kfilip94: We're repeating this in 3 places, do you think we can extract this bit and move it to the `utils` file as a filter-by-query function? Or entire selector if possible :)

Bluesmile82: Yeah, we cant extract the selector but I will create an utils function

kfilip94: Thanks :+1:

Bluesmile82:
Ready again!

PR #922, opened 2019-12-19
Add new card's style to the NDC/Country page

kfilip94:
This PR adds new card styles to NDC/country page. We didn't modify slugs because we are waiting for the client's feedback.
[PIVOTAL](https://www.pivotaltracker.com/story/show/170334122)

image

- I extracted the new card row component from LTS to `CardRowLight` component, let me know if this makes sense ;)

- I copied `card-light` theme from develop, because rebasing only for that was too complicated, we will resolve this once this PR will be merged with `new-ndc-country` branch.


Bluesmile82:
On the design, the separator lines for the card content go all the way through

image

Bluesmile82:
I think we are duplicating code here. I am not sure if we need this CardLight component. It doesn't have any special functionality. Also, even if we have it because we are going to use it a lot then why do we need the cardLight theme? Why don't you just use the styles?

kfilip94:
`cardLight` theme is for `Card` component, not for `CardRow`/`CardRowLight` component. I didn't want to mess up with other places in the app that use old Card and CardRow. But it makes sense revisiting all places wehere we're using Card component, and unifying this into one.

PR #923, opened 2019-12-19
Feature/country dropdown

Bluesmile82:
- Add Country dropdown to NDC page

Note: The styles include some !important 😓 Because of the react-selectize limits

image


kfilip94:
- After changing the country, the `View full text` dropdown and the `View Document` button change the width from 100% to the desired one, it'd be good to fix it.
 oki1i-p9yf4

- When other than the `Summary` tab is selected, changing the country removes the last node from url and navigates back to the `Summary` tab.
qkok8-fyjyu

- Should we add this dropdown to the LTS country page as well?

kfilip94:
Hey @Bluesmile82, excellent job with this dropdown, it works nicely, the search feature is awesome :+1:
I found two small issues, check my comments above ;)

Bluesmile82:
I think the jump will be fixed when we apply the `feature/lts-country-header` branch changes

PR #924, opened 2019-12-26
Feature/share button

Bluesmile82:
This PR gets the share button out of the ButtonGroup and opens a modal instead of a dropdown with the same options.

Note: There is another task to change the rest of the site share buttons.

image

### DIFF
--- a/app/javascript/app/components/button/share-button/share-button-styles.scss
+++ b/app/javascript/app/components/button/share-button/share-button-styles.scss
@@ -18,7 +18,8 @@
 }
 
 .modal {
-  height: initial;
+  overflow: hidden;
+  height: 90%;
 
   @media #{$tablet-portrait} {
     height: 210px;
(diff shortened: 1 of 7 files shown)


app/javascript/app/components/button/share-button/share-button-styles.scss
```diff
@@ -0,0 +1,99 @@
...
+
+  &:hover {
+    box-shadow: none;
+    fill: $theme-color;
+  }
+}
+
+.modal {
+  height: initial;
+
```
kfilip94: Maybe we can set `width: 90%` or `margin: 0 20px` for mobiles? Because now it has 100% width:
Screenshot from 2019-12-27 10-55-53

kfilip94:
Hey, could you hide these scrollbars inside the share modal? It appears on Chrome, Ubuntu distro in my case, and adding `overflow: hidden;` solves the problem 
Screenshot from 2019-12-27 10-05-58

kfilip94:
Hey @Bluesmile82, great job! After solving these two issues and solving conflicts with develop you can merge it :wink:

Bluesmile82:
Hey! thanks for the overflow tip. I couldn't see that one. The other modals expand to the total height and width on mobile except for the nav menu. I just replicated that one.

PR #928, opened 2019-12-27
Feature/lts country header

Bluesmile82:
This PR adapts the LTS country header to the new styles: Adds the back button styles the whole header and the remaining buttons.

- The back button logic has been abstracted to a utils file.
- When we don't have a clear previous page we just show .''Back". This was preferred by the design team.
- The links in LTS are disabled as we still need to create those sections

Extra:

- Added NDC country page changes on cards and slugs: Some data is missing. There will be a new task created for this

image

### DIFF
--- a/app/javascript/app/components/button/button-component.jsx
+++ b/app/javascript/app/components/button/button-component.jsx
@@ -46,7 +46,7 @@ const Button = props => {
     <NavLink
       className={classNames}
       to={link}
-      onClick={onClick}
+      onClick={disabled ? e => e.preventDefault() : onClick}
       target={target}
       {...tooltipProps}
     >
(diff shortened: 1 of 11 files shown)


kfilip94:
:eyes: checking

kfilip94:
Hey @Bluesmile82, great job with these changes :1st_place_medal:  Please check below my comments, I found two issues in back button feature:

- After copying and pasting an URL to the `lts-explore` page, clicking on a country and then going back breaks the URL: `http://localhost:3000/lts/country/null`
ibraw-op0f2

- When switching between `Explore LTS` and `Explore NDCS` pages, sometimes Back button refers to a different page (`Back to LTS Explore` refers to `Explore NDCS` and 
vice versa):
ggcvy-le3x1


- Shouldn't be the space between the `Compare countries and submissions` button and the tabs bigger, as it is on the design?
image

Bluesmile82:
Changes made! Lots of thanks for the comments. We should test the back button a bit more on the release day.

PR #930, opened 2020-01-07
Feature/overview question card

Bluesmile82:
This PR starts the question card with the progress bar, answer taken from the indicator and GHG percentage calculation.

Missing:
- Info button modal: We need to know what info should be in here
- Last indicators data: We need to store this data somewhere in the backend?

image


app/javascript/app/components/ndcs/ndcs-commitments/ndcs-commitments.js
```diff
@@ -1,7 +1,22 @@
...
-const NdcsCommitments = () => <Component data={commitmentsData} />;
+class NdcsCommitments extends PureComponent {
+  componentWillMount() {
+    this.props.fetchNDCS(true);
+  }
+
+  render() {
+    return <Component data={commitmentsData} />;
+  }
+}
```
kfilip94: Hey, what do you think about starting using hooks? Small components like this are perfect for that ;)

app/javascript/app/components/ndcs/ndcs-commitments/question-card/question-card-selectors.js
```diff
@@ -0,0 +1,53 @@
...
+  (indicators, positiveAnswerIsos) => {
+    if (!indicators || !positiveAnswerIsos) {
+      return null;
+    }
+    const emissionsIndicator = indicators.find(i => i.slug === 'ndce_ghg');
+    if (!emissionsIndicator) return null;
+
+    const emissionPercentages = emissionsIndicator.locations;
+    let summedPercentage = 0;
+    positiveAnswerIsos.forEach(iso => {
```
kfilip94: More a suggestion, but I'd replace `forEach` with `reduce` function here:
```suggestion
    const summedPercentage = positiveAnswerIsos.reduce((acc, iso) => {
      return emissionPercentages[iso] ? acc + parseFloat(emissionPercentages[iso].value) : acc;
    }, 0);
```

kfilip94:
Hey @Bluesmile82, great job! please check my comments below:
- The percentage value should be bolded, check the design:
image
- Could you check this percentage value (0%) and how is it formatted? Probably it's very small, but we should show it somehow:
image
- The colors of the progress bars in sections two and three are the same, but on the design they have uniqe colors for each section:
image

Are we leaving the mobile version for later, next iteration?

Bluesmile82:
Changes made! The Marshall Islands are actually not very representative in the global amount 😅 

image

PR #931, opened 2020-01-07
Feature/overview share button

Bluesmile82:
This PR creates the embed pages for the 3 different overview sections
Also, it changes the names to NDC overview section and uses hooks.

Extra:
- Remove duplicated slugs in LTS

Try: Use the share buttons on the overview page

image


kfilip94:
A couple of things that I'm struggling with. "Copy link" option shouldn't copy the link to the current URL? Because now it copies the embed link which is misleading. Also, I think we should include `Copy embed code` and `Copy embed link` on the overview page share modal somehow.
**Overview page share modal**
image
There is also a layout problem in the embed version of the share modal 
**Embed section share modal**
image

Rest of the PR looks good & clear to me, so I'd like to clarify these issues above :point_up:

Bluesmile82:
Thanks for the comments! The share layout is fixed now:
image

About the text on the button. For me, it makes sense to share only the part of the page that has the share button, therefore, the embedded link. We can copy the URL from the browser. @Fanny what do you think?

fannycc:
> Thanks for the comments! The share layout is fixed now:
> image
> 
> About the text on the button. For me, it makes sense to share only the part of the page that has the share button, therefore, the embedded link. We can copy the URL from the browser. @fannycc what do you think?

It makes sense to me!

PR #932, opened 2020-01-08
Open accordion if there is only one item

Bluesmile82:
- Open the accordion component item if it is the only one on the accordion.
- Move to hooks.
The item can be closed after the first render

Try: Go to http://localhost:3000/lts/country/USA/sectoral-information

image

### DIFF
--- a/app/javascript/app/components/accordion/accordion-component.jsx
+++ b/app/javascript/app/components/accordion/accordion-component.jsx
@@ -37,7 +37,7 @@ const Accordion = props => {
         data.map((section, index) => {
           const isOpen = getIsOpen(section, index);
           const title = section.parent
-            ? `${section.parent.name} | ${section.title}`
+            ? `${section.parent.name} || ${section.title}`
             : section.title;
           return (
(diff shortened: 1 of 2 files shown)


app/javascript/app/components/accordion/accordion-component.jsx
```diff
@@ -8,75 +8,74 @@ import dropdownArrow from 'assets/icons/dropdown-arrow.svg';
...
+  };
+
+  return (
+    <div className={className}>
+      {data &&
+        data.length > 0 &&
+        data.map((section, index) => {
+          const isOpen = getIsOpen(section, index);
+          const title = section.parent
+            ? `${section.parent.name} | ${section.title}`
```
kfilip94: I see that it was in the code before, but why we're using here bitwise `OR` instead of the logical one? Apparently it's not a bitwise operation, so I'd like to ask you for changing this small bit because it can cause an error one day :pray:
```suggestion
            ? `${section.parent.name} || ${section.title}`
```

PR #939, opened 2020-01-10
Fix/lts dropdown

Bluesmile82:
Fetch NDCs on the summary pages to avoid the Bug with summary cards when accessing country without data.

Steps to Reproduce:

1- Visit the LTS Explore page;
2- Click on Mexico;
3- Search for Bulgaria;
4- Search for United States...  BUG! (No more) The summary data is not loaded correctly. If we change tabs and back it then loads.

Extra:
- Fix z-index on overview - question cards - info button when mobile and share modal is open
image

### DIFF
--- a/app/javascript/app/components/country/country-lts-overview/country-lts-overview.js
+++ b/app/javascript/app/components/country/country-lts-overview/country-lts-overview.js
@@ -1,11 +1,12 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { connect } from 'react-redux';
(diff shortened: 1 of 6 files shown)


kfilip94:
@Bluesmile82 Good job with that, it seems to be working for the `Summary` tab, but sadly it doesn't work for other tabs, try changing the country on the different tab than `Summary`. Would you like to fix this in this PR as well? :pray:

Bluesmile82:
Thanks for spotting this! Now is fixed :)

PR #946, opened 2020-01-16
Fix/ndcs explore title links

Bluesmile82:
Problem:
The links were not correctly associated with the countries entry when we searched on the table
image

Now this is fixed in NDC explore and LTS explore tables


kfilip94:
Hey @Bluesmile82, thanks for fixing that, nice job! :ok_hand:
A similar problem still occurs after sorting, could you provide a fix for this as well? Thank you in advance ;)
nmx8k-6lsko

Bluesmile82:
Thanks! I realised this is a different bug related with the titleLinks not sorting on the table when the data does. This is failing also in the pathways section. I will add a different bug to PT and fix it as I have to change cw-components too

PR #949, opened 2020-01-17
Fix/design review changes

Bluesmile82:
Some changes related to the design review

**NDC- LTS country**

- Cards- instead of carousel for mobile 2 in a row for tablet and 1 in a row for mobile

- Turn Document button (only when one document available) into a Secondary button

- Back button - Only <--- go to explore LTS, <--- go to explore NDC. We don't need back button component anymore

- Country dropdown - Chevron position: reduce min-width

- Adjust compare Button width

image

**Overview**

- Share button - “Copy embed URL” instead of “Copy link”

image

- Remove the section number when we are on embed mode

- Remove the bottom border line on embed.

- Remove /* min-height: 400px; for mobile .commitmentText

image

**Menu**

- Filter the Commitment menu based on the feature flags - Check design. Also, NDC search should be removed from the menu

- Add FEATURE_ALL_COMMITMENTS_MENU_ITEMS feature flag to show all items of Commitments menu on staging

image

**NDC search**

- NDC search should have a <--- go to explore NDCs

image

**Explore NDC**

- Add search to NDC explore

image


kfilip94:
Hey @Bluesmile82, below I left a few suggestions, probably @fannycc could take a look at them:
- Should we allow for the internal y-scroll inside the Card component on mobiles? Or maybe show the whole content of the card at once?
image
- The country menu needs some tweaks on mobiles, also padding between `View full text` and `Search` is missing
image
- I'd decrease the left and right padding in the progress bar component on mobiles, it's unnecessarily big IMHO
image

I'll take a look at the code now :wink:

Bluesmile82:
Thanks for the comments. Fixed: 
- Links that were not working for mobile on LTS and NDC explore
- Dropdown margin
- Country dropdown min and max-width with an ellipsis
- Overview question card padding on mobile

kfilip94:
Hey @Bluesmile82, thanks for the changes.
It seems that the margin on the left side of the cards is too small:
image

Bluesmile82:
Fixed! Thanks!

kfilip94:
Now it looks, great! Let's merge it! :rocket:

fannycc:
hey folks! sorry I was in a workshop yesterday. Thank you for the suggestions @kfilip94 here are my comments: 
`Should we allow for the internal y-scroll inside the Card component on mobiles? Or maybe show the whole content of the card at once?`
We should show all the content of a card at once otherwise in mobile the scroll to see the page will act weird

`The country menu needs some tweaks on mobiles, also padding between View full text and Search is missing`
Besides the margin that I can read it's fixed now, I think we should decrease the size of the country font and I'd also suggest putting the dropdowns bellow the country name. Now it looks like the dropdowns will affect whatever is underneath.
image

`I'd decrease the left and right padding in the progress bar component on mobiles, it's unnecessarily big IMHO`
Agree, maybe we don't even need to put them inside cards, we can use dividers between then (lines) or just space. 
image


Let me know if you have any doubt regarding these or other comments :)

PR #950, opened 2020-01-17
Feature/killin grays

Bluesmile82:
This PR makes necessary changes after talking with Fanny to reduce all the grays to 3 and fix some other colors.

Extra:

- Fix LTS vision description text on LTS country page

- Add margin-bottom to cards in LTS/NDC country as in the design.
image

- Fix NDC and LTS explore styles to match the design.
image

- Remove 'Menu' word from the Hamburger menu
image

- Change info icon to be always blue and don't change on hover (Country page vulnerability)

image

### DIFF
--- a/app/javascript/app/components/about/about-partners/about-partners-styles.scss
+++ b/app/javascript/app/components/about/about-partners/about-partners-styles.scss
@@ -5,7 +5,7 @@
   padding: 50px 0 $gutter-padding 0;
 
   &:not(:last-child) {
-    border-bottom: 1px solid $border-color;
+    border-bottom: 1px solid $gray2;
   }
 }
 
@@ -60,7 +60,7 @@
     .description {
       margin-top: 10px;
       font-size: $font-size;
-      color: $dark-gray;
+      color: $gray1;
     }
   }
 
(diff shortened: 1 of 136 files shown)


kfilip94:
checking

kfilip94:
@Bluesmile82 Hey, thank you for taking care of that! The one issue I found:
Explore LTS & Explore NDCS -> On mobiles border around info buttons section interfuses with the background color, it's worth to check it with @fannycc how to deal with this:
image

NDC tracker & LTS Tracker: 
image

NDC-SDG Linkages
image

Bluesmile82:
Let's leave them like this right now with a white background. We could do a final review when we change all the styles

fannycc:
Hey! :) I found some issues in the previous screenshots:

image
The grey colour of the map is wrong, as it is now isn't accessible at all. I'd suggest to follow this example (which is also in one of your screenshots above, as you can see it's a darker grey) https://projects.invisionapp.com/d/main#/console/18371781/381684662/preview

image
The cards on the left hand side should be white

Regarding the buttons on top of the grey light background I chatted yesterday about it with @Bluesmile82 for a little bit. https://projects.invisionapp.com/d/main#/console/15803555/328013267/preview Here is an example of them. It's ok they border can't be seen because the border of the buttons is the same colour than the background so it'll create the effect of a margin between them.

PR #951, opened 2020-01-21
Feature/update search styles

kfilip94:
This PR applies the updated styles to the Search component. Now we have two basic versions of the Search component:
[PIVOTAL](https://www.pivotaltracker.com/story/show/170768103)

- Standard version:
6wget-p4r4g

- Transparent version (when prop `transparent = true`):
jusep-g1j5r

We can also pass theme to the Search component as before.


kfilip94:
Ok, so we also need to update the border of the dropdowns there, because right now they have 2,5px as well:
image

I'll try to tweak it, thanks for spotting that :wink:

fannycc:
mm let's do this in the next sprint, bc as I am seeing in your example above we are still keeping the previous styles of headers. So as it is now there isn't enough contrast with the background. All ok! feel free to merge :)

Bluesmile82:
Could we add transparent as variant="transparent" instead of a prop? To be consistent and in the case that we need to add more variants

kfilip94:
I'll change it to the "variant" prop, it doesn't make much difference at this point, but maybe there will be third style commonly used in the future

kfilip94:
@Bluesmile82 Done, thanks for your advice! :+1:

PR #953, opened 2020-01-22
Update ndcs_controller slugs

Bluesmile82:
Updated the ndcs_controller to show the data on the Ndc country / overview / contribution type card
I'm not sure if there is a better way to not hardcode this. But this will work by now.

image


simaob:
Would this work on both staging and prod? =D

PR #955, opened 2020-01-22
Update input and textarea styles

Bluesmile82:
Update styles based on the designs

- Cursor on textarea
- Borders on active and inactive inputs
- Disabled input (not used)

image

### DIFF
--- a/app/javascript/app/components/text-area/text-area-styles.scss
+++ b/app/javascript/app/components/text-area/text-area-styles.scss
@@ -1,3 +1,7 @@
+@import '~styles/settings';
+
 .textarea {
   font-size: 1rem;
+  border-radius: 4px;
+  border: solid 2.5px $gray3;
 }
--- a/app/javascript/app/components/text-area/text-area-styles.scss
+++ b/app/javascript/app/components/text-area/text-area-styles.scss
@@ -3,5 +3,5 @@
 .textarea {
   font-size: 1rem;
   border-radius: 4px;
-  border: solid 2.5px $gray3;
+  border: $dropdown-border;;
 }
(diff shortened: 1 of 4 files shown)


app/javascript/app/components/text-area/text-area-styles.scss
```diff
@@ -1,3 +1,7 @@
+@import '~styles/settings';
+
 .textarea {
   font-size: 1rem;
+  border-radius: 4px;
+  border: solid 2.5px $gray3;
```
kfilip94: I've created the constant for this type of the border, you can import it from settings.scss:
`$dropdown-border: 2.5px solid $gray3;`

app/javascript/app/components/text-input/text-input-styles.scss
```diff
@@ -1,5 +1,9 @@
+@import '~styles/settings';
+
 .input {
   font-size: 1rem;
+  border-radius: 4px;
+  border: solid 2.5px $gray3;
```
kfilip94:
```suggestion
  border: $dropdown-border;
```

app/javascript/app/styles/themes/input/text-input-theme.scss
```diff
@@ -5,7 +5,7 @@
   height: 45px;
   border-radius: 4px;
   color: $theme-color;
-  border: solid 1px $gray2;
+  border: solid 2.5px $gray3;
```
kfilip94:
```suggestion
   border: $dropdown-border;
```

kfilip94:
Great @Bluesmile82!
There is one thing about disabled styles of `input` and `text-area` components - on the design it's applied by setting 50% of the opacity on the element (explicitly written only for Search & Dropdown components).  CSS code in Invision also says `opacity: 0.5` for those elements, so I think that's the preferred way,  I did the same thing for dropdowns before.
image

PR #956, opened 2020-01-22
Fix/metadata modal markdown

kfilip94:
This PR fixes a part of the metadata formatting issue in GHG emissions. These changes enable support for markdown syntax in metadata modal in GHG emissions, but they need to update their data files with the correct markdown syntax
[PIVOTAL](https://www.pivotaltracker.com/story/show/170675310) | [BASECAMP](https://basecamp.com/1756858/projects/13795275/todos/406051530)

Once we merge & release this fix, I'll let them know that they can update their CV file.

### DIFF
--- a/app/javascript/app/components/metadata-text/metadata-text-component.jsx
+++ b/app/javascript/app/components/metadata-text/metadata-text-component.jsx
@@ -23,27 +23,17 @@ MetadataAllProps.propTypes = {
   data: PropTypes.object.isRequired
 };
 
-const urlTitles = ['url', 'Link'];
 const MetadataProp = ({ title, data }) =>
   data &&
   (title === 'logo' ? (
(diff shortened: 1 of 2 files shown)


Bluesmile82:
The format and styles of the document are not the same as the ones we have right now. Could you check that?
- Staging:

image

- PR:

image

- Also, about this 'Not specified' data that you deleted. We should keep this data with the class that it had. So it shows with a different opacity

kfilip94:
Hey, on my machine styles work as they should, did you fetch all the changes?
image

Also, when `data === 'Not specified'`, the `empty` class is added to the element:
image

Let me know if I'm missing something :wink:

Bluesmile82:
Ok, It seems I didn't have all the changes. Thank you!

PR #957, opened 2020-01-28
Fix/donut chart layout

Bluesmile82:
This PR fixes the emission donut chart on NDC and LTS explorer:

- Add different text to the inner label to show that these are GHG emissions.
- Show 'Not applicable or no information' for the missing data on the emissions chart
- Instead of recalculating the percentage use the real percentages
- Remove EU28 from the calculation to avoid double counting

image

Extra:

- Fix NDC to data explorer link:
[Basecamp](https://basecamp.com/1756858/projects/13795275/todos/391404521)
Try: https://climate-watch.vizzuality.com/ndcs-content?category=unfccc_process&indicator=revised and go to Data explorer

### DIFF
--- a/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
+++ b/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
@@ -499,8 +499,10 @@ export const parseExternalParams = createSelector(
             ids.map(f => parseInt(f, 10)).includes(i.id) ||
             ids.includes(i.number) ||
             ids.includes(i.iso_code3) ||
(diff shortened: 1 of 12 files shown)


kfilip94:
checking

kfilip94:
Hey, @Bluesmile82 these donut chart changes look good!
I have doubts about the data explorer link fix, once we have `Category` and `Indicator` params in the URL everything works great, but when we don't have any, because we've landed on the NDC Content page the first time from the menu, it doesn't work:
cm3i2-12r14

Bluesmile82:
Thanks, one default was missing!

PR #958, opened 2020-01-28
Feature/update card styles

kfilip94:
This PR updates the card styles across the platform. It removes legacy card component and renames `CardRowLight` component to `CardRow` component.|
[PIVOTAL](https://www.pivotaltracker.com/story/show/170768112)

**COUNTRIES PAGE:**
- Climate Vulnerability and Readiness:
_Desktop:_
image
_Mobiles (previously it was broken):_
40w67-k8kl0

- Nationally Determined Contribution (NDC) Overview
_Desktop:_
image
_Mobiles (removes horizontal and inside scroll):_
2ekya-st4a2

- Targets in Laws and Policies
_Desktop:_
image

_Mobiles (it doesn't change the layout, only updates visuals):_
8ugz2-vkx89


kfilip94:
@fannycc You're right, I missed that bit, I'll change it. Thanks for spotting that :+1:

app/javascript/app/components/card/card-row/card-row-component.jsx
```diff
@@ -1,57 +1,35 @@
...
-          <p
-            key={descKey}
-            className={cx(styles.description, theme.description, {
-              [styles.descriptionExtraPadding]: subtitle
-            })}
-          >
-            {description}
-          </p>
-        )}
+const CardRowComponent = ({ rowData }) => (
```
Bluesmile82: ↪️ 🙌

app/javascript/app/components/card/card-row/card-row-styles.scss
```diff
@@ -1,49 +1,30 @@
 @import '~styles/settings.scss';
 
-$standard-text-padding: 20px;
-$extra-text-padding: 30px;
-
 .cardRow {
-  background-color: $white;
-  border-top: 1px solid $gray3;
+  background-color: white;
```
Bluesmile82: Why we don't use the $white variable? I know its a bit verbose but I prefer not to have hardcoded colors

app/javascript/app/components/country/country-climate-vulnerability/country-climate-vulnerability-component.jsx
```diff
@@ -108,7 +108,11 @@ class CountryClimateVulnerability extends PureComponent {
                       {section.sectionType === 'LINK' ? (
                         <PrepCard className={styles.prepCard} />
                       ) : (
-                        <Card title={section.title} subtitle={section.subtitle}>
+                        <Card
+                          title={section.title}
+                          subtitle={section.subtitle}
+                          contentFirst
```
Bluesmile82: I prefer to have a variant="contentFirst" or "reverse". What do you think?

app/javascript/app/components/sectors-agriculture/drivers-of-emissions/card-pie-chart/card-pie-chart-component.jsx
```diff
@@ -79,14 +97,25 @@ class CardPieChart extends PureComponent {
         <Card
           theme={cardTheme}
           subtitle={subtitle}
+          // contentFirst
```
Bluesmile82: Should we remove this comment?

app/javascript/app/styles/settings.scss
```diff
@@ -58,13 +57,15 @@ $teal-blue: #045f61;
 $purple: #74356a;
 $light-blue: #035388;
 $error: #d46948;
+$shadow-gray: #e8ecf5;
```
Bluesmile82: We are trying not to add more grays to the settings. Maybe this one is necessary @fannycc ?

Bluesmile82:
Somehow we've lost the padding on the bottom of the cards when we have the possibility to scroll. Could you bring it back? Also if you really feel brave you could add some gradient just in the case that 
we need that scroll:

Staging:
image

Now:
image

app/javascript/app/styles/settings.scss
fannycc: I need to know where is that grey being use and for what in order to answer that question :)

Bluesmile82: So its:
- Border bottom of the cards
- Border of quantification tags?
- Border of laws and policies cards
- Box shadow of the cards

I'm not sure about the box-shadow but I think we can use one of the grays we already have

kfilip94: I used the box-shadow from the UI kit, this color is applied there, and we don't have a similar one.

app/javascript/app/components/sectors-agriculture/drivers-of-emissions/card-pie-chart/card-pie-chart-component.jsx
kfilip94: yep, thanks :wink:

app/javascript/app/styles/settings.scss
fannycc: ok right, we can use the lighter one, there is a small difference it's true...I'd say let's keep it if it's being used consistently in shadows. Let me know what you think

kfilip94: So to be clear, do you want me to keep this lighter version, `#f1f4fb` right? :thinking:

app/javascript/app/components/card/card-row/card-row-styles.scss
kfilip94: Thanks, you're right, I forgot to set it that way when I was messing with card components, will change it ;)

app/javascript/app/components/country/country-climate-vulnerability/country-climate-vulnerability-component.jsx
kfilip94: In this case, I'd rather stay with `contentFirst` flag, because we don't have a third option here, it's also how it's done in `cw-components`. The best option would be to extract the header component and build each card by passing children to the Card component in the right order, so we can build each card from pieces that we want.

app/javascript/app/components/card/card-row/card-row-component.jsx
kfilip94: Do you want me to revert that I assume?

app/javascript/app/components/country/country-climate-vulnerability/country-climate-vulnerability-component.jsx
Bluesmile82: Mmm is, ok. Leave it this way 👍

app/javascript/app/components/card/card-row/card-row-component.jsx
Bluesmile82: Hahaha nono, Hooks! 🆒

app/javascript/app/styles/settings.scss
fannycc: I was putting both options on the table :D so I could know your opinion. We can keep shadows in the same colour they are. If we are using that darker grey consistently across all shadows against white bg I think we can leave it that way. The other option was too light as you said right?

kfilip94: I'm not very sharp today, sorry :D The usage of shadows in our app is very inconsistent, `#f1f4fb` shadow we are using lts & ndcs explore page, `#e8ecf5` I applied for updated styles of the cards as it was in UI kit, in many places I see `box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.09); ` and there are also a few similar monsters so it's hard to tell :/ So I'd vote for leaving this as it is now and maybe it's something to identify and review as a separate task, across the whole platform? What do you think?

fannycc: Let's leave it then. But for future style applications let's put the same shadow colour to all of them. If the #f1f4fb works well let's applied that one due to it's already being used in other components. The other grey #e8ecf5 I reckon was in the cards componen in the UI KIT before talking with Alvaro. But we can review it in another moment

kfilip94:
Ok, all changes have been implemented, ready again!

Bluesmile82:
Thanks!

PR #959, opened 2020-01-29
Feature/design revamp headers

Bluesmile82:
This PR updates the headers of the whole page:
This is a scheme of the pages - subpages.
All pages have a gradient and pages a plain color.
image

Headers import their color from the Scss with an export managed by webpack. They get them from the styles/constants file.

- Also, anchor navs were fixed to have the same color from the page and import it from the scss.

image
image
image
image
image
image
image
image
image
image


kfilip94:
checking

Bluesmile82:
Thanks! I deleted the images, compressed the really big ones and installed a webp plugin for webpack on production

PR #963, opened 2020-02-03
Fix/search visibility

Bluesmile82:
- Fix some search inputs that were not visible on production:
https://www.climatewatchdata.org/ndcs/country/ESP/full
image

- Fix styles of header with search input on country ndc on production
https://www.climatewatchdata.org/ndcs/country/FRA/overview
image


kfilip94:
checking

PR #970, opened 2020-02-05
Feature/custom comparison page

kfilip94:
This PR adds a blank "Custom comparison" page under the `/custom-compare` url, requires a `FEATURE_COMMITMENTS_OVERVIEW` flag to run. You can find this in the main menu.
[PIVOTAL](https://www.pivotaltracker.com/story/show/170676363)

image


app/javascript/app/routes/app-routes/app-routes.js
```diff
@@ -192,6 +194,12 @@ export default [
     exact: true,
     headerGradient: HEADER_GRADIENTS.commitments
   },
+  FEATURE_COMMITMENTS_OVERVIEW && {
```
Bluesmile82: The custom compare doesn't have a nav item on the menu. You access it through the compare all targets.

kfilip94: I know, I added that for testing purpose, until the "compare all targets" page will be ready. Should I remove it anyway for not confusing our client?

Bluesmile82: No, it's ok then. Thanks. We just have to remember

kfilip94: ok, thank you :wink:

PR #971, opened 2020-02-05
 Fix progress bar, Try to create a mask instead of adding it on top

kfilip94:
This PR tweaks a bit the progress bar component on NDCS/LTS Explore pages:
[PIVOTAL](https://www.pivotaltracker.com/story/show/170700083)

**BEFORE:**
image

**AFTER:**
image


fannycc:
also I'm guessing the `-1 party ` is a mistake :)

kfilip94:
Thanks! :wink:   This `-1` bug happens on my local machine, probably because I haven't updated my local database yet

PR #972, opened 2020-02-06
Feature/compare all page

Bluesmile82:
This PR starts the compare all page

- The new menu item is at the end of our long commitments list.
- Added header, info, disabled compare button, search and table
- The table is missing:  'Pre-2020 pledge' and 'Targets in National Policies' columns. We still have to find this data on another PR
- Mobile styles
- True values are represented with a blank right now
- Column widths should be adjusted when we have all the columns

image


kfilip94:
- The hint text is different. If it's a new version you discussed with fanny somewhere, we should consider adding some extra information about the maximum 3 submitted boxes that we can choose, as it was in the old one, @fannycc.
image
Besides, it'd be nice to tweak a bit layout on mobiles:
image

- "Compare" button is too wide in comparison to the design:
image
- Additional margin below main menu:
image

- Weird error on Firefox and Chrome in incognito mode: `Maximum call stack size exceeded:
gwgo7-5r8bz

Bluesmile82:
Sorry about the text. I forgot to change it 👍

kfilip94:
- Also, there is one thing that I'd like to discuss: we're not setting the minimum or fixed height for the layout from the beginning. It means when data is loading, we see the entire footer and then, once data is on place, the entire layout jumps and stretches to the right side. It's not a big issue, but I find it quite disturbing when browsing the internet. It's not only this place where it's happening in CW, so I'd like to know your opinion about that and if we should care to avoid unnecessary jumps in layout while developing new pages @fannycc @Bluesmile82?
1rlro-y670z


I'm checking the code now :mag:

Bluesmile82:
Yeah, sure, I'll add a min-height

fannycc:
Thanks for the heads up! I agree on setting a min-height to avoid jumps and I have 2 additional comments: 
1. Is it possible to align text lines so we don't end up with that weird text paragraph? 
Screenshot 2020-02-06 at 13 36 12

2. Why do we have 2 spinners at the same time running?

Bluesmile82:
Ready again! Thanks

app/javascript/app/pages/ndc-compare-all-targets/ndc-compare-all-targets-styles.scss
```diff
@@ -19,26 +19,33 @@
...
 }
 
 .compareButton {
   width: 100%;
 
-  @media #{$tablet-portrait} {
-    width: 120px;
+  @media #{$tablet-landscape} {
+    max-width: 120px;
+    min-width: 120px;
```
kfilip94: I was wondering why not `width: 120px;` so I checked that, and it seems that foundation overrides stuff weirdly. I'm not a big fan of this library because I feel that we have less control over the CSS at the end, and then we need to hack our own code, but we could build using native flex or grid instead. It just the philosophical thought, no actions needed :D

kfilip94:
Hey, @Bluesmile82 thank you for your changes :+1: The last thing that still needs to be tweaked is the issue described by fanny in paragraph one, it still occurs on mobiles:
image

PR #974, opened 2020-02-10
Fix/loader on iframe

Bluesmile82:
There was a bug warning that the contact form was not displayed on firefox. This might be due a temporary problem on the target website (out of our reach) or due to a long time waiting. For this case I added a loader so the user doesn´t feel its broken.

image

Extra:

- Fix overview page dots. Remove the one at the end of the title and add the one after the number.
- Don't always capitalize the text in buttons: We were ending with some strange behavior like "Countries **And** Submissions"

### DIFF
--- a/app/javascript/app/components/about/about-contact/about-contact-component.jsx
+++ b/app/javascript/app/components/about/about-contact/about-contact-component.jsx
@@ -1,26 +1,34 @@
-import React from 'react';
+import React, { useState } from 'react';
 import cx from 'classnames';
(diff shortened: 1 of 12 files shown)


kfilip94:
Nice job @Bluesmile82 !
Only one suggestion: could you increase space between loading spinner and `Subscribe to our newsletter (...)` text?
Screenshot from 2020-02-11 10-23-06

Bluesmile82:
This was centered and working in all browsers. Maybe you hadn't pushed

PR #976, opened 2020-02-11
Feature/update explore tooltip

Bluesmile82:
This PR updates the NDC and LTS explore map tooltips adding the value of the legend and the emissions of each country

image

Extra: MinAngle config prop is added to display very small values as 'Second NDC submitted'

image


Bluesmile82:
I think this one is 0.01% they asked for one decimal. maybe we can add a second decimal if its less than 0.1 What do you think @fannycc?

fannycc:
@Bluesmile82 makes sense to me, otherwise it looks like there is nothing at all :)

Bluesmile82:
Fixed!

PR #985, opened 2020-02-18
Add share "copy embed code" to the share modal & fix "copy embed link…

kfilip94:
This small PR improves a few issues pointed out in feedback:
image

- removes disabled scrollbars from the share menu,
- fixes broken `Copy embed URL` feature (it was pointing to the page link, not embed version),
- adds `Copy embed code` option to the share modal,

image

However, I discovered another bug in the `Overview` section - embed URL doesn't contain information about shared Section so at the end, this link redirects to the homepage. It happens because clicking on any `Share` button dispatches an action `setShareModal` that displays all three instances of the `ModalShare` component, one on top of the other. I'll put that as [a bug to the Pivotal Tracker](https://www.pivotaltracker.com/story/show/171335591)

2rfer-7qg9c

### DIFF
--- a/app/javascript/app/components/modal-share/modal-share-component.jsx
+++ b/app/javascript/app/components/modal-share/modal-share-component.jsx
@@ -96,13 +96,15 @@ const ModalShare = ({
       onRequestClose={() => setShareModal({ open: false })}
       header={<ModalHeader title="Share" />}
     >
-      <ul className={styles.links}>
-        {shareMenuOptions.map(option => (
-          <li className={styles.linkContainer} key={option.label}>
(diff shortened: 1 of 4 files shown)


Bluesmile82:
We have a problem with tablet portrait as the modal shrinks and the last button goes out of the window. We could use two rows or use the mobile version before. I will go for the mobile version if its easier

image

kfilip94:
hey @Bluesmile82, thanks for your feedback! I decided to use two-rows layout for tablets and desktop:

**Tablet:**
image

**Desktop:**
image

PR #989, opened 2020-02-21
Feature/compare all targets search params

kfilip94:
This PR saves the document selection in the URL as a `targets` param on `Compare all targets`  and  `Custom compare` pages. Navigation through `Go to compare all targets` button passes the `targets` params to the `Compare all targets` page. It also removes the `Custom compare` option from the `Commitments` menu.
[PIVOTAL](https://www.pivotaltracker.com/story/show/171315679)

mc1l1-38xzt

I hope that this solution is not overcomplicated, let me know what do you think ;)

### DIFF
--- a/app/javascript/app/pages/custom-compare/custom-compare-selectors.js
+++ b/app/javascript/app/pages/custom-compare/custom-compare-selectors.js
@@ -1,5 +1,6 @@
 import { createSelector } from 'reselect';
 import qs from 'query-string';
+import { uniq } from 'lodash';
 
 const INDC = ['INDC Submitted', 'First NDC Submitted', 'Second NDC Submitted'];
 const NDC = ['First NDC Submitted', 'Second NDC Submitted'];
@@ -9,6 +10,15 @@ const getIndicatorsData = state =>
   (state.compareAll.data && state.compareAll.data.indicators) || null;
 const getQuery = (state, { search }) => search || '';
 
+export const getBackButtonLink = createSelector([getQuery], query => {
+  if (!query) return '';
+  const targetParams = query.targets ? query.targets.split(',') : [];
+  const documentParams = uniq(
+    targetParams.filter(target => !target.endsWith('-'))
+  );
+  return `/compare-all-targets?targets=${documentParams}`;
+});
+
 const getCountryOptions = createSelector([getCountries], countries => {
   if (!countries) return [];
   return countries.map(({ wri_standard_name, iso_code3 }) => ({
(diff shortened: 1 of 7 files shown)


kfilip94:
@Bluesmile82 ready for reviewing!

app/javascript/app/pages/custom-compare/custom-compare-selectors.js
```diff
@@ -17,42 +27,39 @@ const getCountryOptions = createSelector([getCountries], countries => {
...
-    document0,
-    document1,
-    document2
-  };
+  const queryTargets = query.targets ? query.targets.split(',') : [];
+  return [1, 2, 3].map((value, i) => {
+    // targets are saved as a string 'ISO3-DOCUMENT', e.g. 'USA-NDC'
+    const target =
+      queryTargets && queryTargets[i] && queryTargets[i].split('-');
+    const country = target && target.length > 0 && target[0];
```
Bluesmile82: Why this extra length > 0? And shouldn't it be only length > 0 for document?

app/javascript/app/pages/custom-compare/custom-compare.js
```diff
@@ -17,19 +22,52 @@ const mapStateToProps = (state, { location, route }) => {
...
+    };
     history.replace(getLocationParamUpdated(location, params));
   };
 
-  return <Component {...props} handleFilterChange={handleFilterChange} />;
+  const handleCountryFilterChange = (targetKey, newCountry) => {
+    const { selectedTargets } = props;
+    const newTargetParams = [];
+    selectedTargets.forEach(({ key, country, document = '' }) => {
+      if (key === targetKey) newTargetParams.push(`${newCountry}-`);
```
Bluesmile82: Shouldn't we always have a country and a document? Why add a country without a document?

kfilip94: After changing the country we're not selecting automatically the document, we leave the dropdown empty. The question is what do we want to do with that @fannycc @Bluesmile82?
vaq2b-z5etk

app/javascript/app/pages/custom-compare/custom-compare-selectors.js
kfilip94: :man_facepalming: Oh, you're right,  we don't need to check `target.length` in both cases, we will get `undefined` anyway, thanks!

kfilip94:
Hey @Bluesmile82, ready again!

Bluesmile82:
Great! I can see now that we have the case of Lybia without documents. Works perfectly!

PR #1000, opened 2020-03-06
Stories and NDC Texts import from backend

simaob:
This Pull Requests makes changes to the Admin side of Climate Watch to enable importing blog stories and NDC Full texts via the admin interface. This was not possible before, but now it is!

I've removed the file management from the NDC Texts interface because it was not working due to the size of the files, but WRI have access to S3, so they can upload it there and then run the importer here.

To test this you will need to run a few commands on your terminal:

```
bundle install
```

And then you should have these two pages available:

- http://localhost:3000/admin/global_cw_platform_ndc_texts
- http://localhost:3000/admin/global_cw_platform_blog_stories

And they should work! =)


simaob:
Hey @Bluesmile82 can you have a look at this PR when you have the chance?

PR #1001, opened 2020-03-10
Add back button and remove 2020 tracker from menu

Bluesmile82:
- Add Go to 2020 NDC Tracker in NDCs explore
- Add back to explore NDC's Tracker in 2020 NDCs tracker
- remove 2020 from commitments menu

image
image
image


Bluesmile82:
I think these are different pages with different indicators so I don't think we should preserve the params. It's not a real back button for me but just a link between pages

kfilip94:
That's fine, I see that we don't keep URL params in different places either, where we use `Go to` button. Let's merge it then!

PR #1003, opened 2020-03-11
Feature/explore sticky dropdowns

Bluesmile82:
Add a new sticky bar for indicator selection on NDC and LTS explore. 
Don't include the info Button

sticky

### DIFF
--- a/app/javascript/app/components/ndcs/ndcs-explore-map/ndcs-explore-map-component.jsx
+++ b/app/javascript/app/components/ndcs/ndcs-explore-map/ndcs-explore-map-component.jsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/no-danger */
-import React, { PureComponent } from 'react';
+import React, { useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import { TabletLandscape } from 'components/responsive';
 import Map from 'components/map';
@@ -15,6 +15,8 @@ import CustomInnerHoverLabel from 'components/ndcs/shared/donut-custom-label';
 import LegendItem from 'components/ndcs/shared/legend-item';
 import ShareButton from 'components/button/share-button';
 import ModalShare from 'components/modal-share';
+import Sticky from 'react-stickynode';
+import cx from 'classnames';
 
 import newMapTheme from 'styles/themes/map/map-new-zoom-controls.scss';
 import layout from 'styles/layout.scss';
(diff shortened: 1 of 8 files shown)


Bluesmile82:
@fannycc do you mean overlapping the search input? The user won't be able to search then. I thought it was part of the table

fannycc:
yes, it is and we should show it (the search input). I understand then that you are making a fixed nav (white) and not only the 2 inputs. So the search input will cover by white fixed space

app/javascript/app/components/ndcs/ndcs-explore-map/ndcs-explore-map-component.jsx
```diff
@@ -76,70 +78,69 @@ const renderLegend = legendData => (
...
-  renderDonutChart = emissionsCardData => (
-    <div
-      className={styles.donutContainer}
-      ref={r => {
-        this.setState({ pieChartRef: r });
-      }}
-    >
+function NDCSExploreMap(props) {
+  const tooltipParentRef = useRef(null);
+  const pieChartRef = useRef(null);
```
kfilip94: Thanks for using hooks here! :+1:

db/secondbase/structure.sql
```diff
@@ -5,23 +5,10 @@ SET client_encoding = 'UTF8';
...
-CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
-
-
---
--- Name: EXTENSION plpgsql; Type: COMMENT; Schema: -; Owner: -
---
-
-COMMENT ON EXTENSION plpgsql IS 'PL/pgSQL procedural language';
-
-
```
kfilip94: Are these changes intended?

db/structure.sql
```diff
@@ -5,23 +5,10 @@ SET client_encoding = 'UTF8';
...
-CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
-
-
---
--- Name: EXTENSION plpgsql; Type: COMMENT; Schema: -; Owner: -
---
-
-COMMENT ON EXTENSION plpgsql IS 'PL/pgSQL procedural language';
-
-
```
kfilip94: Are these changes intended? [2]

db/secondbase/structure.sql
Bluesmile82: 🤔 Maybe I should remove those

Bluesmile82: @simaob are this alright?

simaob: hmmm these look weird, but maybe your version of Postgresql caused them. Might be worth not including them.

PR #1004, opened 2020-03-12
Add labels to icons dropdown

Bluesmile82:
This PR adds the labels to the icons dropdown only when is opened. This needs passing a string as a customLabel to be able to reposition the dropdown only when is opened. This can be a first iteration for the new dropdown styles.

Laptop:
image

Tablet:
image

### DIFF
--- a/app/javascript/app/components/ghg-emissions/ghg-emissions-styles.scss
+++ b/app/javascript/app/components/ghg-emissions/ghg-emissions-styles.scss
@@ -37,10 +37,16 @@ $min-height: 500px;
 }
 
 .col4 {
-  @include columns((6, 6, 6, 6, 8, 4));
+  @include columns(12);
 
   > div {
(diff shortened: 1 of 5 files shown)


app/javascript/app/components/ghg-emissions/ghg-emissions-styles.scss
```diff
@@ -88,3 +94,14 @@ $min-height: 500px;
...
   }
 }
+
+:global .icons-dropdown {
+  &.open {
+    min-width: 240px;
+
+    @media #{$tablet-landscape} {
+      margin-left: -130px;
+      background-color: white;
```
kfilip94: Could you use `$white` constant here?

Bluesmile82:
Thanks for the review! The filters are very crowded to have individual icons with tooltips as you will see in the designs

PR #1010, opened 2020-03-19
Feature/update indc importers

simaob:
This PR updates the INDC importers and endpoints to allow us to store indicators per document: INDC, first NDC, second NDC, etctera NDC.

To test this PR you will need to run the following tasks:

```
bundle exec rails db:migrate
bundle exec rails indc:import
```
That last command will import the updated data from the staging bucket and as normally it will take a bit.

In this PR the main changes are:

- Adding a new Indc::Document table to collect information on the different documents submitted by the countries;
- Updating the import scripts to link different country-indicator-value combinations to documents, so that we can query specific submission values;
- Adding a new importer for indicators that are not document specific to be used in the Overview page for instance, but in other places as well;
- Adding a new filtering param for the `/api/v1/ndcs` and `/api/v1/ndcs/:iso/content_overview` endpoints. The new param is called `document` and the values to be used are the slugs of the submitted documents. For now these are available in: `api/v1/data/ndc_content/documents`.

These changes will allow us to update the NDC Explore Country page with the Documents dropdown and then filtering per document the indicators shown. And also the Compare All Flow.

The missing endpoint that I would like to discuss with you @Bluesmile82 is how to pass to the front end the information about the documents available to each country, should we add a filter to that documents endpoint, or provide you a full list of Country: [Documents] information? Let me know!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ language: ruby
 rvm:
 - 2.5.1
 addons:
-  postgresql: 12
+  postgresql: 9.6
 before_script:
   - bundle exec rake db:create db:structure:load RAILS_ENV=test
   - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ language: ruby
 rvm:
 - 2.5.1
 addons:
-  postgresql: 9.6
+  postgresql: 12
 before_script:
   - bundle exec rake db:create db:structure:load RAILS_ENV=test
   - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
--- a/app/controllers/api/v1/data/ndc_content/documents_controller.rb
+++ b/app/controllers/api/v1/data/ndc_content/documents_controller.rb
@@ -1,0 +1,17 @@
+module Api
+  module V1
+    module Data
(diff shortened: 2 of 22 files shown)


Bluesmile82:
The endpoint doesn't seem to be providing with the emissions indicator (slug: 'ndce_ghg') anymore. Could you check this?

simaob:
That indicator seems to be missing from the data. I'm going to look at the data! =)

simaob:
OK, the indicator exists in the data, but there's no data for the indicator, so it's removed after the import. It's a data issue, so something they'll fix! =) No biggie.

simaob:
But did the others work, Álvaro?

Bluesmile82:
Thanks for the countries_documents endpoint. It's working fine! We probably don't need the created_at and updated_at. And maybe if we have all the information on the documents array we won't need the data array unless we need it for the laws or policies.

simaob:
But the data array is where we link countries to documents, no?

Bluesmile82:
Ah, ok. Sorry. I see it now. I thought it was the other way. Also, I see we still need to add the document date. Maybe we can get it from another indicator? It will be nice to have this here too

image

simaob:
oh, right... hmmm, need to figure out how to do that. I'll get back to you when it's done.

simaob:
I will only be able to look at the date thing this afternoon. I'll keep you posted.

simaob:
I've excluded the created_at and the updated_at. Looking at the data, it doesn't seem that we have the dates needed for all the documents, so let's leave that bit out for now.

PR #1013, opened 2020-03-27
Feature/lts explore feedback

kfilip94:
This PR solves three issues from the client's feedback:

- Remove emissions value from map tooltip:
[PIVOTAL](https://www.pivotaltracker.com/story/show/171926847)
image

- Add tooltips to the dropdown filters:
[PIVOTAL](https://www.pivotaltracker.com/story/show/171926835)
Screenshot from 2020-03-27 18 09 43

- Changes the type of the cursor displayed on the map, from `grab` to `pointer`
[PIVOTAL](https://www.pivotaltracker.com/story/show/171926787)
za811-7kqmq-001

Each of these changes has been applied to both Explore pages, LTS and NDCs.

### DIFF
--- a/app/javascript/app/components/ndcs/lts-explore-map/lts-explore-map-component.jsx
+++ b/app/javascript/app/components/ndcs/lts-explore-map/lts-explore-map-component.jsx
@@ -150,7 +150,9 @@ function LTSExploreMap(props) {
                         value={selectedCategory}
                         hideResetButton
                         plain
-                        showTooltip
+                        showTooltip={
+                          selectedCategory && selectedCategory.label.length > 14
+                        }
                       />
                       <Dropdown
                         label="Indicator"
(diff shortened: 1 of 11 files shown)


app/javascript/app/components/ndcs/lts-explore-map/lts-explore-map-component.jsx
```diff
@@ -150,6 +150,7 @@ function LTSExploreMap(props) {
                         value={selectedCategory}
                         hideResetButton
                         plain
+                        showTooltip
```
Bluesmile82: Maybe an improvement could be adding a condition for this to show if:
`showTooltip={selectedCategory && selectedCategory.length > 14}`
I think is quite annoying that it shows all the time. But I leave it up to you

kfilip94: Nice suggestion, I pushed a commit with this tweak. Thanks! :+1:

PR #1014, opened 2020-03-29
Feature/ndc country document dropdown

Bluesmile82:
This PR adds a dropdown that selects a document from the country NDC page

- The Full document link is now always a button that is linked to the document selection
- Both summary and accordion data is updated with the document selection
- When we start without a document the last document available is filled into the URL param. This is necessary for the filtering and the link
- The dropdown is only showing the English documents as we only have the information in English for this page. Once on the full page, you can select different languages. 

Try: Go to an NDC country page and change the document, country, tab, go to the full document, ...

image

### DIFF
--- a/app/javascript/app/components/anchor-nav/anchor-nav-component.jsx
+++ b/app/javascript/app/components/anchor-nav/anchor-nav-component.jsx
@@ -22,56 +22,57 @@ const AnchorNav = props => {
     <div>
       <div className={cx(styles.anchorContainer)}>
         <nav className={cx(className, theme.anchorNav)}>
-          {links.map((link, index) => {
-            const linkProps = {
-              key: link.label,
-              className: theme.link,
-              activeClassName: theme.linkActive,
-              to: {
-                search: link.search || query,
-                pathname: link.path,
-                hash: link.hash
-              }
-            };
-            if (link.activeQuery) {
-              linkProps.isActive = (match, location) => {
-                const activeSearchQuery = qs.parse(location.search)[
-                  link.activeQuery.key
-                ];
-                const linkSearchQuery = link.activeQuery.value;
-                return (
-                  activeSearchQuery === linkSearchQuery ||
-                  (index === 0 && !activeSearchQuery)
-                );
+          {links &&
+            links.map((link, index) => {
+              const linkProps = {
+                key: link.label,
+                className: theme.link,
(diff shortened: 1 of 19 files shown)


app/javascript/app/components/anchor-nav/anchor-nav-component.jsx
```diff
@@ -22,56 +22,57 @@ const AnchorNav = props => {
...
-                  el.scrollIntoView(true);
-                  if (offset) window.scrollBy(0, offset[index]);
-                }}
-                replace
-              >
-                {link.label}
-              </NavHashLink>
-            );
-          })}
+            })}
```
kfilip94: Hey @Bluesmile82, could you move all this logic to the container component? Too much stuff is going on here and it's harder to read I think.

Bluesmile82: I just added the links clause because it was breaking on some edge cases. Let's add a task to clean this if needed

app/javascript/app/components/country/country-ndc-overview/country-ndc-overview-selectors.js
```diff
@@ -1,25 +1,65 @@
...
-export const getLastDocument = createSelector(
-  [getDocuments, getIso],
+const getIso = (state, { iso }) => iso || null;
+const getDataDocuments = state =>
+  (state && state.ndcsDocumentsMeta.data) || null;
+const getOverviewData = state =>
+  state.ndcContentOverview.data && state.ndcContentOverview.data.locations;
+const getCountryOverviewData = createSelector(
+  [getOverviewData, getIso],
+  (data, iso) => (data && data[iso] && data[iso]) || null
```
kfilip94: Do we need the second `data[iso]` here?

kfilip94:
Is the `FIRST` option descriptive enough? Shouldn't we set `FIRST NDC` or something?
image

app/javascript/app/components/anchor-nav/anchor-nav-component.jsx
kfilip94: I'll add the task then, otherwise we will postpone it on never :)

Bluesmile82:
Let's merge this. I think the 'First' problem is about a wrong data. It should be 'first-ndc' translated to 'FIRST NDC'

PR #1017, opened 2020-04-02
Feature/start ghg zoom

Bluesmile82:
This PR adds the data zoom selector to the GHG emissions chart
Missing: We have to position the selector between the chart and the legend. That will need some work on the climate-watch-components.

The change only occurs after the handle is released. The rendering of the chart was not fluid enough.

image


kfilip94:
- I think the line on the timeline chart looks slightly different on the design (thicker line, light grey), here it's thinner and blue.
**DESIGN:**
image
**IMPLEMENTATION**
image
- I think we should add the tooltip to the handlers while changing the date, we can't see what dates we're choosing

app/javascript/app/components/ghg-emissions/data-zoom/data-zoom-component.js
```diff
@@ -0,0 +1,122 @@
...
+
+  const handleStop = () => {
+    onYearChange(getYear('min'), getYear('max'));
+  };
+
+  const handleDrag = (ui, handleType) =>
+    setPosition({
+      ...position,
+      [handleType]: position[handleType] + ui.deltaX
+    });
```
kfilip94, line 56: Could you move all these functions and hooks to the container component? And leave only render methods here? Also please change the extension of this component file from `.jsx` to `.js`

Bluesmile82:
I will change the line. About the tooltip, what do you think @faustoperez ? I'm not sure if its needed

kfilip94:
Everything else looks great, good job @Bluesmile82, this new slider works very smoothly! :1st_place_medal:
About the slider tooltip, I think we need to give feedback to the user about the date that the slider is currently pointing. Without it, the user can only guess where is the desired date on the timeline. It doesn't have to be a tooltip, let's wait for Fausto's comment here :)

Bluesmile82:
Ready again I included it in the right position, fixed the styles and moved the logic code to the container

image

kfilip94:
Hey @Bluesmile82, thank you for your changes!
One last thing that I'd like to ask you: since you've made some changes in `ghg-emissions-component.jsx`, could you move all hooks, handle methods and maybe data objects to the container component `ghg-emissions.js`? We have handle functions there so it's a bit noisy to have them in both files.

Bluesmile82:
Hey! I moved some logic to the container. I prefer to leave the data objects on the component as they will be easier to find. Take a look!

PR #1019, opened 2020-04-03
Feature/update overview slugs

Bluesmile82:
This PR updates the overview page slugs so the last two indicators can be loaded
To get the data you should `rails indc:import` with the staging CW_API


Bluesmile82:
I had some. Did you reimported and pointed to staging?

PR #1020, opened 2020-04-07
Feature/split table

Bluesmile82:
This PR adds the splittedColumns prop to the local Table component that creates two different tables synced by a ScrollSync component (from React virtualized). Also, it applies it to the GHG emissions page table and styles it according to designs.

table


kfilip94:
I can't scroll the table fully to the right side, it clips a few columns there:
[demo link](http://localhost:3000/ghg-emissions?regions=AILAC%2CBRICS%2CEAP%2CECA%2CBRA%2CCHN%2CIND%2CRUS%2CZAF%2CTOP&source=17)
image

kfilip94:
- Show data by -> `Regions-Per GDP` - the unit doesn't fit the row, not sure if this is an old issue or related with these table changes:
image

- Sometimes the columns are too narrow for greater values:
[demo link](http://localhost:3000/ghg-emissions?breakBy=gas&chartType=line&gases=59%2C60%2C61%2C63%2C62&regions=TOP&source=17)
image

- It doesn't look good on the mobiles, but I assume we can live with that for now?:
image

Bluesmile82:
Ok, this needed a couple of fixes. Thanks for the comments. I couldn't reproduce this one 'Sometimes the columns are too narrow for greater values:' Could you check it and if it fails give me the exact parameters? The ids are not the same in different environments (This is a bug we have to fix)

kfilip94:
Hey, try these filters:
image

Bluesmile82:
I updated this PR adding:

- Fix for the border style
- Fix minimum width for every column so the numbers don't get cut
- Added links for the countries and EUU leading to the country page

Ready again!

app/javascript/app/components/ghg-emissions/ghg-emissions-selectors/ghg-emissions-selectors-data.js
```diff
@@ -503,8 +508,31 @@ export const getTableData = createSelector(
...
       )
     }));
+    return tableData;
+  }
+);
 
-    return pivot;
+export const getTitleLinks = createSelector(
+  [getTableData, getModelSelected, getRegions, getCountries],
+  (data, model, regions, countries) => {
```
kfilip94, line 517: Both selectors `getRegions` and `getCountries` can return `null`, could you add guards for `regions` and `countries` here?

PR #1022, opened 2020-04-13
Fix/lts countries count

Bluesmile82:
- The count of countries represented in LTS explore map should be - 1 as we were receiving the EUU as a country. Now its updated.

- Removed LTS tracker code and menu item and redirected this URL to LTS explore
https://basecamp.com/1756858/projects/13795275/todos/415101842

image


kfilip94:
Two errors when I try to run the app from this branch:
image

PR #1026, opened 2020-04-20
Overview feedback

Bluesmile82:
https://www.pivotaltracker.com/story/show/172358900

- Info is not working when clicked
- Embed URL has an empty space below the chart that should be reduced (same for all embed links)


weberjavi:
@Bluesmile82 I would need some more context on how to test this PR. Is this the url I should join to test it:
http://localhost:3000/ndc-overview

PR #1030, opened 2020-04-21
Change Ids to slugs in GHG emissions

Bluesmile82:
Source, gas, and sector had ids on the URL instead of slugs. Now name is used for sources and kebabcased label for gases and sectors.

image
This translates now to:
http://localhost:3000/ghg-emissions?breakBy=sector&chartType=line&gases=all-ghg&regions=TOP&sectors=agriculture&source=CAIT

Try changing the filters. Using the defaults and reloading the page. It should all work

Also, in Country pages:
image
Should have this URL:
http://localhost:3000/countries/USA?source=UNFCCC_AI


simaob:
Hi @Bluesmile82 we should also review the defaults to make sure they are working? When I go to the page no chart is visible, maybe because the defaults are not being found?

Bluesmile82:
Good catch! Now is fixed. I also fixed the GHG emissions chart on the country pages to use the slugs instead of ids. We will have to notify them also about this

PR #1032, opened 2020-04-22
Adds option for aggregated results

Bluesmile82:
This small PR adds an option to the show by dropdown: Regions - Total aggregated
This option works the same as the Regions total option but only displays aggregated regions without expanding them like on the other Regions option.

image


kfilip94:
checking

Bluesmile82:
Yes, that is the expected behavior

kfilip94:
Ok, thanks 👌

PR #1041, opened 2020-04-23
Feature/new png download dropdown

weberjavi:
### Description
This PR adds a download dropdown on the `buttonGroup` of the GHG page. It also updates the share button and adds descriptions to the Intro and the title.
### Designs
Designs could be found [here](https://projects.invisionapp.com/d/main#/console/18371781/381684668/preview) and [here](https://projects.invisionapp.com/d/main#/console/18371781/381684669/preview) (with some more detail).
Also take into account this [basecamp conversation](https://basecamp.com/1756858/projects/13795275/messages/90482167).
I'm not sure if we should update the `styles` for the `SimpleMenu` to reflect the changes on the designs, that might impact many parts of the app. Just let me know if you think it makes sense make it in this same PR.
### Testing instructions
Go to `localhost:3000//ghg-emissions` and test the new button (png download has no functionality implemented yet)
### Feature relevant tickets
[feature task](https://www.pivotaltracker.com/story/show/171697624)

### DIFF
--- a/app/javascript/app/components/ghg-emissions/ghg-emissions-component.jsx
+++ b/app/javascript/app/components/ghg-emissions/ghg-emissions-component.jsx
@@ -97,6 +97,27 @@ function GhgEmissions(props) {
       onClick: handleInfoClick
     },
     {
+      type: 'download',
+      section: 'ghg-emissions',
+      link: downloadLink,
+      tooltipText: 'View or download raw data'
+    },
+    {
+      type: 'downloadCSV',
+      tooltipText: 'Download data in csv',
+      onClick: handleDownloadDataClick
+    },
+    {
+      type: 'addToUser'
+    }
+  ];
+
+  const buttonGroupGHGemissions = [
+    {
+      type: 'info',
+      onClick: handleInfoClick
+    },
+    {
       type: 'downloadCombo',
       options: [
         {
@@ -222,7 +243,9 @@ function GhgEmissions(props) {
   const renderButtonGroup = () => (
     <ButtonGroup
       className={styles.buttonGroup}
-      buttonsConfig={buttonGroupConfig}
+      buttonsConfig={
+        FEATURE_NEW_GHG ? buttonGroupGHGemissions : buttonGroupConfig
+      }
     />
   );
 
@@ -245,19 +268,16 @@ function GhgEmissions(props) {
             sharePath={'/embed/ghg-emissions'}
           />
         </TabletLandscape>
-        <p className={styles.bodyText}>
-          Explore GHG emissions from multiple data source (CAIT, PIK, UNFCCC)
-          and understand their differences in the
-          <a
-            className={styles.link}
-            href="about/faq/ghg"
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            {' '}
-            FAQ
-          </a>
-        </p>
+        {FEATURE_NEW_GHG && (
+          <p className={styles.bodyText}>
+            Explore GHG emissions from multiple data source (CAIT, PIK, UNFCCC)
+            and understand their differences in the
+            <a className={styles.link} href="about/faq/ghg">
+              {' '}
+              FAQ
+            </a>
+          </p>
+        )}
       </div>
       <WorldBankDataProvider />
       <RegionsProvider />
--- a/app/javascript/app/components/ghg-emissions/ghg-emissions-component.jsx
+++ b/app/javascript/app/components/ghg-emissions/ghg-emissions-component.jsx
@@ -16,9 +16,11 @@ import EmissionsProvider from 'providers/emissions-provider';
 import RegionsProvider from 'providers/regions-provider';
 import WorldBankDataProvider from 'providers/wb-country-data-provider';
 import ButtonGroup from 'components/button-group';
+import ShareButton from 'components/button/share-button';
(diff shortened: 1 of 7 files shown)


db/secondbase/structure.sql
```diff
@@ -3,21 +3,35 @@ SET lock_timeout = 0;
 SET idle_in_transaction_session_timeout = 0;
```
Bluesmile82, line 3: Not sure if we should include this migration? @simaob

app/javascript/app/pages/ghg-emissions/ghg-emissions-component.jsx
```diff
@@ -29,7 +29,10 @@ class GhgEmissions extends PureComponent {
         {!isPageContained && (
           <Header route={route}>
             <div className={cx(layout.content, styles.header)}>
-              <Intro title="Historical GHG Emissions" />
+              <Intro
+                title="Historical GHG Emissions"
+                description="Climate change causing greenhouse gas emissions have increased 50 fold since the late 1800s. Energy makes up nearly two-thirds of global emissions, followed by agriculture. Within the energy sector, electricity and heat generation make up the largest portion of emissions, followed by transportation and manufacturing. In 2014, 60% of global greenhouse gas emissions came from just 10 countries, while the 100 least-emitting contributed less than 3%. "
```
Bluesmile82, line 34: Thank you for adding this. There are some links and we would have to check where they go to

Bluesmile82: We need to how all these changes. The dropdown and the text only if we have FEATURE_NEW_GHG=true. Sorry for the late notice. We have been deploying features to production and we want to be more strict about this

app/javascript/app/components/ghg-emissions/ghg-emissions-component.jsx
```diff
@@ -97,11 +97,22 @@ function GhgEmissions(props) {
...
+          label: 'Download current data (CSV)',
+          action: handleDownloadDataClick
+        },
+        {
+          label: 'Save as image (PNG)',
+          action: () => {}
+        },
+        {
+          label: 'Go to data explorer',
+          link: downloadLink
```
Bluesmile82, line 112: This is opening a new tab and probably we dont want to do that as it is internal navigation. What do you think @faustoperez ?

app/javascript/app/components/ghg-emissions/ghg-emissions-styles.scss
```diff
@@ -1,14 +1,15 @@
 @import '~styles/layout.scss';
+@import '~styles/ui-extends.scss';
 
 $min-height: 500px;
 
 .titleContainer {
   @include columns((12, 12));
-
+  margin: $gutter-padding 0px;
```
Bluesmile82: You can remove the 'px' here. It's not needed

app/javascript/app/components/ghg-emissions/ghg-emissions-styles.scss
```diff
@@ -68,11 +72,20 @@ $min-height: 500px;
 }
 
 .title {
-  font-size: $font-size-large;
-  font-weight: $font-weight;
+  @extend %headline;
```
Bluesmile82, line 77: Nice extends 📶

weberjavi:
Thanks for the review @Bluesmile82. Gonna address the comments.
@simaob , @faustoperez please let me know your thoughts on the bits where you were mentioned.

app/javascript/app/pages/ghg-emissions/ghg-emissions-component.jsx
weberjavi: Are you referring to the buttons? I mean there should be two versions of the `buttonGroup`? One with the old set up and one with the new?

**new**:
image

**old**:
image


If that's the case wouldn't it be better to have a separated branch and not merge it to `develop`?

db/secondbase/structure.sql
simaob: @weberjavi , what version of postgres are you running locally? If you didn't change the migrations, then maybe don't include the changes to this file! =)

weberjavi: I'm using version `9.6.0.0`
I did not changed the migrations, so I will revert those changes.

app/javascript/app/pages/ghg-emissions/ghg-emissions-component.jsx
Bluesmile82: I was referring to all the changes. I'm sorry for the late notice but we don't want to show the changes directly on production. As for the buttons is ok if we keep the share button, but the download should show as a button on production. 
I prefer to handle this with a feature flag to avoid conflicts between the different tasks and easier to deploy when we need it. We already have the FEATURE_NEW_GHG for this. If you feel that this is not the best option we can discuss this.

weberjavi: Ok. Thanks for the clarifications. I'll add the `csv` download and the `data explorer download` to the button group then.

weberjavi:
@Bluesmile82 ☝️ addressed the comments. Let me know if there's something missing.

Bluesmile82:
Thank you so much for the changes! Is perfect now

PR #1043, opened 2020-04-24
Fix data zoom filter changes

Bluesmile82:
This PR fixes the data zoom problems when we changed the dropdowns on the GHG emissions.

- When we change the source: The Data zoom is reset and the user will have to change it again (This is needed as the years might not be the same)
- When we change other filters: The Data zoom handles are kept

image


kfilip94:
checking

PR #1045, opened 2020-04-27
Feature/png modal

weberjavi:
### Description
This PR creates a modal that displays the chart that will be later converted to `png` and downloaded. The conversion to `png` will be handled in another PR (as explained in this [Pivotal ticket](https://www.pivotaltracker.com/story/show/171697654))

image

### Designs
[InVision](https://projects.invisionapp.com/d/main#/console/18371781/381684669/preview)
### Testing instructions
Go to `localhost:3000//ghg-emissions` and check the `download .png` modal by clicking on the download data and save as image.
 
image

### Feature relevant tickets
[Pivotal ticket](https://www.pivotaltracker.com/story/show/171785471)


Bluesmile82:
Could we remove the plus button? This is meant for interaction and we won't have any on the PNG
image

Bluesmile82:
On small screens, the legend is cut and incomplete sometimes:
image

weberjavi:
@Bluesmile82 this is ready to review again ☝️ 
If you think it would make sense I could keep working in this branch to wrap up the implementation of the `download png` functionality. The only [task remaining](https://www.pivotaltracker.com/story/show/171697654) is implementing the create and download the png.

weberjavi:
This PR is ready to review ☝️ 
I've implemented the download functionality. There are a couple of clarifications that worth mentioning:
- Due to the way we are handling `svg` assets on `webpack` I had to add the CW logo as an inline `svg` on the modal (instead of `import`ing it) for it to be displayed on the resulting `png` (we might use an image instead, but I would need to ask the design team for it).
- I've been forced to use two different `npm` modules depending on the `userAgent`. The `html-to-image` returns a `dataURL` and does not work on Safari and Explorer so for those two browser we are using `html2canvas` that returns a `canvas` that needs to be converted to `dataURL`, there are some issues with this last module on mobile layout (probably due to the `scrollY` value)

app/javascript/app/utils/utils.js
```diff
@@ -105,9 +105,17 @@ export const sanitizeUrl = url =>
 // Detects if user browser is Edge or Explorer 11
 export const isMicrosoftBrowser = () => {
   const ua = navigator.userAgent;
-  return ua.indexOf('Edge') !== -1 || ua.indexOf('Trident/') !== -1;
+  return ua.indexOf('Edge') !== -1 || isInternetExplorer();
```
Bluesmile82, line 108: 👍

app/javascript/app/components/modal-png-download/modal-png-download-component.jsx
```diff
@@ -25,7 +23,17 @@ const modalPngDownloadComponent = ({
       header={<ModalHeader title={header} />}
     >
       <div ref={modalContentRef}>
-        <Icon className={styles.logo} icon={cwLogo} />
+        <svg
```
Bluesmile82, line 26: We cant use the Icon component because we have to add a class? What about importing it from the assets folder and adding a class afterward?

weberjavi: No, the reason is that through the `svg-sprite-loader` we are getting the logo from a sprite and the way it adds the `svg` to the DOM is through a [`#shadow-root`](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot) so the module that we are using to create the `png` can't access the `svg path` inside it

Bluesmile82: Ok, could you add this as a comment? Feel free to merge afterward

weberjavi: On the code you mean?

Bluesmile82: Yes, this is not so obvious at plain sight

PR #1050, opened 2020-04-27
Feature/compare all new endpoint

Bluesmile82:
Added data from new endpoint to populate the table and fixed styles

image

Try: Go to http://localhost:3000/compare-all-targets

### DIFF
--- a/app/javascript/app/pages/ndc-compare-all-targets/ndc-compare-all-targets-component.jsx
+++ b/app/javascript/app/pages/ndc-compare-all-targets/ndc-compare-all-targets-component.jsx
@@ -92,7 +92,7 @@ const NDCCompareAllTargets = props => {
             <div className={styles.buttonAndSearch}>
               <Button
                 variant="secondary"
-                className={styles.compareButton}
+                className={styles.actionButton}
                 disabled={selectedTargets.length === 0}
                 onClick={() => handleTargetsChange(null)}
               >
@@ -100,7 +100,7 @@ const NDCCompareAllTargets = props => {
               </Button>
               <Button
                 variant="primary"
-                className={styles.compareButton}
+                className={styles.actionButton}
                 disabled={selectedTargets.length === 0}
                 link={`/custom-compare?targets=${selectedTargets.join(',')}`}
               >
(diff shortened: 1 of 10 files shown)


Bluesmile82:
Apr-28-2020 10-42-09
I thought we have a problem with this accordion but probably is because we don't have content inside.

kfilip94:
checking

fannycc:
@kfilip94 i agree! please find here the designs updated with the clear button https://projects.invisionapp.com/d/main?origin=v7#/console/18371781/381684665/preview?scrollOffset=173 (it will be active when the user selects a first indicator). Regarding the right hand side of the table, that's made to be consistent with the rest of the tables we have on the site, with the exception that in this case that user can select indicators. The grey grid you see on the table is meant to be a "hint" of affordance but maybe it's enough if we show the selected status on hover?

Question, how many rows are we showing at once? maybe we can expand them a bit more? it rings a bell that they requested this already but I'm unsure. @Bluesmile82

Bluesmile82:
Hi @fannycc ! I don't understand the part about the selected on hover and the right hand side of the table. Yes, I will check if we can expand the height of the table. Add the clear button and see if I can fix the extra space on the width of the table

kfilip94:
FYI I added the clear button ticket to pivotal:
https://www.pivotaltracker.com/story/show/172554610
I'm also struggling with the "right hand side" part - not sure we're talking about the same thing, I meant removing the empty space on the right side
image

fannycc:
ok, let me explain this better :) The "extra space on the width" shouldn't be extra since the table has the same width size than the other tables on the Climate Watch website. The only difference here is that the "grey grid" you see surrounding the icons on the table was designed only for this table since it should show users that those squares will be colored (selected). 
My train of thought here is: The first design was taking into account that all indicator can be selected, but actually only the `submitted ones` can be selected. So I am debating between 2 options to avoid that feel of extra space on the right and I would like your opinion here :) 

1) Leave the grey border only around the selectable indicators (the submitted ones)
2) Remove all grey border and let users discover that they can be selected by hovering on top of them. The indicators will be colored on hover.

Please let me know if I did a better job now at communicating ☎️ 😄

Bluesmile82:
I added the Clear button and made the table a bit taller. 

We agreed that we could postpone or rethink the change that Fanny wanted to implement related to the way the user knows that the squares are clickable.

Also, for the table width I prefer to wait until we know the exact number of columns and then fix it.

kfilip94:
Looks great! ⭐

PR #1052, opened 2020-04-28
Feature/adjust accordion styles to compare all page

kfilip94:
This PR adds the accordion component and update the styles of the "Custom comparison page"
[PIVOTAL](https://www.pivotaltracker.com/story/show/170676458) | [DESIGN](https://projects.invisionapp.com/d/main?origin=v7#/console/18371781/381684667/preview?scrollOffset=600) | [LOCALHOST TEST](http://localhost:3000/custom-compare/overview)

image
- `CustomCompareAccordion` component has been created based on the `NDCsCountryAccordion` component,
- The data you can see there is the placeholder data, based on the data for `NDC comparison` page,
- I decided to mark the current version of the  `DefinitionsList` component as a legacy by changing the name of the folder and files to `definition-list-legacy***`. The new version is very similar to the old one but it required a few changes regarding styles and structure, so I kept that in the separate folder `definition-list`. The legacy version will be removed once we release the compare epic,
- The selected section changes how the accordion is rendered, so for `Sectoral Information` we have accordions with sub-accordions:
image


app/javascript/app/components/custom-compare-accordion/custom-compare-accordion.js
```diff
@@ -0,0 +1,81 @@
...
+      }
+    ]
+  },
+  {
+    slug: 'summary_information',
+    title: 'Title 2'
+  }
+];
+
+const CutomCompareAccordion = props => {
```
Bluesmile82, line 65: Typo here

app/javascript/app/components/definition-list-legacy/definition-list-legacy-styles.scss
```diff
@@ -0,0 +1,51 @@
+@import '~styles/layout.scss';
+
+.definitionCompare {
+  @include columns((3, 3, 3, 3));
+  @include xy-gutters(40px, $gutter-position: ('top', 'bottom'), $gutter-type: padding);
+  @include xy-gutters(60px);
+
+  border-bottom: solid 1px rgba(#9090b2, 0.15);
```
Bluesmile82, line 8: Can't we use one of the colors that we already have? In case of doubt please ping Fanny.

Bluesmile82:
Apr-28-2020 10-42-09
I thought we have a problem with this accordion but probably is because we don't have content inside.

app/javascript/app/components/custom-compare-accordion/custom-compare-accordion.js
kfilip94: 🤦  Nice catch, fixed!

app/javascript/app/components/definition-list-legacy/definition-list-legacy-styles.scss
kfilip94: This is a legacy component, it was like that since a beginning so I'm not sure if it's worth changing anything there since we're going to remove it anyway soon, what do you think?

Bluesmile82: Ah, sure I didn't realize it was de legacy component. Let it be

kfilip94:
> I thought we have a problem with this accordion but probably is because we don't have content inside.

exactly, I provided mockup data  only for `Title 1/Subtitle 1`

kfilip94:
Ok, I'm merging it, thanks for the review 👍

PR #1053, opened 2020-04-28
Enables comparison betweeen locations-documents

simaob:
This PR implements the changes to enable comparing NDC indicators between countries and different documents, it builds on the existing endpoint to enable a new parameter:

e.g.:
`locations_documents=ALB-first_ndc,DZA-indc`

This will be used instead of the current `location` parameter.

The route is then:

`/api/v1/ndcs?locations_documents=DZA-indc,ALB-first_ndc&category=overview&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer`

It expects a comma separated list of {iso-code3}-{document_slugs} pairs.

This endpoint doesn't yet include all of the data, as some bits are still work in progress, but they won't change the format of neither the request nor the response.

In regards to the laws and policies that is still missing overall, but we'll try to use the same endpoint for those as well. More on that later.

### DIFF
--- a/app/controllers/api/v1/ndcs_controller.rb
+++ b/app/controllers/api/v1/ndcs_controller.rb
@@ -145,9 +145,9 @@ module Api
       end
 
       def locations_documents
-        return nil unless params[:locations_docs].present?
+        return nil unless params[:locations_documents].present?
 
-        params[:locations_docs].split(',').map do |loc_doc|
+        params[:locations_documents].split(',').map do |loc_doc|
           loc_doc.split('-')
         end
       end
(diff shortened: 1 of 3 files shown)


simaob:
Note that I changed to `locations_documents` instead of `locations_docs` for consistency with the filtering using `document`.

kfilip94:
ok, checking

kfilip94:
Probably I'm doing something wrong, but once I open the request:
`http://localhost:3000/api/v1/ndcs?locations_documents=DZA-indc,ALB-first_ndc&category=overview&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer`
I get all documents data for all countries, so I suspect something goes wrong there? Or do I need to update something else?

simaob:
My bad! forgot to push the final change!

PR #1055, opened 2020-04-30
Feature/feed custom comparison page with data

kfilip94:
This PR feeds the `Custom comparison` page with the data from the backend. It allows filtering the data not only by country but also by document type:
[PIVOTAL](https://www.pivotaltracker.com/story/show/170676821)

image

- now URL for paths `custom-compare` and `compare-all-targets`  uses documents slugs instead of column titles, e.g.:
`targets=AND-first_ndc,AND-indc,AFG-first_ndc` 

**TO DO:**
- Use ` /api/v1/ndcs/countries_documents` endpoint for fetching metadata about documents (names, slugs etc.), instead of hardcoding slugs and names, for `Custom compare` and `Compare all targets` pages,
- `Custom-compare-accordion` selectors can be a bit messy, most of them come from NDC Compare page, it can be refactored in the future,
- Caching or keeping already downloaded data in redux could make the experience smoother and nicer for the user


Bluesmile82:
On small screens, the header columns are displaced with the data columns. Not sure if we should fix this. And definitely not now
image

kfilip94:
On mobiles, the header columns have a different layout, so the idea that the document data is displayed exactly below the document dropdown doesn't make much sense there. Maybe something to rethink @fannycc?
image

kfilip94:
Hey @Bluesmile82, I pushed a small fix regarding the positioning on the landscape resolutions, it's now better aligned:
image

app/javascript/app/components/definition-list/definition-list-styles.scss
```diff
@@ -2,11 +2,8 @@
 @import '~styles/layout.scss';
 
 .definitionCompare {
-  @include columns((3, 3, 3, 3));
-  @include xy-gutters(40px, $gutter-position: ('top', 'bottom'), $gutter-type: padding);
-  @include xy-gutters(60px);
-
-  margin: auto;
+  display: grid;
```
Bluesmile82, line 5: This last change is sadly breaking the IE layout. IE doesn't like grid. That is why we are using foundation. You will have to add some ms-grid attributes but I recommend reverting this change for now until we have a complete solution for small screens and mobile

image

PR #1056, opened 2020-04-30
Remove the FEATURE_NEW_GHG from the id to slug changes

Bluesmile82:
This PR Removes the FEATURE_NEW_GHG from the id to slug changes

Try: Go to the GHG emissions page with the flag on false. And try that the slugs are used instead of the ids on the URL when you change the source, gas and sector filters. Also on the country pages (GHG chart source)


simaob:
checking

PR #1059, opened 2020-05-04
Enable comparing indicators across different sources/types of documents

simaob:
This PR implements the comparison of indicators across Pledges, INDC and LTS. It does this by adding a `normalized_slug` and `normalized_label` to the Indc::Indicators, and when fetching the data finding only one of the normalized indicators and grouping the data across the documents under that one.

This doesn't require any change on the front end, as the endpoint doesn't change. 

To test it you should run:

```
bundle exec rails db:migrate
```

And then reimport the indc data from staging:

`bundle exec rails indc:import`

You should now be able to compare data from Pledges and First NDC for instance, or in the endpoint: http://localhost:3000/api/v1/ndcs?locations_documents=AUS-pledges,ALB-first_ndc&category=overview&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer , you should get this indicator: `nrm_summary`, which merges data from `indc_summary` and `pledges_summary`.

I have added some specs to test this out, and refactored the controller a little to tidy it up, and to only return indicators with data, so less data to be handled by the frontend.

### DIFF
--- a/app/controllers/api/v1/ndcs_controller.rb
+++ b/app/controllers/api/v1/ndcs_controller.rb
@@ -127,6 +127,12 @@ module Api
             where(locations: {iso_code3: locations_documents.map(&:first)},
                   indc_documents: {slug: locations_documents.map(&:second)})
         end
+
+        if params[:category]
+          parent = ::Indc::Category.includes(:category_type).
+            where(indc_category_types: {name: 'global'}, slug: params[:category])
+          indicators = indicators.joins(:categories).where(indc_categories: {parent_id: parent.map(&:id)})
+        end
         indicators.sort_by{|i| i.order}
       end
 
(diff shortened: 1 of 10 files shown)


kfilip94:
Hey, about the sectoral information data and `nrm_summary` - the categories ids that I see e.g. 224,
219 are not related to any category we fetching, so this will not be displayed. Do you think we should merge the categories too?
Check this endpoint:
http://localhost:3000/api/v1/ndcs?locations_documents=DZA-pledges,DZA-first_ndc,DZA-indc&category=sectoral_information&filter=overview&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer

kfilip94:
Also, I need to fix the sectoral information section, it copies one value for all selected documents from the same country:
image
also
image

simaob:
Hi Kamil it might be best to do some pair programming on this, as I'm not 100% sure I understand how the frontend is dealing with the data I'm passing to it. I have a meeting from 16h00 to 17h00. But we could chat at 17h00 if you are still around.

kfilip94:
sure :ok:

PR #1060, opened 2020-05-04
Feature/ghg new calculations

Bluesmile82:
This PR adds two new calculation options to the GHG chart:

- Cumulative over years: This calculation sums the value of the prior years to show the accumulated value. This is data zoom selection dependant, so if we move the start year for the data zoom this is recalculated

image

- Percentage Change: Each year presents the percentage of change compared to the previous year. The first year of data shows as blank in the chart and  'n/a' in the tooltip as it hasn't anything to compare too

image


simaob:
Looks really nice =)

kfilip94:
checking

PR #1063, opened 2020-05-05
Fix document options on ndc and full ndc documents

Bluesmile82:
This PR fixes the NDC country page document options. Before we had NDC and First NDC together and now we are getting these documents from the country documents endpoint.

Also allows the full document page to receive documents without language as the documents country documents don't have a language. Before we were defaulting to English and that is what we continue doing. The user can select other documents on the full-text page with the dropdown.

image

### DIFF
--- a/app/javascript/app/pages/ndc-country-full/ndc-country-full-selectors.js
+++ b/app/javascript/app/pages/ndc-country-full/ndc-country-full-selectors.js
@@ -1,12 +1,14 @@
 import { createSelector } from 'reselect';
 import uniqBy from 'lodash/uniqBy';
 
+const DEFAULT_LANGUAGE = 'EN';
+
 const getCountries = state => state.countries.data;
 const getSelected = state => state.document || null;
 const getContent = state => state.content || null;
 
 export const getCountry = createSelector(
-  [getCountries, (countries, iso) => iso],
+  [getCountries, (_, iso) => iso],
   (countries, iso) => countries.find(country => country.iso_code3 === iso)
 );
 
@@ -19,7 +21,7 @@ export const getSelectedContent = createSelector(
     return content.find(
       item =>
         item.document_type === splitSelected[0] &&
-        item.language === splitSelected[1],
+        (item.language === splitSelected[1] || DEFAULT_LANGUAGE),
       10
     );
   }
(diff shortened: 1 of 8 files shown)


kfilip94:
@simaob I tried the link  that you send and it seems to be working, but you need to wait a few minutes (!) to load, check if it works the same on your machine:
image

simaob:
Here it opened pretty quickly. And the dropdown works, the problem is that I don't think that the values on the interface are being updated. I changed the value of some indicators, but when I change the dropdown it doesn't seem to change.

Screenshot 2020-05-05 at 18 50 23

Bluesmile82:
I think these two calls return very similar data, I can't find a difference so it's noticeable on the frontend:

First NDC:
http://localhost:3000/api/v1/ndcs?location=BRA&category=overview&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer&document=first_ndc&filter=overview

INDC:
http://localhost:3000/api/v1/ndcs?location=BRA&category=overview&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer&document=indc&filter=overview

Bluesmile82:
> Here it opened pretty quickly. And the dropdown works, the problem is that I don't think that the values on the interface are being updated. I changed the value of some indicators, but when I change the dropdown it doesn't seem to change.
> 
> Screenshot 2020-05-05 at 18 50 23

The indicator call is filtered by document, you shouldn't get two types of documents on the same call

simaob:
The underlying data is quite similar, you could see differences if you run this in the rails console:

```
Indc::Value.joins(:location, :document, :indicator).where(locations: {iso_code3: 'BRA'}).where(indc_indicators: {slug: 'indc_summary'}).where(indc_documents: {slug: 'first_ndc'}).update_all(value: 'derp di doo')
```

This will update one of the values for Brazil and then you can test it on the interface!

Bluesmile82:
It's working for me, please be sure that you have the FEATURE_NDC_FILTERING on true

image

simaob:
I might've missed the flag, testing with that!

simaob:
You are absolutely right! I forgot the flag 🤦 it's working! Sorry about that

PR #1065, opened 2020-05-05
Revert lts-tracker embed route

kfilip94:
This tiny PR reverts the embed route for the LTS-tracker page. It seems to be working, I can't fully check it because I think I don't have data to display anything on the widgets there:
image


simaob:
👁️ 👁️

simaob:
I think you'd need to change some flags. I'll check it locally.

simaob:
It works with the production data, which is the key bit! Thanks!

PR #1067, opened 2020-05-05
Include total submitters on the countries_documents endpoint

simaob:
Adds a new attribute to the documents array with the total submitters, which can be used on the places where we need this value!

E.g.: http://localhost:3000/api/v1/ndcs/countries_documents?location=BRA


Bluesmile82:
Working fine! Thanks!

PR #1069, opened 2020-05-06
Fix/first ndc count

Bluesmile82:
The first NDC count on the overview and NDC explore pages was depending on the last submitted indicator but this was incorrect as countries could submit a second NDC but after submitting the first NDC. 

In this PR I use the countries documents endpoint to get the isos of the countries with the first NDC submitted and calculate the different emissions and numbers in Overview and NDC emissions page on the summary card.

Overview page (first section - second question):

image

NDC explore page (Summary card):

image


weberjavi:
:eye: :pig_nose: :eye:

PR #1076, opened 2020-05-07
Restore ndc documents for production

Bluesmile82:
This PR recovers the previous implementation of the document options and selection on NDC country page. It was not actually hardcoded but coming from NDC documents meta provider.
Is dependant on the FEATURE_NDC_FILTERING so it should be on for staging and local and off for production


simaob:
It crashes with the updated data from staging when I go to this page: http://localhost:3000/ndcs/country/BRA

Bluesmile82:
I was missing a guard. Please check again

PR #1077, opened 2020-05-08
Release/fixed release docs

simaob:


Bluesmile82:
We already had a later release probably we should close this one

PR #1079, opened 2020-05-08
Add more points to map

Bluesmile82:
The client asked to make Singapore and other little countries more visible. Check the feedback document for more info. AJ did the rest:

image
image


weberjavi:
Should we tag this as Work in Progress while the design bit of it is being discussed?

Bluesmile82:
This one has green light!

PR #1082, opened 2020-05-11
Feature/ignore Western Sahara on the ndcs-/lts- explorer map

kfilip94:
This PR blocks Western Sahara on the `ndcs`- and `lts-explore` pages - the tooltip doesn't appear, and the click doesn't open the country page
[PIVOTAL](https://www.pivotaltracker.com/story/show/172762378)
5uvw1-2iqj1
**Test**: Go to the [lts-explore](http://localhost:3000/lts-explore) and [ndcs-explore](http://localhost:3000/ndcs-explore) pages and try to click and hover actions on Western  Sahara and Taiwan.
***
It also changes the `LTS explorer` description according to the feedback - replaces `"GHG"` with `"greenhouse gas"`
[PIVOTAL](https://www.pivotaltracker.com/story/show/172762551)
image
**Test**: Check the description on the  [lts-explore](http://localhost:3000/lts-explore)  page.
***
I also removed the `country-codes.js` file from the `data` folder - we weren't using those constants anywhere.


app/javascript/app/data/ignored-countries.js
```diff
@@ -0,0 +1,2 @@
+// We won't show Taiwan and Western Sahara as an independent country
+export const IGNORED_COUNTRIES_ISOS = ['TWN', 'ESH'];
```
Bluesmile82, line 2: This could actually live in constants.js but it's ok, it's more tragic this way.

kfilip94: Yes, maybe having a separate file is a bit too much :) I was thinking where to put this const, but `constant.js` seems to be very "crowded" to me, we should organize the constants in a more logical way

PR #1084, opened 2020-05-12
Adds indicator_ids to sectors object on the ndcs endpoint

simaob:
This PR adds indicator_ids to the sectors objects on the NDC api call. As per this [PR task](https://www.pivotaltracker.com/story/show/172764124).

Example URL: http://localhost:3000/api/v1/ndcs?location=CAN&category=summary&source=CAIT&filter=overview

I'm now thinking that maybe I should filter the sectors per the other filters passed.

### DIFF
--- a/app/controllers/api/v1/ndcs_controller.rb
+++ b/app/controllers/api/v1/ndcs_controller.rb
@@ -41,7 +41,7 @@ module Api
 
     class NdcsController < ApiController
       def index
-        sectors = ::Indc::Sector.all
+        sectors = ::Indc::Sector.includes(values: :indicator)
         indicators = filtered_indicators
         categories = filtered_categories(indicators)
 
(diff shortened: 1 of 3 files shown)


simaob:
Regarding the indicators repeating per sector, that's how the data is. The same country might "answer" the same questions in relation to each of the sectors it answers to. So the indicators are not sector specific.

simaob:
You couldn't see data for categories and indicators because that URL was not a valid one, try this one: http://localhost:3000/api/v1/ndcs?location=CAN&category=mitigation&source=CAIT&filter=overview

I will filter the sectors data and push it soon.

simaob:
Added filtering of sectors. It will still include some sectors without indicators, but those are parent sectors that we use in the labels on the NDC country page.

simaob:
Because we are now filtering sectors, they won't show up in all the calls, so one to test is this one: http://localhost:3000/api/v1/ndcs?location=BRA&category=sectoral_information&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer&document=first_ndc&filter=overview

PR #1085, opened 2020-05-12
Tweak country-ghg chart values rounding

kfilip94:
This PR changes the rounding of the  `country-ghg` chart values  - now they always have two decimal digits for stacked-chart metrics (`Cumulative across...`, `Percentage change...` and `Total`).
[PIVOTAL](https://www.pivotaltracker.com/story/show/172721275)

I added new format features to the `TooltipChart` component:
- `customD3Format` -  custom d3 format, before it was the number of significant digits
- `customFormatFunction` - custom format function,

Also, I added a new utils file `utils/d3-custom-format.js` with `formatSIwithDecimals`  function. It returns a value with an  SI prefix and with a fixed number of decimal digits, which was not possible using a regular format.

image

### DIFF
--- a/app/javascript/app/components/charts/line/line-component.jsx
+++ b/app/javascript/app/components/charts/line/line-component.jsx
@@ -93,7 +93,7 @@ class ChartLine extends PureComponent {
       height,
       margin,
       domain,
-      forceFixedFormatDecimals,
+      customD3Format,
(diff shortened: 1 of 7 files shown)


kfilip94:
Yes, let's wait for the feedback 👍 Thanks for the review!

PR #1087, opened 2020-05-14
Filter out "Cumulative" calculation option when UNFCCC Non-Annex is selected

kfilip94:
This small PR filters out the "Cumulative..."  calculation option when `UNFCCC Non-Annex I` source is selected. This change is required by the client, check the pivotal tracker ticket:
[PIVOTAL](https://www.pivotaltracker.com/story/show/172815461)
image

### DIFF
--- a/app/javascript/app/components/ghg-emissions/ghg-emissions-selectors/ghg-emissions-selectors-filters.js
+++ b/app/javascript/app/components/ghg-emissions/ghg-emissions-selectors/ghg-emissions-selectors-filters.js
@@ -115,10 +115,29 @@ const getBreakByOptions = () =>
       }
     ]);
 
+// Filtered calculation selectors
+const getFilteredCalculationOptions = createSelector(
+  [getCalculationOptions, getSourceSelected],
+  (calculationOptions, sourceSelected) => {
+    if (!calculationOptions || !sourceSelected) return null;
+    if (sourceSelected.name === 'UNFCCC_NAI') {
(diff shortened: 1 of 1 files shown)


app/javascript/app/components/ghg-emissions/ghg-emissions-selectors/ghg-emissions-selectors-filters.js
```diff
@@ -115,23 +115,40 @@ const getBreakByOptions = () =>
       }
     ]);
 
+// Filtered calculation selectors
+const getFilteredCalculationOptions = createSelector(
+  [getCalculationOptions, getSourceSelected],
+  (calculationOptions, sourceSelected) => {
+    if (!calculationOptions || !sourceSelected) return null;
+    if (sourceSelected.name === 'UNFCCC_NAI') {
+      return calculationOptions.filter(({ value }) => value !== 'CUMULATIVE');
```
Bluesmile82, line 124: Could we use the constant here as: CALCULATION_OPTIONS.CUMULATIVE.value ?

kfilip94: Good catch! Fixed and ready to merge!

Bluesmile82: There are going to be some conflicts with my next PR

PR #1095, opened 2020-05-20
Feature/country lts section

Bluesmile82:
This is the new LTS section for the country page.
It has:

- Title, similar to the NDCs section
- Content of the LTS summary of the LTS page
- Info button opening metadata for LTS
- Compare button linking to the new compare-all page with the LTS document selected
- Share button with a new LTS summary embed
- Primary explore button leading to the LTS page

This section doesn't show if the LTS doesn't have content

image

### DIFF
--- a/app/javascript/app/components/button/button-styles.scss
+++ b/app/javascript/app/components/button/button-styles.scss
@@ -88,7 +88,7 @@
 
 .v-secondary {
   background-color: $white;
-  border: solid 2.5px $gray2;
+  border: solid 2.5px $gray3;
 
   &:hover {
     background-color: $gray3;
--- a/app/javascript/app/components/country/country-lts-overview/country-lts-overview-component.jsx
+++ b/app/javascript/app/components/country/country-lts-overview/country-lts-overview-component.jsx
@@ -127,59 +127,55 @@ const CountryLtsOverview = props => {
     />
   );
 
-  const renderNoContent = () =>
-    (isCountryPage ? null : (
-      <NoContent message="No data" className={styles.noContentWrapper} />
-    ));
   return (
     <div
       className={cx(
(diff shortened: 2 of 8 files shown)


kfilip94:
checking

Bluesmile82:
@fannycc The style is here. Maybe we merge that gray with the darker one. Is that what you don't like? 🦆  image

fannycc:
Screenshot 2020-05-20 at 13 26 05
yeah, those are correct. The compare has a different style tho :) not the secondary style.

kfilip94:
- In the embed view of this component we don't see any information about the country that we selected, how we should deal with that @fannycc?
image
the same for the ndc section:
image
- How about the case when a country doesn't have LTS submitted, but the `LTS Content Overview` section is still visible:
x81cg-utywo

Bluesmile82:
Yes, maybe we should remove also that item from the navbar if the LTS doesn't have data. But we can't know it until we do the fetch. Should we show always an LTS section without content? Or do you prefer to remove it completely?

fannycc:
hey, so reading you and without further details I guess that it doesn't make much sense to select the embed section for something that can't be embed?. Or the problem is that when I embed the LTS of let's say Algeria in the embed it doesn't say the `country name` ? 
And about the overview shown in a selected country, i guess we could put a placeholder like: This country hasn't submitted any Long Term Strategies, that's also useful information. Please feel free to ping me during the day in slack about this if necessary

kfilip94:
> Or the problem is that when I embed the LTS of let's say Algeria in the embed it doesn't say the country name

☝️ this one ^^

About the `LTS section` it makes sense to show the message about lack of LTS document, totally agree 👍

Bluesmile82:
I added the new text to the No content countries, on the country and Lts page. 
Now the section always shows
Also, I changed the secondary button border color.

image
image

PR #1097, opened 2020-05-21
Links indc_submissions with indc_documents

simaob:
Includes submission date on the /api/v1/ndcs/countries_documents?location=USA
endpoint.

To test you need to run:

```
bundle exec rails db:migrate
bundle exec rails indc:import
```

When hitting this endpoint:

`/api/v1/ndcs/countries_documents?location=USA` , with our without filter you will get the submission date for each of the countries document, on the `data` object.


simaob:
This gives the date we need to show on the NDC Explorer Country page.

simaob:
Don't forget about this PR =)

PR #1100, opened 2020-05-26
Feature/integrate laws target data

simaob:
This PR brings Climate Laws and Policies information to the `http://localhost:3000/api/v1/ndcs/countries_documents` endpoint, to be used on the Compare All Targets page, as well as the comparison page.

On the http://localhost:3000/api/v1/ndcs/countries_documents main endpoint the response will include new information on the data block, that tells us if there are targets in Laws and Policies for each country, e.g.:

Screenshot 2020-05-26 at 10 58 12

These will come always at the bottom of that data object, and if one of them is not included be it Laws or Policies it means that there aren't any targets on those for that country.

Then on the filtered endpoint: http://localhost:3000/api/v1/ndcs/countries_documents?location=USA, we include information about the Laws and Policies of a given country, e.g.:

Screenshot 2020-05-26 at 10 59 42

This can then be used on the documents dropdown of the comparison page. The laws and policies are only included when filtering by a location, to keep the response smaller (otherwise there would be 2k laws and policies).

Let me know if this works for the front end integration.

The next step will be to fetch the data for the accordions. For that I'm wondering how to pass to the backend the pair of country-law/policy . Wondering if we should use something like: `USA-law-3834` or just `USA-3834`. Would this work for you guys?

Let me know!

### DIFF
--- a/app/controllers/api/v1/lse_laws_and_policies_controller.rb
+++ b/app/controllers/api/v1/lse_laws_and_policies_controller.rb
@@ -6,7 +6,7 @@ module Api
       def show
         id = params[:id] == 'EUU' ? 'EUR' : params[:id]
 
-        laws_and_policies = SingleRecordFetcher.new(LSE_API, id).call
+        laws_and_policies = SingleRecordFetcher.new(LSE_API, id, id).call
         render json: laws_and_policies
       end
     end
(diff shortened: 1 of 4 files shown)


simaob:
Yes, I think we should select the first one on the list, for lack of a better way to pick.
OK I'll go with those keys. And I'll inject the data into the other response anyway, so that it shows alongside the rest.

I will work on that bit this afternoon.

Going to resolve the conflicts and merge this one now.

PR #1103, opened 2020-05-27
Adds slug to api/v1/data endpoints resources

simaob:


kfilip94:
Hey @simaob, thanks a lot for adding this. There are things I found:
- **data-sources** slugs are always `null`, can you confirm that you see the same problem?
image

- I don't see slugs for Pathways filters

simaob:
OK on data sources I'm seeing it as well, though not sure why. I'll check it after my next meeting.
For Pathways I need to do that on another project and then deploy it. It's a separate application.

simaob:
OK, the source slug should now be fixed! =)

simaob:
Pathways is also fixed.

PR #1112, opened 2020-06-03
Feature/data explorer use slugs in url

kfilip94:
This PR updates the date-explorer page by replacing the ids with the slugs in the URL. This is the first PR of this feature, second PR is going to update all download links across the platform.
[PIVOTAL](https://www.pivotaltracker.com/story/show/172835211)

**Test:** Check all sections and filters on the [data-explorer page](http://localhost:3000/data-explorer/historical-emissions).

### DIFF
--- a/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
+++ b/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
@@ -72,7 +72,6 @@ const getMetaForNoModelFilters = createSelector(
         slug: kebabCase(v)
       }));
     });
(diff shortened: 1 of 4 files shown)


Bluesmile82:
Great work with this! Could you check the gases on the historical emissions? I think they still have an id. Also, on the NDC-SDG Linkages section I can't see any data. This could be related to these changes as I see the data on staging.

kfilip94:
Hey @Bluesmile82, thanks for the review, gases should use slugs now. I've just reimported all the data and NDC-SDG works, so the issue was/is probably on the data side. Ready again ☝️

PR #1123, opened 2020-06-17
Adds ghg_sources information to regions endpoint

simaob:
As discussed in this story: https://www.pivotaltracker.com/story/show/173197745
I've extended the existing regions endpoint to include information about the GHG emissions sources available for each country.

To avoid cluttering the endpoint unnecessarily I thought it would be best to use a param to request this, so I added `ghg_sources=true` as the param for this, to be appended to the endpoint. Let me know if this makes sense or if it's best to include it all the time?

thanks!

### DIFF
--- a/app/controllers/api/v1/locations/regions_controller.rb
+++ b/app/controllers/api/v1/locations/regions_controller.rb
@@ -10,7 +10,8 @@ module Api
 
           render json: regions,
                  each_serializer: Api::V1::LocationSerializer,
-                 topojson: params.key?(:topojson)
+                 topojson: params.key?(:topojson),
+                 ghg_sources: params.key?(:ghg_sources)
         end
       end
     end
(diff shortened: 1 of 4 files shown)


Bluesmile82:
Some regions have empty GHG sources. I'm not sure what this actually means. Is this something missing?
image

simaob:
This means we don't have GHG data for that region. Should I check if we have for the region's members?

Bluesmile82:
So I mean if we don't have data for that region we won't show them on the dropdown but I think we are doing it. Is this different depending on the region?

simaob:
Maybe we are calculating the region data from its members, when there isn't a specific row for that region. In the example of AILAC, there seems to be data in prod: https://www.climatewatchdata.org/ghg-emissions?regions=AILAC%2CBRICS, but it might just be the countries data being shown.

I can check if there is data for the member countries and show those sources for the regions. So if some of the countries members of AILAC have data for CAIT we include CAIT with AILAC.

Bluesmile82:
I think that makes sense, as is just an association of countries in this case. Thanks!

simaob:
Updated!

PR #1127, opened 2020-06-18
show the scope value instead of sector as it has more details

simaob:
This is a simple change but fixes the confusion found here: https://basecamp.com/1756858/projects/13795275/todos/418613657 , or at least part of it=)

It updates the country page lse block:

http://localhost:3000/countries/IND?sector=energy#laws-and-policies

Screenshot 2020-06-18 at 12 00 25


kfilip94:
Not strictly related to the changes, but the `Targets in Submitted NDC` link doesn't work:
image
It redirects to `http://localhost:3000/ndcs/country/null/`

Bluesmile82:
image
I can see some strange chars in Canada. I'm going to check why the parsing fails

PR #1129, opened 2020-06-21
Change order of 'Others' selection (calculation => sort => others)

Bluesmile82:
This PR solves two problems on the GHG chart:

- Fixes cumulative calculation chart. There was a clause missing for dataZoomYears.min
image

- More importantly, changes the order in which we calculate the 'Others' grouped values in the chart. 

Before, we grouped the values out of the 10 items into an 'Others' item, then sorted them and then did the calculations (like per capita).

Now, we do the calculations, then we sort the values and then we group the Others. This fixes the selection of shown countries when we have other calculations than 'Total' selected.

image

- Also, the legend selector is fixed so it has the same functionality as the dropdown. If we have an expanded region and we add a country or region it will contract the region and this prevents any 'Others' problem to happen. Previously the top 10 countries were added to the selection and this was really unexpected for the user.

### DIFF
--- a/app/javascript/app/components/ghg-emissions/ghg-emissions-component.jsx
+++ b/app/javascript/app/components/ghg-emissions/ghg-emissions-component.jsx
@@ -168,6 +168,22 @@ function GhgEmissions(props) {
     );
   };
 
+  const handleLegendChange = v => {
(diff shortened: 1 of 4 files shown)


simaob:
Looks good to me, function wise!=)

PR #1130, opened 2020-06-23
Fix/description links

Bluesmile82:
Added description links linking to the same page with some params. I styled them with some hover

image


simaob:
Coolio! =)

kfilip94:
Hey, can we do something about the default tooltips here? Right now, after hover, they show only the first word of the link text:
image

Bluesmile82:
This is the title attribute. I will check it later

Bluesmile82:
This PR is updated with the correct title links now. I have added a bit more of text to make the sentences readable

PR #1131, opened 2020-06-23
Feature/integrate framework sectoral data with custom compare filters

kfilip94:
This PR adds `sectoral` and `framework` document filters to the `custom-compare` page

**Testing:** Try to play with different filters (countries and documents) between [Custom compare](http://localhost:3000/custom-compare) and [Compare all targets](http://localhost:3000/compare-all-targets) pages

The main challenge here was to sync up the URL params between `custom-compare` and `compare-all-targets` pages:

~~the `target` URL param format has changed  from:
**`<iso>-<document-slug>`**
to:
**`<iso>-<document-slug> || <iso>-<group>-<document-slug>`**
where **`group`** is the slug of the documents group , in this case `sectoral` and `framework`~~

Navigating between `custom-compare` and `compare-all-targets` pages should work without problems.

@simaob, there is one thing that could be fixed on the backend - could you update the slugs of all `sectoral` and `framework` documents with the group slug?
e.g. from `very-nice-document` to `sectoral-very-nice-document` /  `framework-very-nice-document`?
The problem is that these titles => slugs are not unique, so sometimes it happens that for the same country we have two documents with the same slug for different categories:
image


simaob:
Hey Kamil, thanks for this. yeah I can update those slugs. Is it OK if I do as we discussed and update the slugs to be actually: `sectoral-LAW_ID` and `framework-LAW_ID`. then you can just use the same flow to pass this to the back to get the results. What do you think?

simaob:
can I change this on this branch? :D

kfilip94:
Yes, do it ;)

kfilip94:
I'll update the frontend code if needed

simaob:
done!

kfilip94:
I've made some quick tweaks for the new document slug format, ready for review!

PR #1132, opened 2020-06-23
Feature/compare intends to submit

Bluesmile82:
Add intends to submit icon when we have a second ndc document but no submitted_date

image


simaob:
Noice! =)

PR #1138, opened 2020-06-25
Fix unit and scale depending on calculations

Bluesmile82:
Cumulative calculations: change unit on the table to MtCO2 and scale accordingly

http://localhost:3000/ghg-emissions?calculation=CUMULATIVE
image

Percentage calculations: change unit on the table to %

image

### DIFF
--- a/app/javascript/app/components/table/table-component.jsx
+++ b/app/javascript/app/components/table/table-component.jsx
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react';
+import React, { useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { Column, Table, AutoSizer, ScrollSync } from 'react-virtualized';
 import MultiSelect from 'components/multiselect';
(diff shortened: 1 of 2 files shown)


Bluesmile82:
I added a scroll to the last year and changed the Table component to hooks

app/javascript/app/components/table/table-component.jsx
```diff
@@ -11,192 +11,203 @@ import styles from './table-styles.scss';
...
+    firstColumnHeaders,
+    flexGrow,
+    splittedColumns,
+    theme,
+    titleLinks
+  } = props;
+  const rightTable = useRef();
+  useEffect(() => {
+    rightTable.current.scrollLeft = rightTable.current.scrollWidth;
+  }, rightTable.current);
```
kfilip94, line 42: Hey, there is a warning in the console:
image
shouldn't the dependency prop be inside an array?
```suggestion
  }, [rightTable.current]);
```

Bluesmile82:
I'm wondering if I should check if rightTable exists too but it should work like this

PR #1145, opened 2020-06-30
Sort explore-ndcs/-lts table data on load

kfilip94:
This small PR fixes the problem with the initial sorting of the data table on `explore-ndcs/ -lts` pages.
[PIVOTAL](https://www.pivotaltracker.com/story/show/173564731)

The sorting feature works well on the `Table` component, the thing is that there is no initial sorting after data the fetch.
The data is sorted by `iso` on the backend, so maybe that would be something we could change in the future @simaob? Like to sort by the data name of the country?

**Test:** Go to [ndc-explore](http://localhost:3000/ndcs-explore) and [lts-explore](http://localhost:3000/lts-explore) pages, and check if the rows are sorted by `country` column after first load. Also, check if manual sorting works ok.


simaob:
What's the endpoint being used here? `/ndcs` ?

PR #1146, opened 2020-07-01
Fix/overview use slugs

Bluesmile82:
- Update texts
- Capitalize 'Parties'
- Request only selected indicators for the Overview page.

This last one has a problem. Now that we fetch only selected indicators (and categories on the NDC explore page) we can't preserve the NDC data on the store as we did before. So we will have to find some other solution, either some kind of cache or checking if we have the data we need before fetching.

After this is merged for example on NDC content it will fetch always the NDC data no matter if we have got the data before


Bluesmile82:
Its actually [Solve NDC fetching - caching](https://www.pivotaltracker.com/story/show/173609684)
We could use some cache solution like https://github.com/RasCarlito/axios-cache-adapter or https://github.com/tannerlinsley/react-query

PR #1147, opened 2020-07-01
Always pick the last document

Bluesmile82:
This should fix a problem in NDC country page where if we don't have a selected document we didn't pick the last one. Now we should


simaob:
This seems to work in principle, but now I'm not sure if you are filtering the first page correctly, as the content_overview request is returning info for all the documents of a country.

Check this PR https://github.com/Vizzuality/climate-watch/pull/1150 as it can help fix this.

simaob:
Though ideally you'd filter the request, as the sectors object in that response would need to change format, to allow us to include the document_slug.

PR #1149, opened 2020-07-01
Order donut chart data by value

Bluesmile82:
This little PR orders the donut chart data on the NDC Explore and LTS Explore pages
Before:
image
After:
image


kfilip94:
I've just noticed, it breaks the relation between hovered country and highlighted part of the pie-chart, check the gif below;
oimtz-csn0o

Bluesmile82:
This is ready again. I fixed the interaction. Please check it again!

PR #1150, opened 2020-07-02
include document_slug for ndc_content overview

simaob:
I realised that we are not including the document_slug for the content_overview endpoints, and that is needed to filter the Summary pages.

Just to remind you that you could also filter the request by document, see: http://localhost:3000/api/v1/ndcs/ARG/content_overview?document=indc


simaob:
I think we should use that document filter on The country page, as we are fetching everything for content_overview, so we might not be including the latest document. If you don't have the document slugs available, maybe we could have a filter called `for_latest_document=true`, and let the backend send you the right info.

PR #1152, opened 2020-07-02
Adds filtering by subcategory, and improves the filtering code a bit

simaob:
You can now filter the `/api/v1/ndcs` endpoint with an extra param: `subcategory`.
The goal is to load less data each time on the NDC Content/NDC Explorer page.

Example: http://localhost:3000/api/v1/ndcs?location=BRA&category=overview&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer&document=first_ndc&filter=overview&subcategory=ndc

It will still return all of the subcategories under the `category` being passed to the endpoint, so that you can fill the categories dropdown. =)

We'll need to run a few tests once this is integrated into the flow, to ensure that we are not leaving data behind. I hope we aren't =)

### DIFF
--- a/app/controllers/api/v1/ndcs_controller.rb
+++ b/app/controllers/api/v1/ndcs_controller.rb
@@ -55,10 +55,15 @@ module Api
       before_action :set_locations_documents, only: [:index]
 
       def index
-        indicators = filtered_indicators
-        categories = filtered_categories(indicators)
+        # params[:source] -> one of ["CAIT", "LTS", "WB", "NDC Explorer", "Pledges"]
+        if params[:source].present?
+          source = ::Indc::Source.where(name: params[:source])
+        end
+        indicators = filtered_indicators(source)
+        categories = filtered_categories(source)
         sectors = ::Indc::Sector.joins(values: :indicator).
           where(indc_indicators: {id: indicators.map(&:id)})
+
         parents = ::Indc::Sector.where(parent_id: nil).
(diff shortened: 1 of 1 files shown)


kfilip94:
Hey, I was looking on this today and I''ve a question.

Now after entering the `NDCS Explore` page, we fetch all data from the endpoint:
**`/api/v1/ndcs?filter=map&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer`**

After changes from this PR we can pass `subcategory` parameter, which is a `Category` filter from the NDCS Explorer map (a bit confusing I must admit), so the endpoint can look like that:
**`/api/v1/ndcs?filter=map&source[]=CAIT&source[]=WB&source[]=NDC%20Explorer&subcategory=finance_and_support`**

The thing is, if we don't have selected category, how do I know which one should I select? Because I need to have one to pass it to the endpoint, as `subcategory` param, at this moment I don't have any data about categories available, because I need to fetch it first, which is a classic the egg or the chicken problem. Could you clarify this, it's possible that I'm missing something

Bluesmile82:
Won't we have a default category? If we don't the endpoint should return the default or first one if we pass a special parameter ('default'? maybe not the best)

simaob:
You're right, Kamil. One option would be to hardcode the first subcategory. Which should be OK, because we do have a default selection that categories dropdown: UNFCCC Process.

So if we use that as the default, we should be good?

kfilip94:
yes, this should solve the issue :+1: 
@Bluesmile82 If we don't pass category explicitly, we will fetch all data as we're fetching now, so it's not a benefit. We can fetch categories first, by specifying the `subcategory=null`, but then we need to do another fetch for data with the selected category

PR #1153, opened 2020-07-06
Sets a default for document slug

simaob:
This PR aims to reduce the size of the response of the `/api/v1/ndcs` calls. We already had the possibility of filtering these calls by document, by passing `document=indc` as a parameter to the call. But when the parameter was not passed it was defaulting to all of the values. This PR sets a default to `first_ndc`. By doing this it reduced the response size by half, so using it alongside of the `subcategory` can have a really positive impact in performance.

But this change has some caveats, as by setting a default it means that where we were relying on the whole data being available, it no longer is, so we need to test it carefully.

The `document` filter supports multiple values, so where needed we could pass a comma separated list of document slugs: e.g.: `document=first_ndc,indc` .

One thing still outstanding related with these document slugs is the first page of the NDC Country page: http://localhost:3000/ndcs/country/BRA/?document=first_ndc, we are not passing the document slug to the `content_overview` endpoint, so we are always displaying the same data. We should update it to call the dataset with the selected document_slug. 

Álvaro if you feel that this would require more information or is too much trouble, we can also ignore it!

### DIFF
--- a/app/controllers/api/v1/ndcs_controller.rb
+++ b/app/controllers/api/v1/ndcs_controller.rb
@@ -53,6 +53,7 @@ module Api
 
     class NdcsController < ApiController
       before_action :set_locations_documents, only: [:index]
+      before_action :set_document
 
       def index
         # params[:source] -> one of ["CAIT", "LTS", "WB", "NDC Explorer", "Pledges"]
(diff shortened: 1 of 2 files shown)


Bluesmile82:
I think we should leave this improvement for later if it doesn't fix some bugs. It might need some changes

simaob:
Makes sense! Added labels to reflect that! =)

simaob:
this PR no longer makes sense. deleting it.

PR #1154, opened 2020-07-07
Feature/fetch ndcs by category

Bluesmile82:
On NDC explore page we used to fetch all NDC data and now we filter by category.

- The default category is hardcoded in data/constants 
- The NDC store now doesn't delete data only stores new unique indicators by id
- This functionality is not copied to LTS as we don't have much data on the initial fetch
- A cache needs to be implemented to avoid double calls and requesting again the same data (We have a task on PT for this)
- An extra call to pick the emissions indicator is made with the prop additionalIndicatorSlug
- There are some categories that still have 2+ MB so the call is still slow

image

Extra: Update footer sitemap

image


simaob:
checking!

Bluesmile82:
Oh, I will look at that. It should be fixed already

Bluesmile82:
Ready again!

simaob:
Nice! =)

PR #1155, opened 2020-07-07
Update text and make filters sticky

Bluesmile82:
Update text

image

Sticky filters on custom compare

image


Bluesmile82:
It's actually the same size. Maybe some optical effect

PR #1162, opened 2020-07-09
Fix/map categories

Bluesmile82:
This PR fixes the problem with storing all the NDC data on the store without deleting it. 
Now we filter the categories that we want to display on NDC Explore and NDC content. By doing this the flow is much clearer. 
We might want to include at some point the LTS data into the NDC storage because is actually just another source.

Try: Go to overview => go to NDC Explore or NDC content => It should show only the correct categories

Note: We seem to be displaying 'NDC Explore' source categories on the NDC content. Not sure if this is a bug. It can be solved easily now.


simaob:
Going to check this now. But that is not a bug, the "NDC Explore" is just another source, and it was OK to show on the NDC Content.

PR #1169, opened 2020-07-13
Feature/ndcs cache

Bluesmile82:
This PR uses 'axios-cache-adapter' to create a new service on the existing api service file and use it in most of the ndcs pages.
This package stores the cache for the request so once we don't have to repeat them. 
This improvement doesn't help with the first fetch but will create a better experience after that.
Try: NDCS Explore => fetch a category go to another and back. It should load instantly.

Also, add an info button to compare all with metadata:

image
image

Also, updated some text on the overview page according to feedback


Bluesmile82:
I'm merging this PR as is not actually about the nested-dropdown. Let's make the fixes in some other

PR #1170, opened 2020-07-13
Adds caching to the /api/v1/ndcs endpoint

simaob:
This PR adds caching to the /api/v1/ndcs, it also removes some includes clauses that was slowing things down. So overall should be quicker!

To test the caching locally you'll need to run `bundle exec rails dev:cache` and to turn that off just use the same command again. =)


simaob:
let me know if you see missing data or some weird behaviour, I use the params as the cache key.

PR #1171, opened 2020-07-13
Feature/data explorer url

kfilip94:
This PR merges the updated URL structure of the data explorer (slugs instead of ids) to develop. Changes had been approved in the [previous PR](https://github.com/Vizzuality/climate-watch/pull/1121), so here I rebased the codebase with recent develop changes.

**Test:** Please check if data-explorer filters and download links across the app works

### DIFF
--- a/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
+++ b/app/javascript/app/components/data-explorer-content/data-explorer-content-selectors.js
@@ -145,8 +145,8 @@ const findSelectedValueObject = (meta, selectedId) =>
     option =>
       option.iso_code === selectedId ||
       option.iso_code3 === selectedId ||
+      option.slug === selectedId ||
       option.name === selectedId ||
-      String(option.slug) === selectedId ||
       String(option.id) === selectedId
   );
 const addTopEmittersMembers = (isosArray, regions, key) => {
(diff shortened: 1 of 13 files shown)


Bluesmile82:
Ah, no. Sorry, I was thinking of the GHG emissions. Its fine

Bluesmile82:
We still don't get the gases going from and to the GHG emissions page

kfilip94:
@Bluesmile82, yes, it's the first point that I've described [here](https://github.com/Vizzuality/climate-watch/pull/1121#issuecomment-645956464).
I'd leave it for now and take care of this issue in another PR, because it's related to GHG emissions page and the way how we deal with slugs there - we need to replace the `kebabCase` method with backend slugs

Bluesmile82:
Pathways is also failing 😢

Bluesmile82:
NDC-SDG linkages is not showing data on the Data explorer page. I remember this was a problem before. Maybe the merge?

Bluesmile82:
Also, the indicators and categories passed to NDC and LTS explorer on 'Visualize in ...'  are passed as ids and they don't get selected

Bluesmile82:
For the slugs, maybe we could use something like this function: https://github.com/stevenpersia/captain-hook/blob/master/useSlug.js

kfilip94:
Hey, did you fetch & pull the latest changes and/or import the data? I can't reproduce any of these issues on my machine o.O

Bluesmile82:
Ok, now NDC-SDG works. 

Some links from Pathways are failing http://localhost:3000/pathways/models?category=10&currentLocation=267&indicator=388&model=3&scenario=180%2C182%2C181%2C183&subcategory=36 
Maybe let's add this on a new task

Also, there are indicators in NDC and LTS that are not part of the map so the links don't work. But this makes sense. Maybe we have to add a message at some point.

Let's merge so we don't have more rebase problema and you can continue working on these tasks

PR #1176, opened 2020-07-16
Adds link to Climate Laws at end of each target for nrm_summary

simaob:
This PR adds a "View on Climate Laws" link to the end of the LSE related targets' summary data.

You can see it in this example: http://localhost:3000/custom-compare/overview?section=ndc&targets=DZA-framework_1002%2CDZA-first_ndc

Screenshot 2020-07-16 at 10 37 33


simaob:
So sometimes it might not go to the same page 😅 , but I could make it a little more fancy and check first. But maybe for a first version could be like this.

simaob:
Ohhhh... actually no! you're right, it will always be the same link. Let me fix this.

PR #1177, opened 2020-07-16
Add LSE data to Overview with new provider

Bluesmile82:
On the Overview page for the last question we had to import some data from an external endpoint. This data is stored on the state via a provider. Then we calculate the number and emissions from the isos that are provided.

Currently, the endpoint returns an empty array so the number is 0 but when this change it will show the correct number and data.

image

### DIFF
--- a/app/javascript/app/providers/lse-provider/lse-provider-actions.js
+++ b/app/javascript/app/providers/lse-provider/lse-provider-actions.js
@@ -1,0 +1,40 @@
+import { createAction } from 'redux-actions';
+import { createThunkAction } from 'utils/redux';
+import isEmpty from 'lodash/isEmpty';
+import { apiWithCache } from 'services/api';
+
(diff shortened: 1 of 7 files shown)


app/javascript/app/providers/lse-provider/lse-provider-actions.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const fetchLSE = createThunkAction(
+  'fetchLSE',
+  () => (dispatch, state) => {
+    const { lse } = state();
+    if (lse && isEmpty(lse.data) && !lse.loading) {
+      dispatch(fetchLSEInit());
+      apiWithCache
+        .get(
+          '//laws-pathways-staging.vizzuality.com/cclow/api/targets/economy-wide-countries'
```
simaob, line 18: you can already point it to production: climate-laws.org

PR #1179, opened 2020-07-21
Fix/top emitters

Bluesmile82:
This little PR updates the TOP Emitters region and fixes the unit for the y-axis on GHG emissions

image


Bluesmile82:
Thank you! I changed the top emitters. It was another constant. Now we have only one.

The problem with European Emissions is about the lenght of the span we can add it as a little task but we would have to check design
image

PR #1182, opened 2020-07-22
Fix/full content page params

kfilip94:
This PR fixes two issues related to `NDC Search` and `NDC Full Content` pages.
[PIVOTAL](https://www.pivotaltracker.com/story/show/173054830)

fc4ak-e75kj

- Now the search text is correctly passed, from NDC Search to the NDC Full content page. There is a problem with the `NavLink` url in the `ResultCard` component. Between `...searchBy=${search.searchBy}` and `&query=${search.query}...` there is an "enter" character there, what in template literals breaks the line of the string, instead of formatting the code in the code editor, short explanation below:
image

- The selection of the document on the `NDC Full Content` works correctly now; there is a bug when a country has more than one language version of a document, so it always chooses the default language instead of the selected one. A small change of the logical expression solved the problem.

**Test:** On the `NDC Search`  type in the search text and click at one of the results, and on the `NDC Full Content` page test documents with different language versions, check if they display correctly.

### DIFF
--- a/app/javascript/app/components/result-card/result-card-component.jsx
+++ b/app/javascript/app/components/result-card/result-card-component.jsx
@@ -22,8 +22,7 @@ const ResultCard = props => {
         result.matches.map(match => (
           <NavLink
             key={`${match.fragment}-${match.idx}`}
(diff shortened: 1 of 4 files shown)


simaob:
Don't forget to review this! =)

PR #1183, opened 2020-07-28
Enabling importing targets for multiple NDC documents

simaob:
This PR is the first step to fixing a couple of issues with the NDC SDG Linkages pages, that are meant to be done under the ongoing maintenance contract, the issues to fix are:

- matching SDG goals and targets against multiple versions of the NDC documents;
- importing of translated versions of the documents (see: https://basecamp.com/1756858/projects/13795275/todos/420299366#comment_769964171)

To test this you'll need to run the ndc_sdg_targets importer, but to be safe you might want to reimport these:

```
bundle exec rails ndcs:full:import ndcs:full:index
bundle exec rails sdgs:import
bundle exec rails ndc_sdg_targets:import
```

The last one is the one that has been changed in this PR.

After the importing is ran you should see that more files on the Full Text pages have targets on the Goals/Targets dropdown. One good example to test it with is Andorra: http://localhost:3000/ndcs/country/AND/full?document=first_ndc-EN

There's something that I'm not 100% sure that it's working which is rendering the Translated versions, see the case of Andorra where there are a lot of options in there, not sure if the ones that are translated work well (they'll have true on the translated attribute of the API response.

If you need help with the importers, Agnieszka should be able to help as she has worked on this (ages ago) or reach out to me on WhatsApp! =)

Take care!

PS: I'm PRing from a fork, because I removed myself from the Vizz org... 😅

### DIFF
--- a/app/controllers/api/v1/ndc_texts_controller.rb
+++ b/app/controllers/api/v1/ndc_texts_controller.rb
@@ -124,8 +124,8 @@ module Api
           ndcs = ndcs.where(id: ::NdcSdg::Sector.ndc_ids(params[:sector])) if params[:sector]
         end
 
-        linkages = Ndc.linkages(params)
         ndcs.map do |ndc|
+          linkages = Ndc.linkages(params, ndc)
           ndc_linkages = linkages.
             select { |linkage| linkage.ndc_id == ndc.id }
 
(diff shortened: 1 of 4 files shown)


simaob:
This PR https://github.com/Vizzuality/climate-watch/pull/1182 might be related with the selection issue I mention above.

PR #1199, opened 2020-08-06
More fixes to ndcs controller endpoint

tsubik:
This PR is fixing specs locally and custom compare page which was not showing all indicators for categories. Indicators have `category_ids` attribute and for some indicators not all categories were present. This happened because of using distinct on clause the `category_ids` relation for indicator was not returning all indicator categories but only those from the initial query.

image


tsubik:
I initially tried to filter indicators to present only those for which `category` is also returned in `categories` collection. Meaning, for all valid categories based on filter parameters. But, looks like that broke application in a different point, LTS Explore, because that map relies on `lts_ghg` indicator which is not in `map` category but was also returned by the endpoint, even if it would be more correctly to return only `map` indicators. This could be fixed next week when Alvaro is back as I'm not that familiar with the frontend part of this part of application.

app/controllers/api/v1/ndcs_controller.rb
```diff
@@ -205,8 +211,11 @@ def filtered_indicators(source=nil)
         indicators = indicators.where(source_id: source.map(&:id)) if source
 
         if @indc_locations_documents
+          # TODO: do not understand why below are needed
+          # if indicator belongs to many cateogires then only one will be in category_ids
+          # that's why I'm going to reset the query at the end
           indicators = indicators.select('DISTINCT ON(COALESCE("normalized_slug", indc_indicators.slug)) indc_indicators.*')
```
tsubik, line 217: @Bluesmile82 do you know what is normalized_slug and why it is used when passing locations_documents parameter?

PR #1201, opened 2020-08-10
Fix 2020 tracker

tsubik:
Label slug was not returned for indicator values.


tsubik:
I totally do not understand why sometimes label slug is not returned, and when it should be returned as clearly on that page was needed.

Bluesmile82:
Before, the label was returned only when we had a correspondent indicator with a value. I thought we hadn't changed that but maybe we did

tsubik:
Take a look now @Bluesmile82 , both LTS Explorer and 2020 NDC tracker should work fine now.

Bluesmile82:
Thanks! The modules are working fine now again!

PR #1214, opened 2020-08-13
Dont filter ndc comparison by overview and merge categories

Bluesmile82:
Don't merge: We want to be sure if we should apply the filter=overview or not to the NDC comparison and Custom compare pages. Waiting for WRI answer

If we don't want this filter we should remove it from Custom compare call and apply the same merge of categories to this data.

On NDC comparison page we had duplicated categories as we don't have any category type filter like in custom compare (filter=overview). On the frontend, we have merged the indicators from these categories to display them together.

The sectoral info, which data is treated separately, doesn't have its indicators merged as we don't see this need. Take it into account if there is any problem.


Bluesmile82:
I think this is outdated, closing...

PR #1223, opened 2020-08-19
Fixing INDC categories relations

tsubik:
Based on the client feedback, the `map` category has no relation with `global` category. Global category is a parent category for overview categories. That's why I'm removing relations between `map` and `global` categories. 

Another thing to change was the uniqueness of categories slug. From now on it will be unique in the scope of category_type (map, overview, global) and parent_id to allow the same-named categories under different parents. Example: `Planning Process` is under both `Overview` and `Adaptation`, that's why we will create 2 different categories.

This hopefully should fix the issues with duplicated categories and missing indicators in comparison pages. 

Importer was changed, so please run `bin/rails indc:import`

### DIFF
--- a/app/models/indc/category.rb
+++ b/app/models/indc/category.rb
@@ -14,6 +14,6 @@ module Indc
 
     validates :slug, presence: true
     validates :name, presence: true
-    validates :slug, uniqueness: {scope: :category_type}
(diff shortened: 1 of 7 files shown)


Bluesmile82 (approved):
Thanks! I think this is fixed

PR #1271, opened 2020-09-18
add ndcs controller index cache

tsubik:
Adding cache for ndcs index controller, but only when almost no filter applies.


agnessa:
to my future self: this will attempt to use the cached version when filter + source only are specified, which is the slow 5MB request on explore page

agnessa (approved):

PR #1308, opened 2020-10-20
fix query to get indicators when location, document pair is found

tsubik:
Fixing issue with fetching too many indicators, even if some do not have any value for any location, document pair.
https://basecamp.com/1756858/projects/13795275/todos/408564766#comment_784446273

The query was wrong to check if the indicator has a value for any selected document and any selected location, not a value for the pair of document-location.


Bluesmile82 (approved):
Looks fine! Let's deploy it for them to check

PR #1318, opened 2020-11-10
Net zero - importer, some frontend changes

tsubik:
- improving INDC importer, works 3x faster now
- changing Net Zero Tracker source name to ECIU
- couple frontend changes to test the data, not changing everything or renaming `lts` prefixes to `netZero` or smth.
- nz_ghg indicator has some strange data, not like lts_ghg, or ndcs_ghg which are percentage

New data was uploaded to staging folder on AWS. To test it just reimport indc data

```
bin/rails indc:import
```


Bluesmile82 (approved):
Thanks! It's working fine! I've added some changes to create a new NetZero reducer and remove all LTS mentions. I will continue the frontend work from this

PR #1321, opened 2020-11-12
Fix/ghg gdp data

Bluesmile82:
This PR fixes two problems:

- WB country data import: The rails model had a validation that prevented missing GDP or population data to be imported. This was hiding existing population data on the frontend. The validation has been removed.

- Calculation double counting: Due to changes in the way we group the regions the calculation data (population, gdp) was double or multiple counted. Fixed now:

Before:

image

After:

image

### DIFF
--- a/app/javascript/app/components/ghg-emissions/ghg-emissions-selectors/ghg-emissions-selectors-data-utils.js
+++ b/app/javascript/app/components/ghg-emissions/ghg-emissions-selectors/ghg-emissions-selectors-data-utils.js
@@ -9,35 +9,14 @@ const expandRegionToCountries = (iso, regions) => {
   return flatMap(region.members, expandRegionToCountries);
 };
 
-const expandRegionsToCountries = (isos, regions) =>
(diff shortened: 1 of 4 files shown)


agnessa (approved):

app/javascript/app/components/ghg-emissions/ghg-emissions-selectors/ghg-emissions-selectors-data-utils.js
```diff
@@ -9,35 +9,14 @@ const expandRegionToCountries = (iso, regions) => {
...
-) => {
+export const getMetricData = (year, iso, metricField, calculationData) => {
   const getMetricForYearAndRegion = (y, r) =>
     metricField &&
     calculationData &&
     calculationData[y] &&
     calculationData[y][r] &&
     calculationData[y][r][metricField];
-  let metricData = getMetricForYearAndRegion(year, column.iso);
-  // if no metric data for expandable column then use expanded regions to
```
tsubik, line 29: Is that not needed anymore? I think that was for regions for which there was no aggregated data.

PR #1335, opened 2020-11-19
add net zero data explorer endpoints

tsubik:
The difference is that I removed sectors from Net Zero tab, as there is no sectoral data.


tsubik:
Hmm, I thought I removed it, but still showing up at the end of the table.

Bluesmile82 (approved):
Thanks! I'm merging this to continue working

PR #1345, opened 2020-12-03
lists need bullets also on second hierarchy

Bluesmile82:
Updated the styles so the nested lists also have bullets

image


Bluesmile82:
@sw1115 Please confirm if they should show like this :)

sw1115:
> @sw1115 Please confirm if they should show like this :)

@Bluesmile82 Thanks! yes it looks great! :)

PR #1346, opened 2020-12-03
Key visualizations frontend

jookyboi:
There are additional components to add for this to be feature complete but I'm submitting this to ensure I get most of the heavy lifting out of the way.

Key Visualizations | Climate Watch Data 2020-12-03 09-45-30


jookyboi:
Closing for cleaner PR.

PR #1350, opened 2020-12-07
Key visualizations frontend

jookyboi:
Key Visualizations | Climate Watch Data 2020-12-07 08-53-08


Bluesmile82 (approved):
Thank you very much for cleaning the history. I replaced some style values with variables. Actually, there were some other missing in the application

PR #1361, opened 2020-12-16
net-zero intro text

sw1115:
hi @Bluesmile82 - we've just made some final changes to the net-zero intro text. Thanks!


sw1115:
also added explore NDC definition in the faq document. Thanks!

Bluesmile82 (approved):

PR #1365, opened 2020-12-17
Key visualizations QA

jookyboi:
- Moved button group to header for visualization preview.
- Added Info button for each visualization.
- Added border to separate preview from rest of page.
- Updated behavior of multi-select tags/geographies.
- Moved "Key Visualizations" from NDCS to About menu.

Key Visualizations | Climate Watch Data 2020-12-17 13-01-01


Bluesmile82 (approved):
Thanks!

PR #1400, opened 2021-02-25
Missing abort option + transaction for NDC texts importer

agnessa:
The NDC importer is not using the data uploader gem to render the view, because it is slightly different (no user upload) and so needs a custom view. That  is why the abort link was missing, it was added in the gem and not in the custom view.

Additionally, I wrapped the importer in a transaction.


tsubik (approved):
LGTM :+1:

PR #1406, opened 2021-03-10
GitHub actions

tsubik:
- Adding GH actions
- force_ssl set to true for production
- Adding Brakeman and bundle audit for periodic security checks
- Travis config removed


tsubik:
Hmm, I wonder if that old ruby version is not messed up, create database commands randomly fails. I would upgrade Ruby for this project to `2.5.7`. Would love to do it to at least `2.7.2` but code coverage is pretty low! And I'm not sure if that won't break anything

agnessa (commented):
I'm not sure this is all that's needed when upgrading the ruby version - do we need to specify a different image for deployment? I am not against upgrading ruby, since 2.5 is approaching EOL, but would be good to double check is anything else needs doing.

Also, what is the random error you get with the older ruby version? Not sure if related to your case, but I occasionally get `/home/runner/work/trase/trase/vendor/bundle/gems/ffi-1.13.1/lib/ffi/library.rb:112: [BUG] Illegal instruction at 0x00007f739c7ecf00 ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-linux]` on ruby 2.6.3 and as far as I could figure out this is not due to ruby version, but to caching gems. I haven't found a solution yet, I just hit re-run every now and then.

tsubik:
Yes, exactly the same error https://github.com/ClimateWatch-Vizzuality/climate-watch/runs/2080197354?check_suite_focus=true. 
I never ever encountered that on Laws and Pathways which is on Ruby 2.7.2, so I thought it might be related to the old version. Dockerfile was changed, what else should be changed to upgrade Ruby, I think that will be it. I would like to at least upgrade with the latest patches, as for 2.7.2 it's totally failing and would need to ask for a budget to do it and test it.

tsubik:
Kinda also scared that code coverage is like 50% :D

agnessa:
Interesting, in that case I'll propose to upgrade in Trase as well.

tsubik:
Could be also related to bundler version? maybe

tsubik:
Going to merge this and we will put that on staging.

PR #1407, opened 2021-03-11
Notifications

tsubik:
image

API at `/api/v1/notifications` returned only 9 latest notifications ordered by date. Note: It does not return notification with the date set in the future.

image


tsubik:
I added model annotations, so sorry that it created such a big PR.

Bluesmile82 (commented):
Thanks!

PR #1408, opened 2021-03-11
proxy newsletter signup to bypass cors problems

tsubik:
Still no response for newsletter problem, but I would try with this simple proxy through our backend. I also corrected error handling for the both requests, ehh that `fetch` API :man_facepalming:


tsubik:
Even if I have 200 response from that form handler I still don't know if that works. Not getting any email message like "Thanks for subscribing".

tsubik:
Finally, welcome message was set up and this approach with proxing by our backend works fine. Error handling was improved, depends on the error origin it will show `Error while processing your feedback` or if there was an error while submitting to the newsletter `Error while submitting to the newsletter`. In both cases the data will still be downloaded, but modal won't be closed. (as I didn't want to block the download). I think this could be moved to staging for testing any time or directly to prod @Bluesmile82

Bluesmile82 (approved):
Looks nice. Lets see it work on staging

PR #1414, opened 2021-03-24
2020 NDC text changes

mengping:
Include:
- Text above the map
- Text in left side panel
- Year of emissions data
We may expect additional changes to the wording of enhancement category still


Bluesmile82 (approved):
Thanks! I'll add a title tag on the link for accessibility

PR #1417, opened 2021-03-26
Upgrade mimemagic

Bluesmile82:
We can't use mimemagic 0.3.5 anymore as it has been removed and there are some license problems with 0.3.6 so we changed the version to 0.3.8

You will probably need to install shared-mime-info in your system. This is updated on the readme:
https://github.com/rails/rails/issues/41757

(mac)
```
brew install shared-mime-info
```

(linux)
```
sudo apt-get install shared-mime-info
```

https://github.com/rails/rails/issues/41750
https://github.com/rails/rails/issues/41757

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ ENV USER_REPORT_KEY $USER_REPORT_KEY
 RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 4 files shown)


agnessa (approved):
OK, let's try to deploy that, Tomasz found a different solution we could try if this doesn't work: https://github.com/rails/rails/issues/41757#issuecomment-808024283

tsubik:
I think dockerfile must be updated. To install that shared-mime-info

nezirz:
I had issue with my docker app and solution from @Bluesmile82 helped me. By adding sudo apt-get install shared-mime-info into docker file. Thanks man!

PR #1436, opened 2021-04-21
Feature/update webpack

Bluesmile82:
-- Update webpack to 5.32 --

This update is due to performance issues mostly
It will be followed at some point with some performance optimization (split code with dynamic imports, improve cache ...)

WIP: This change has to be thouroghly tested
There is a deprecation warning although I can see any immediate problems on the code apart from:

- I have found a style problem after the change in pathways page.
- Also MY-CW authentication doesn't seem to work (the callback url doesn't go back to the page)

I also tried a fast implementation of vite in the feature/vite branch but I couldn't go much further as we are using directory-named-webpack-plugin and I couldn't find something similar with vite (maybe we have to make it custom)


tsubik:
Hmm, shouldn't that be done with upgrading Rails webpacker to 5.x version first?

app/javascript/app/components/country/country-ghg-emissions/country-ghg-emissions.js
```diff
@@ -115,7 +115,7 @@ class CountryGhgEmissionsContainer extends PureComponent {
       this.props.setModalMetadata({
         category: 'Country',
         slugs: isPageContained ? [source] : [source, 'ndc_quantification_UNDP'],
-        customTitle: 'Greenhouse Gas Emissions and Emissions Targets',
+        customTitle: 'Gruse Gas Emissions and Emissions Targets',
```
tsubik, line 118: some accident ;]

tsubik:
I think bin/webpack must be present otherwise rails commands are erroring with "webpack binstubs not found."

Bluesmile82:
> Hmm, shouldn't that be done with upgrading Rails webpacker to 5.x version first?

Very true. That made the first run much slower (It's still awfully slow). Not sure about the binstubs change

tsubik:
I always don't know the details of webpack config and find it hard to customize (as not doing that often) and docs are pretty big. I would first do upgrade of webpacker and then `rails webpacker:install` using defaults (I never changed much there tbh), and overwrite what is needed after (as I see there is already some customization done in this project).

tsubik:
I see that original setup was not using webpacker default configuration at all, instead the whole config is done in the repo itself.

Bluesmile82:
> I see that original setup was not using webpacker default configuration at all, instead the whole config is done in the repo itself.

I will run the default command and check the differences. I know more about the custom webpack config that we have now

tsubik:
Couple more hours and going through the list https://webpack.js.org/guides/build-performance/ could make it better, but no guarantees if a way better.

tsubik:
I see couple more things are not working. Reloading page on update script or stylesheet, that could be fixed by adding `target: "web"` into webpack config, but still HMR does not work, and this is something that could vastly improve development. Now, any style changes took the whole page reload and this almost 10 sec at least.

Bluesmile82:
Thanks for checking. We want to spend more time on this before merging then. Yes, I also experienced that was quite slow

PR #1460, opened 2021-06-08
Feature/geolocation backend

tsubik:
`MAXMIND_LICENSE_KEY` is needed to use real DB. That should be set in Jenkins for building docker image. On every docker image build, we will be downloading fresh IP database.

User country ISO code2 is available in content of meta tag named `cw-user-country`.


app/services/geolocation_service.rb
```diff
@@ -0,0 +1,24 @@
+class GeolocationService
+  include Singleton
+
+  class << self
+    delegate :call, to: :instance
```
agnessa, line 5: love this, why did I never think of that ❤️ will steal this approach to singletons going forward

agnessa (approved):

PR #1463, opened 2021-06-17
Feature/abbr

Bluesmile82:
This PR creates a way to add abbr tags into the text on the whole site:

image

This new feature needs the FEATURE_ABBREVIATIONS tag to be true
 
To achieve this the plan is to parse the text and components which are suitable to contain abbreviations (to scope the possible errors and limit performance leaks) to do this a new AbbrReplace component and a replaceStringAbbr function have been created. 

The function is just for dangerouslySetInnerHtml so we only replace the text.

The component parses its children and replaces the abbreviations adding <abbr> native tags, also:

- Adds subscripts when needed E.g. CO2
- Is fault tolerant. If we have any error defaults to the unaltered children
- Manages conflicts between the abbreviations with a constant. The Longer Abbreviation should be first on the abbreviations list and added to the conflict constant.
- In some cases adding the extra tag could mess with the layout so the solution is to add a containing div. This is done by passing a 'fixLayout' prop

MISSING:
Some components that don't still have the abbreviations:

- Info modal content.  We will probably need this. It's using react-markdown and none of this strategies work for now. Most probably we will have to open another PR for this.
- Map checkbox - We will need to tweak CW components
- Map legend- We will need to tweak CW components
- Table Headers - This will need some work. It probably can be done. But for some tables we will have to change CW components
- Table cells - This is working in some tables that use the CW components table, not on the ones that use the one on the app. It could be expensive performance wise though. We might want to consider this.
- Dropdowns and Multidropdowns and similar components - Not planned
- Search placeholder - Maybe not needed


tsubik (approved):
Looks pretty complex but great! I'm not sure how I would proceed with such a task, ideally abbrevations tags should be set just directly on the abbrevations, but with existing layout that seems hard, especially when flex is removing those whitespaces.

PR #1492, opened 2021-09-16
Merge shared db strucutre into main db

tiagojsag:


tiagojsag:
they are the old  numbers manually butchered to be at the end of the migration list in the right order

agnessa (approved):

PR #1525, opened 2021-10-26
Feature/us states

tsubik:
For testing change env variable `CW_FILES_PREFIX=climatewatch.org/development/climate-watch/` and reimport locations, members, and historical emissions:
`bundle exec rails locations:import`
`bundle exec rails location_members:import`
`bundle exec rails historical_emissions:import`


Bluesmile82:
I made some frontend changes to fit the new source. Some gases are missing on the data but are supposed to be there as the meta says. Also, the socioeconomic data is also missing but we already know that. I will write on Basecamp

PR #1572, opened 2022-04-13
Chore/upgrade backend deps

tsubik:
Upgrade ruby, bundler and vulnerable dependencies.


agnessa (approved):
I'm guessing this is so that security checks only run on push now, which is fine (I had no idea they could run on schedule to be honest 😊 )

tsubik:
I was annoyed by failing security job emails :D, maybe it should just write something on slack and not email me :P

PR #1579, opened 2022-05-17
Country profiles feedback - 1st round

andresgnlez:
Feedback based on [this document](https://docs.google.com/document/d/1930oZq1Hh0bozutdjvzZlenJ_jifNVGS/edit).

### DIFF
--- a/app/javascript/app/components/country/country-subnational-actions/country-subnational-actions-component.jsx
+++ b/app/javascript/app/components/country/country-subnational-actions/country-subnational-actions-component.jsx
@@ -5,6 +5,7 @@ import uniq from 'lodash/uniq';
 import orderBy from 'lodash/orderBy';
 import isEmpty from 'lodash/isEmpty';
 import mapValues from 'lodash/mapValues';
+import { format } from 'd3-format';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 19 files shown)


app/javascript/app/components/country/country-subnational-actions/country-subnational-actions-component.jsx
```diff
@@ -110,11 +116,25 @@ function SubnationalActions({ indicators, loading, handleInfoClick }) {
     <div className={styles.noData}>No data available.</div>
   );
 
+  const countryName = useMemo(() => {
```
Bluesmile82, line 122: We normally do this on selectors but is ok 👌

Bluesmile82 (approved):
Great job! Thanks

andresgnlez:
@Bluesmile82 I've tried, a couple of times, doing those selectors and, _for some _reason__, the props I was relying on sometimes are undefined leading to crashes, so I opted out for hooks, whenever possible. But yep, my initial intention was keeping the current structure with selectors.

Bluesmile82:
No worries, for the future let's try with selectors. Maybe we can check that together to see what is happening

PR #1580, opened 2022-05-23
Release/2.16.0

tsubik:
- Text changes.
- Ruby version upgraded to v2.6.10.
- Fix issues with the latest document selection when using "All documents" option in NDC explorer map.
- NDC adaptation commitments - allow having multiple values for the same indicator and group them by grouping indicator. Add separator for indicator groups to compare NDC accordions.
- Remove links to old ndc/compare page .


Bluesmile82 (approved):
Thanks!

PR #1584, opened 2022-05-25
Emplyment and costs charts

andresgnlez:
- Upgrades employment to chart to receive data dynamically. 
- Adds costs chart.
- Updates data source text.
- GHG emissions: fixes ranking color and position.
image
image
image


Bluesmile82:
Could you check the height of the grid on the renewable employment chart? I know is not part of the PR but it gets short when we have many values 
image

Bluesmile82 (approved):
Check the comment, the rest looks great. Thanks!

PR #1587, opened 2022-05-31
ndcs: merges lts and net-zero indicators

andresgnlez:
image

Merges LTS and Net Zero to NDCs.

The feature works with the feature flag `FEATURE_NDC_UPDATE` set to `true`. Otherwise, previous data should appear as usual

### DIFF
--- a/app/javascript/app/components/country/country-ndc-overview/country-ndc-overview-component.jsx
+++ b/app/javascript/app/components/country/country-ndc-overview/country-ndc-overview-component.jsx
@@ -35,7 +35,8 @@ function CountryNdcOverview(props) {
     isNdcp,
     handleInfoClick,
     handleAnalyticsClick,
-    selectedDocument,
+    ndcsDocument,
+    ltsDocument,
     countryName
   } = props;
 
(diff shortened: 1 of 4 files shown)


Bluesmile82:
We need to change the buttons to just one with the following text, like on the design
image

app/javascript/app/components/country/country-ndc-overview/country-ndc-overview-component.jsx
```diff
@@ -94,72 +96,224 @@ function CountryNdcOverview(props) {
...
+                <CardRow
+                  rowData={{
+                    title: 'Target type',
+                    value: values.ghg_target
+                  }}
+                  theme={cardTheme}
+                />
+              )}
+
+              {values?.mitigation_contribution_type && (
```
Bluesmile82, line 121: This one should be first as in the design

Bluesmile82 (changes requested):
Check the comments. I will probably take care of the changes anyway

Bluesmile82 (changes requested):
The links looks and works good now, thanks! But for some reason I'm not getting data on the indicators of the section for Canada. If you get it maybe I''m just missing to import it so we can try on staging. 

Also maybe we can add here or as a new task changing the names of the anchors in the menu. This one should be "Content of Commitments" if the feature tag is active

Bluesmile82:
If I remove the feature tag the title of the NDC section now is: 
image

Can you put back the "Nationally Determined Contributions (NDC) Overview" if we don't have the feature tag active?

Bluesmile82 (changes requested):
Hi! The title is still changed when the tag is false. It is now

image

And it should be 
image

Bluesmile82 (approved):
Great thanks!

PR #1591, opened 2022-06-22
country profile: adds gradient to navbar

andresgnlez:
image

Adds gradient to the navbar. ~~There's an issue with the right gradient: as the background header is based on another gradient, it is not possible to determine which colour the navbar gradient should be based on as there will be always some slight contrast due to the header gradient. **A discussion with design is pending to solve this.**~~


Bluesmile82 (approved):
Looks great! Thanks!

PR #1595, opened 2022-06-24
makes ghg-emissions chart responsive

andresgnlez:
image
image
image

Makes GHG-emissions chart responsive on resize


Bluesmile82 (approved):
Nice, it works great! Thank you!

PR #1596, opened 2022-06-24
Feature/country ghg previous targets

Bluesmile82:
image
This PR adds the Show previous targets switch on the GHG emissions section of the country page. 
When active the chart shows all "quantifications" / targets. When deactivated only shows the last and the lts ones.


andresgnlez:
@Bluesmile82 good job!

andresgnlez (approved):

PR #1662, opened 2023-06-01
fix: Make docker compose work

martintomas:
Minor tweaks of Dockerfile and docker-compose so it works again.


SARodrigues:
@martintomas @Bluesmile82
Just tried this fix and it works wonderfully, provided the `.env` is correctly setup.

Bluesmile82 (approved):

PR #1663, opened 2023-06-08
refactoring: Make iso_code3 code visible at HistoricalEmissions downloaded file

martintomas:
This PR adds `ISO` to HistoricalEmissions downloaded file.

https://basecamp.com/1756858/projects/13795275/todos/479334083

### DIFF
--- a/app/services/api/v1/data/historical_emissions/csv_content.rb
+++ b/app/services/api/v1/data/historical_emissions/csv_content.rb
@@ -7,15 +7,16 @@ module Api
         class CsvContent
           def initialize(filter)
             @grouped_query = filter.call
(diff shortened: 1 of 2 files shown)


martintomas:
Just wondering @agnessa , I see that `csv` is gonna humanize name of this new column and it is gonna be `Iso code3` which does not look like user friendly. I was thinking to set alias to `alias: ISO`, but then I would need to remove `humanize` part which I am not sure how much implications it is gonna have, because from my point of view, this `humanize` stuff is necessary there because of other fields

agnessa (approved):
wow, that takes me back, I forgot we had a column configuration :)

PR #1667, opened 2023-06-30
Fix NDC Enhancement page crashing/not loading

SARodrigues:
## Description

This PR fixes the _"NDC Enhancement Tracker"_ page crashing when the user visited _Explore NDCS_ first. 

It also includes some minor fixes to address some of the console errors found during the debugging process, as well as a couple tweaks in an attempt to make it so that similar issues don't crash the page in its entirely.

### DIFF
--- a/app/javascript/app/components/ndcs/ndcs-enhancements-map/ndcs-enhancements-map-selectors.js
+++ b/app/javascript/app/components/ndcs/ndcs-enhancements-map/ndcs-enhancements-map-selectors.js
@@ -296,7 +296,7 @@ export const getPreviousComparisonCountryValues = createSelector(
   [getPreviousComparisonIndicators, getISOCountries],
   (previousComparisonIndicators, isos) => {
     if (!previousComparisonIndicators) return null;
-    const previousComparisonCountryValues = [];
+    const previousComparisonCountryValues = {};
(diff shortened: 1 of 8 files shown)


app/javascript/app/components/ndcs/ndcs-enhancements-map/ndcs-enhancements-map-selectors.js
```diff
@@ -296,7 +296,7 @@ export const getPreviousComparisonCountryValues = createSelector(
   [getPreviousComparisonIndicators, getISOCountries],
   (previousComparisonIndicators, isos) => {
     if (!previousComparisonIndicators) return null;
-    const previousComparisonCountryValues = {};
+    const previousComparisonCountryValues = [];
```
Bluesmile82, line 299: This looks like an object. We are adding attributes to it

Bluesmile82 (approved):
Looks nice! Check my comment

app/javascript/app/components/ndcs/ndcs-enhancements-map/ndcs-enhancements-map-selectors.js
SARodrigues: Thank you for catching that.

PR #1670, opened 2023-07-19
Fix visualizing individual EU members on NDCS/LTS/Net-Zero explorers

SARodrigues:
## Description

This PR fixes a couple bugs affecting the following pages: 
- **Explore NDCS** (`/ndcs-explore`)  
- **Explore LTS** (`/lts-explore`)  
- **Net-Zero Tracker** (`/net-zero-tracker`)  

&nbsp;

### The issues
- Map
  - EU as a region and individual EU countries not correctly highlighted
    _Eg: with location set to "World" and the "Visualize individual submissions of EU Members on the map" checkbox unchecked, the EU is not highlighted. Similarly, with the location set to "European Union (party to the UNFCCC)" and the checkbox checked, the individual EU countries are not highlighted_
- Table
  - Similarly to what happens with the map, the table is not displaying the entries correctly and is responding to the checkbox's state

&nbsp;

### Investigation  

There are two ways in which EU countries can be selected:
- By the `EUU` iso code  
- By the EU countries' respective iso codes (`PRT`, `ESP`, etc)  

`pathsWithStylesFunction` (`components/ndcs/shared/selectors.js`), responsible the paths displayed on the map as well highlighting countries/regions, does its thing based on an array of iso codes provided by `selectedCountriesISO` as well as the _"Visualize individual submissions"_ checkbox. 

There are a couple issues with this:  
- If the EU countries are selected by the `EUU` iso code, and we want to display individual submissions, we lack the list of individual EU countries' iso codes  
- If the EU countries are selected by their individual iso codes, but we want to display the EU as a whole, we lack the `EUU` iso code  
- Basing the display solely on whether or not the checkbox is checked, can break the map when it doesn't make sense to display the EU as a whole eg: the user selected a single EU country. 

Similarly, the documents displayed on the table are based on `selectedCountriesISO`. Note that this function is also used to process the data and display statistics, especially on the sidebar. There is currently no mechanism to allow us to display documents based on the checkbox's state. 

&nbsp;

### Proposed solution  

The first obvious place to manipulate the ISO in order to fix the map display would be `selectedCountriesISOFunction` (or perhaps `selectedCountriesFunction` where the list of selected countries is coming from in the first place)(`components/ndcs/shared/selectors.js`). 

However, because these functions are complex and used to parse/manipulate data to display in the pages (eg: sidebar statistics), and we really only to manipulate the ISOs for the map display, it seems less bug prone to create a map specific selector which takes the resulting ISO array and manipulates it for the map display. 

**Map**  

Addition of a `selectedMapCountriesISOFunction`, to be used by the `pathsWithStylesFunction` instead of the original function. 
It takes the list of ISOs from `selectedCountriesISO` as well as the checkbox's state, and manipulates it to address the issues described in the _Investigation_ section: 
  - If `EUU` is selected and we need to display individual countries, we:
    - remove the `EUU` iso code and add the individual countries' ISO codes  
  - If the individual ISO codes are selected but we need to display the EU as a whole:
    - remove the individual ISO codes and add the `EUU` one
  - We restrict this manipulation to only where it makes sense: when all EU countries are displayed, be it by the `EUU` iso code or their individual ones. This fixes the issue where a single EU country is selected and the checkbox breaks the display. 

**Table**  

Because the original data parsing is untouched, and we have a correct list of ISO codes to fix the map display, we can simply sync the table with the map display (which will include the checkbox's state) by simply using the new `selectedMapCountriesISOFunction` instead of `selectedCountriesISOFunction` one. 

&nbsp;

## Testing  

For each of the following:
  - `/ndcs-explore`  
  - `/lts-explore`  
  - `/net-zero-tracker`  

Test the following: 
- Set _Location_ to _World_: 
  - When the _"Visualize individual (...)"_ checkbox is unchecked, verify that:
    - the map displays the EU as a whole, correctly highlighted
    - the table contains an _"European Union"_ entry  
  - When the checkbox is checked, verify that:
    - the map displays individual EU countries  
    - the table contains the individual countries' entries
- Set _Location_ to _European Union (region)_
  - _Same as above_  
- Set _Location_ to _Top Emitters_
  - _Same as above_  
- Set _Location_ to _European Union (party to the UNFCCC)_
  - _Same as above_  
- Set _Location_ to one or more EU countries (eg: Portugal, Spain)  
  - When the checkbox is checked, verify that: 
    - The map displays the countries correctly
    - The table includes the countries' entries  
  - When the checkbox is unchecked, verify that:
    - The map and table aren't affected
      _(displaying the EU as whole doesn't make sense in this case)_  

&nbsp;

## Tracking 

[Basecamp](https://basecamp.com/1756858/projects/13795275/todos/479155492)

### DIFF
--- a/app/javascript/app/components/ndcs/lts-explore-map/lts-explore-map-selectors.js
+++ b/app/javascript/app/components/ndcs/lts-explore-map/lts-explore-map-selectors.js
@@ -20,6 +20,7 @@ import { TOP_EMITTERS_OPTION } from 'data/constants';
 import {
   selectedLocationsFunction,
   selectedCountriesISOFunction,
+  selectedMapCountriesISOFunction,
   selectedCountriesFunction,
   categoryIndicatorsFunction,
   pathsWithStylesFunction,
@@ -135,6 +136,11 @@ const getisDefaultLocationSelected = createSelector(
 export const getSelectedCountriesISO = createSelector(
   [getSelectedCountries],
   selectedCountriesISOFunction
+);
+
+export const getSelectedMapCountriesISO = createSelector(
+  [getIsShowEUCountriesChecked, getSelectedCountriesISO],
+  selectedMapCountriesISOFunction
 );
 
 export const getCategories = createSelector(getCategoriesData, categories =>
--- a/app/javascript/app/components/ndcs/lts-explore-table/lts-explore-table-selectors.js
+++ b/app/javascript/app/components/ndcs/lts-explore-table/lts-explore-table-selectors.js
@@ -6,7 +6,7 @@ import isEmpty from 'lodash/isEmpty';
 import { replaceStringAbbr } from 'components/abbr-replace';
 import {
   getMapIndicator,
-  getSelectedCountriesISO
+  getSelectedMapCountriesISO
 } from 'components/ndcs/lts-explore-map/lts-explore-map-selectors';
 
 const getCountries = state => state.countries || null;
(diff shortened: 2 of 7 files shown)


Bluesmile82 (approved):
Thanks!

PR #1702, opened 2024-10-22
Fix/2025 ndc tracker feedback v2

SARodrigues:
- [ ] 1.0 - What happens if we have +8 submissions on a same day?   
- [x] 1.1 - Change main subtitle
- [x] 1.2 - Adjust hyperlinks  
- [x] 1.3 - Adjust alignment of white boxes and their titles  
- [ ] 1.4 - Adjust area that is actionable in the toggle feature  
  _In progress (separate branch)_
- [x] 1.5 - Copyedit   
  _This refers to the trailing space in the TrackerChartComponent's description_  
- [x] 1.5 - Copyedit
  _This refers to the Abbr Replacements that are native to the platform_  
  - [x] Page Title  
  - [x] Timeline Description

### DIFF
--- a/app/javascript/app/components/intro/intro-component.jsx
+++ b/app/javascript/app/components/intro/intro-component.jsx
@@ -18,7 +18,8 @@ const Intro = props => {
     button,
     customButton,
     className,
-    tag
+    tag,
+    skipAbbrReplace
   } = props;
   const actionButton =
     (customButton || button) &&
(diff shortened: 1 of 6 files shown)


Bluesmile82:
For 1.0 we already have the scroll on the tooltips

SARodrigues:
> For 1.0 we already have the scroll on the tooltips

It's not the tooltips that appear when we hover over a day "circle". It's the list of countries displayed below said circle.  
If we display enough countries they'll overflow the blue "intro" area and they'll be hidden.  

As for scrolling I don't think it'll be the correct solution, because when duplicating entries for testing, if you have a Mac and depending on its default scrolling centers, it may happen that the user doesn't see a scrollbar at all indicating that there are more countries to be displayed, so perhaps adding the "..." would provide a visual indicator that the list is not complete; at least it looks like the easiest way to implement this without an overhaul.

